Issue #1, opened 2017-03-14
Environment variables?


ajoslin:
It would be awesome if Helmfile could cause Charts.yml to interpolate environment variables.

I currently have my project's deployment in a chart directory in my repo. I pass environment variables in via `set`:

```
helm upgrade my-project-$NODE_ENV ./kubernetes/my-project/ \
     --set namespace=$NODE_ENV \
     --set web.imageName=$WEB_NAME \
     --set api.imageName=$API_NAME \
     --set env.dockerRepoUrl=$DOCKER_REPO_PREFIX \
     --set env.dockerImageTag=$DOCKER_REPO_TAG \
     --set env.hostname=$NODE_ENV.$KUBERNETES_HOST \
     --install --debug
```

This is just about the only imperative part of my build step. I'd love to move it into the Helmfile format, but it'd be nice not to have to add a step to run my Helmfile through `envsubst`.

What do you think of the idea of Helmfile doing environment variable substitution on its own?

roboll:
I'm not quite convinced this is a good idea, but it would be relatively simple to template the file. One of the things I like most about this approach is having everything encoded in the helmfile.

I'm not using kubernetes much right now though, so if you think this is a worthy change and can't think of a way to do it without env vars in the file, feel free to submit a pr and I'll accept. 

Running it through `text/template` with a `env` function would let you have something like `{{ env "KEY" "fallback-value" }}` maybe.

natevecc:
I created #2 to address my need for environment variable support. It doesn't do anything fancy with text/template but instead exposes a way for the Chart.yaml file to grab them from the current execution context. Interested in your feedback.

roboll:
This is available in `v0.4`.

PR #2, opened 2017-04-12
Environment Variable Support

natevecc:
Add support for the following flags:

- `--kube-context`: set what kube-context should be passed down to helm
- `charts --values`: add additional value files to the existing list of value files (for use when using overrides for different environments during CI)
- in the Charts.yml file added a new property `env` for environment vars. It's exactly the same as the `set` property but all the value fields are run through `os.Getenv` to resolve their values. Needed this for using secrets in our CI system.

Note: this is my first time working with Go so any and all feedback welcome!

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -32,6 +32,7 @@ type ChartSpec struct {
 	Namespace string     `yaml:"namespace"`
 	Values    []string   `yaml:"values"`
 	SetValues []SetValue `yaml:"set"`
+	EnvValues []SetValue `yaml:"env"`
 }
 
 type SetValue struct {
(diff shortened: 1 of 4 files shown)


state/state.go
```diff
@@ -152,5 +162,17 @@ func flagsForChart(chart *ChartSpec) ([]string, error) {
...
 		flags = append(flags, "--set", strings.Join(val, ","))
 	}
+	if len(chart.EnvValues) > 0 {
+		val := []string{}
+		for _, set := range chart.EnvValues {
+			value, isSet := os.LookupEnv(set.Value)
+			if isSet {
+				val = append(val, fmt.Sprintf("%s=%s", set.Name, value))
+			} else {
+				return nil, errors.New(fmt.Sprintf("Unset env var: %s", set.Name))
```
roboll: nit: This error message could be read as an instruction, e.g. please unset env var {}. Can you change it to something clearer?

natevecc: Will do!

natevecc:
@roboll ready for another look

roboll:
Perfect. Merging this, thanks again. Closes #1.

Issue #8, opened 2017-06-17
Restrict charts to work on via argument


fuhbar:
Hi, 

just stumbled over this tool and could not image how to live without it! Idea is brilliant as it is simple. Thanks for that. :)

I was just wondering, if there is an argument which limits the list of charts to work on by some kind of expression. Its not always desired to iterate across the whole list of charts. Nevertheless its easily possible to write just a `charts.yaml` which lists only the desired charts, but this is some kind of cumbersome. 

Image something like that: `helmfile charts --limit "metrics*"` which just works on the charts where the given glob pattern match against.

roboll:
Thanks for the feedback, glad you are making good use of it. 😎 

This seems reasonable, I guess for cases where you have a single large helmfile. It could be dangerous in that it could lead to some skew between the charts.yaml and kube, but I guess thats up to the user. 

You could also just split out into `metrics.yaml`, `apps.yaml`, whatever and `helmfile -f metrics.yaml ...`.

I'd accept a PR for this but I won't have time to implement it myself for a while.

roboll:
Closing this due to inactivity, if someone has a strong feeling about this and wants to open a PR I'm open to it.

PR #9, opened 2017-06-25
Add filter mechanism to charts command

fuhbar:
As  mentioned in #8 this PR implements the `--filter` parameter for the `charts` command. Purpose is to restrict the concerning charts to be worked on by an glob based expression matching.

### DIFF
--- a/main.go
+++ b/main.go
@@ -84,6 +84,10 @@ func main() {
 					Name:  "values",
 					Usage: "additional value files to be merged into the command",
 				},
+				cli.StringSliceFlag{
+					Name:  "filter",
+					Usage: "only update charts matching the given filter",
+				},
 			},
 			Action: func(c *cli.Context) error {
 				state, helm, err := before(c)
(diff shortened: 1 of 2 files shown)


roboll:
This looks reasonable. Thanks!

The project uses https://github.com/kardianos/govendor for vendoring dependencies, can you add the glob dependency so tests pass? `govendor fetch github.com/gobwas/govendor` should do it.

fuhbar:
@roboll Sorry, missed your comment. Will do that. 

I will rework my changes so that its not only possible to filter desired charts to work on, but rather to be able to exlcude charts as well. If you agree I will remove the `--filter` flag in favour of `--include` and `--exclude`. What do you think?

roboll:
No problem.

The only problem with `--include` and `--exclude` is how they interact. If you want to go that route, make sure docs are clear and you error properly when they are passed in an invalid combination.

roboll:
Any interest in moving this forward?

roboll:
Going to close this, feel free to reopen if you want to pick it back up.

PR #12, opened 2017-08-18
Adding in diff plugin execution

jrnt30:
This allows the execution of the `helm diff` plugin for the entire state file, which for us makes it very easy to visualize the configuration drift of a given environment for auditing and greater confidence in the deployment similar to a `terraform plan` on an environment. 

NOTE: 
- Simple copy-paste for the most part from sync job.  I had started down
the path of adding in a meta PluginCommand directive and trying to make it
more modular, but in the end there are some small differences between
the execution that were a bit difficult to model and it just got ugly.
Figured keeping it simple would be easier to manage. 

If a more extensible form of this is desired I can work on modeling that a bit to make it easier to add in arbitrary sets of functionality but I wanted to get the conversation started

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,12 +49,10 @@ NAME:
 USAGE:
    main [global options] command [command options] [arguments...]
 
-VERSION:
-   0.1.0
-
 COMMANDS:
      repos   sync repositories from state file (helm repo add && helm repo update)
      charts  sync charts from state file (helm repo upgrade --install)
+     diff    diff charts from state file against env (helm diff)
      sync    sync all resources from state file (repos && charts)
      delete  delete charts from state file (helm delete)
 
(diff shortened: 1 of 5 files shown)


roboll:
Cool! Thanks for the contribution.

I won't have a chance to review this for a couple days, but hang tight.

roboll:
I think the code looks okay as is. 

Can you add documentation with pointers on how to get helm-diff set up and update the README with the new command?

jrnt30:
Will do tonight, thanks.

roboll:
I haven't used helm diff before, but assuming you are currently using this and it's good to go, I'm happy to merge when you are ready.

jrnt30:
Appreciate it, I think we are good to go

roboll:
Excellent! Merging this, and I'll cut a new release once we merge your other PR.

PR #20, opened 2017-11-10
Allow env var interpolation in `set` and `values`

fishi0x01:
Thanks a lot for sharing this tool - it really helps out in my workflow!

For my use-case I had to make some adjustments though.
I need to use environment variables inside the paths of the `--values` section.
Also I need to interpolate env vars with `--set` strings. Currently we have the `env` section to use env vars, but it cannot be used for interpolation. 
My use-case is like that:

```
---
charts:
  - name: proxy
    namespace: default
    chart: ./helm/charts/PlatformProxy
    values:
    - ./helm/values/${PLATFORM_ENV}/${PLATFORM_ID}/config.yaml
    - ./helm/values/${PLATFORM_ENV}/${PLATFORM_ID}/secrets.yaml
    set:
    - name: proxy.base_url
      value: "${PLATFORM_ID}.my-domain.com"
```

This PR adds the functionality. In order to render the environment variables reliable, they **MUST** be surrounded by `${}`. An error is thrown, if a variable is not set.   
I removed `env`, because `set` would now provide the same functionality, but if you have backwards compatibility concerns I could also easily integrate it again.

It works for me and I thought this might be worth sharing. Whats your take on this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ charts:
       - name: address
         value: https://vault.example.com
       - name: db.password
-        value: {{ env "DB_PASSWORD" }}                   # value taken from environment variable. Will throw an error if the environment variable is not set. $DB_PASSOWRD needs to be set in the calling environment ex: export DB_PASSWORD='password1'
+        value: {{ env "DB_PASSWORD" }}                   # value taken from environment variable. Will throw an error if the environment variable is not set. $DB_PASSWORD needs to be set in the calling environment ex: export DB_PASSWORD='password1'
       - name: proxy.domain
         value: "{{ env \"PLATFORM_ID\" }}.my-domain.com" # Interpolate environment variable with a fixed string
 
--- a/README.md
+++ b/README.md
@@ -23,23 +23,25 @@ repositories:
 
 charts:
   # Published chart example
-  - name: vault                          # helm deployment name
-    namespace: vault                     # target namespace
-    chart: roboll/vault-secret-manager   # chart reference (repository)
-    values: [ vault.yaml ]               # value files (--values)
-    set:                                 # values (--set)
+  - name: vault                               # helm deployment name
+    namespace: vault                          # target namespace
+    chart: roboll/vault-secret-manager        # chart reference (repository)
+    values: [ vault.yaml ]                    # value files (--values)
+    set:                                      # values (--set)
       - name: address
         value: https://vault.example.com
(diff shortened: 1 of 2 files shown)


roboll:
Thanks for the contribution, I think I like the idea, on the fence about the implementation.

One of the things I considered when I wrote the tool initially was allowing go templating in some of the strings - if we went that route it could look like `./helm/values/{{ env PLATFORM_ENV }}/config.yaml` for example. This is preferable because it's extensible to other functions that might be useful. I'm also not crazy about not being able to escape the `${}`. This way also makes it clear that the templating is being done by helmfile, and not by the shell invocation, which could be confusing.

What do you think?

fishi0x01:
I agree. Using go templating makes sense. I will add it in the next couple days.

fishi0x01:
Ok go templating added. About the `env` section: should I also remove it for good or keep it for backwards compatibility?

roboll:
Let's keep it for backward compatibility, I don't think it's hurting anything. I'll remove it if/when there are other breaking changes and do a major version release.

README.md
```diff
@@ -23,23 +23,25 @@ repositories:
...
+    namespace: vault                       # target namespace
+    chart: roboll/vault-secret-manager     # chart reference (repository)
+    values: [ vault.yaml ]                 # value files (--values)
+    set:                                   # values (--set)
       - name: address
         value: https://vault.example.com
-    env:                                 # values (--set) but value will be pulled from environment variables. Will throw an error if the environment variable is not set.
       - name: db.password
-        value: DB_PASSWORD               # $DB_PASSOWRD needs to be set in the calling environment ex: export DB_PASSWORD='password1'
+        value: {{ env "DB_PASSWORD" }}                   # value taken from environment variable. Will throw an error if the environment variable is not set. $DB_PASSOWRD needs to be set in the calling environment ex: export DB_PASSWORD='password1'
```
roboll: nit: there's a typo there, `DB_PASSOWRD`

roboll:
I'll test this out tomorrow. The change looks good.

I am concerned that this moves away from the helmfile being self contained, and now relies on external inputs (env vars). However, it does make the tool more flexible, so that's something.

fishi0x01:
I understand your concerns. For my use-case env vars allow the same `charts.yaml` file to be easily used across multiple environments (similar to how the same helm chart should also be able to be deployed to any environment). I think that from a CI/CD point of view this might be a valuable feature.

PR #21, opened 2017-11-14
Added ability to set context in charts.yaml

jayste:
I would like to have the ability to set the kubernetes cluster context in the charts.yaml.

This means that upgrade pipelines and users can just use "helmfile sync" and will work correctly across multiple cluster contexts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ repositories:
   - name: roboll
     url: http://roboll.io/charts
 
+context: kube-context					 # kube-context (--kube-context)
+
 charts:
   # Published chart example
   - name: vault                          # helm deployment name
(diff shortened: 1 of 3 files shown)


roboll:
I'm not convinced it's a good idea - it makes the helmfile less portable. You _should_ be able to use the same helmfile across multiple different contexts. However, I don't think it hurts the tool to be flexible, so I'll accept the change. 

It needs some documentation, and it should error when both the command line flag and the yaml entry are defined, rather than overwriting the value from the command line as it does now.

jayste:
I have made the two changes as suggested.

I agree that setting it makes the helmfile less portable, but it means that it is cleaner for deployment pipelines and for operations. Having to helmfile --kube-context BLAH sync every time you run the command is prone to human error.

For the overriding behaviour I was just mapping similar functionality as helm. eg If you use namespace: x in a deployment.yaml that will override --namespace, though definitely agree it is not very transparent.

jayste:
@roboll  Have you had a chance to look at this yet?

Issue #23, opened 2017-12-05
Limit the number of concurrent helm calls


andytom:
I have quite a large helmfile and helm (or tiller) is starting to time out when deploying to my local cluster in minikube. Would it be possible to have an argument to limit the number of concurrent requests being made to helm at a time. Maybe something like `--max-connections=X`, if unset you could leave it as unlimited, but if set `helmfile` should only make at most `X` requests at a time.

roboll:
Seems like a reasonable change. It would need to go [here](https://github.com/roboll/helmfile/blob/master/state/state.go#L91), probably instead of a goroutine per chart, a pool of workers and a work channel.

I'm not going to implement this right now. Feel free to submit a PR.

PR #24, opened 2017-12-07
Add option to limit concurrent helm calls

andytom:
Have added a new argument `--max-connections` that can be used to set a
limit on the number of concurrent calls to helm at one time. To enable
this I have switched `SyncCharts` from using a WaitGroup to using a pool
of workers and a queue of jobs. No changes have been made to the logic
that performs the sync, however that might be something to look at in
the future.

Fixes https://github.com/roboll/helmfile/issues/23

### DIFF
--- a/main.go
+++ b/main.go
@@ -84,6 +84,11 @@ func main() {
 					Name:  "values",
 					Usage: "additional value files to be merged into the command",
 				},
+				cli.IntFlag{
+					Name:  "concurrency",
+					Value: 0,
+					Usage: "maximum number of concurrent helm processes to run, 0 is unlimited",
+				},
 			},
 			Action: func(c *cli.Context) error {
 				state, helm, err := before(c)
@@ -97,8 +102,9 @@ func main() {
 				}
 
 				values := c.StringSlice("values")
+				workers := c.Int("concurrency")
 
-				if errs := state.SyncCharts(helm, values); errs != nil && len(errs) > 0 {
+				if errs := state.SyncCharts(helm, values, workers); errs != nil && len(errs) > 0 {
 					for _, err := range errs {
 						fmt.Printf("err: %s\n", err.Error())
 					}
@@ -164,6 +170,11 @@ func main() {
 					Name:  "values",
 					Usage: "additional value files to be merged into the command",
 				},
+				cli.IntFlag{
+					Name:  "concurrency",
+					Value: 0,
+					Usage: "maximum number of concurrent helm processes to run, 0 is unlimited",
+				},
 			},
 			Action: func(c *cli.Context) error {
 				state, helm, err := before(c)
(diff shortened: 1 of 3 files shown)


andytom:
@roboll I've had a go at implementing the changes required to set the max number of connections. Can you give this a quick review when you get time and let me know if there is anything that should be changed.

roboll:
Thanks for this work.

At first glance, you are appending to `errs` within the goroutine, which isn't safe. Can you make sure this is all thread safe?

andytom:
> Thanks for this work.
>
> At first glance, you are appending to errs within the goroutine, which isn't safe. Can you make sure this is all thread safe?

I can create a channel for the errors and collect them from that channel. Should I leave the if statement on [Line 113](https://github.com/roboll/helmfile/pull/24/files?w=1#diff-f6be34387a141f652301c16cf7a5e36cR113) I'm not sure about that. If one chart fails then it would cause all the others to not be deployed?

roboll:
I _think_ the most intuitive behavior would be: do all operations (there's no explicit order to the operations), print all errors at the end and set exit status based on whether there were any errors.

andytom:
> I *think* the most intuitive behavior would be: do all operations (there's no explicit order to the operations), print all errors at the end and set exit status based on whether there were any errors.

I've added the error channel and (hopefully) updated the logic to match this behavior.

andytom:
Fix an error where it was breaking out of the wrong loop if there was a problem working out the absolute path of a values file.

andytom:
@roboll have you had a chance to look at the changes I've made?

roboll:
Sorry, @andytom - not yet. I will be able to look tonight though.

andytom:
@roboll sorry to keep chasing you but have you had a chance to look at this yet?

andytom:
Have rebased this on the latest master.

matthewellis:
Tested this myself and it seems to work well @andytom, would love to see some movement on this @roboll 👍

main.go
```diff
@@ -84,6 +84,11 @@ func main() {
 					Name:  "values",
 					Usage: "additional value files to be merged into the command",
 				},
+				cli.IntFlag{
+					Name:  "max-connections",
```
mumoshu: @andytom Hi, thx for the PR!

Sorry for being very nit-picky, but I'd prefer `concurrency` over `max-connections` because:

- Helmfile doesn't explicitly deal with connections
- While it doesn't always match `parallelism`, as goroutine's parallelism is capped at GOMAXPROCS

Would you mind if I contributed the change to `concurrency` after this PR gets merged?

andytom: Not at all, I can change that in this PR if you want?

mumoshu: Yes, I appreciate it!

andytom: Done.

andytom:
@mumoshu I've changed the name are there any other changes that need to be made?

mumoshu:
@andytom LGTM. Thanks a lot for your efforts!

Issue #28, opened 2018-01-09
Tags?


xeor:
I really want to use 1 `charts.yaml` file for my project. But I want to be able to control the charts separate as well..

Would a tagging system be a good idea?

Example charts.yaml:

```
charts:
  - name: a1
    chart: ../charts/c1/
    tags:
      - tag1
  - name: a2
    chart: ../charts/c1/
    tags:
      - tag1
      - tag2
  - name: b1
    chart: ../charts/c1/
    tags:
      - tag3
  - name: c1
    chart: ../charts/c1/
```

Then helmfile could be used like:

```sh
# sync a1, a2
helmfile --tags tag1 sync

# sync a1, a2, b1
helmfile --tags tag1,tag3 sync

# sync all
helmfile sync
```

mumoshu:
@xeor This is a great suggestion!

I'd also like to use it like, for example:

```
charts:
  - name: a1
    chart: ../charts/caching-proxy/
    tags:
      - frontend
  - name: a2
    chart: ../charts/web-server/
    tags:
      - frontend
  - name: b1
    chart: ../charts/api-server/
    tags:
      - backend
  - name: c1
    chart: ../charts/db-proxy/
    tags:
      - backend
```

and run `helmfile --tags frontend sync` and/or `helmfile --tags backend sync`.

sstarcher:
I like the idea.  It might make sense to follow the Kubernetes approach of
```
labels:
  key: value
```

xeor:
yea, that labels approach looks better.
Maybe there could be some meta-tags starting with `_`? Like `_name`..? :)

mumoshu:
Thanks for suggestions 👍 

Yeah, the `labels` approach sounds cool.
One thing I'm not sure though is - what if I just want to specify a key but omit its value.

Would my prev example turn into something like the below?

```yaml
charts:
  - name: a1
    chart: ../charts/caching-proxy/
    labels:
      frontend:
  - name: a2
    chart: ../charts/web-server/
    labels:
      frontend:
  - name: b1
    chart: ../charts/api-server/
    labels:
      backend:
  - name: c1
    chart: ../charts/db-proxy/
    labels:
      backend:
```

Hmm, the verbosity does seem like a concern, as you pointed out :)

mumoshu:
@xeor Thx! Would you mind sharing us what's your use-case of the meta tags?
It is just that I like to always start with an use-case :)

mumoshu:
@sstarcher 

[...]

Issue #25, opened 2017-12-10
Helm charts and releases: inconsistent terminology?


ilyasotkov:
Using Helm CLI shows that one Helm release can include several Helm charts - one main chart and several subcharts. Am I missing something or should `charts.yaml` therefore be `releases.yaml`?

Example: 

Generic chart: rails-app
Unique releases: dev-example-com, prod-example-com, dev-example-org, prod-example-org

Workflow:

1. Create a generic Helm chart with some default `values.yaml`
2. Add one or more releases, each with its own unique `values.yaml`

One cluster, many charts, more releases 🤔

“Helm deployment” here should be “Helm release” also. “Deployment” is a very specific Kubernetes term, and not a Helm term at all.

Am I wrong?

xeor:
I've just found helmfile as well and was thinking the same. `charts` and `deployment` are terms that shouln't be reused in this project. At least not `charts`

roboll:
> Using Helm CLI shows that one Helm release can include several Helm charts - one main chart and several subcharts

Yes but helm deals only in `chart`s - there is no difference between a single chart or an aggregate chart. The `charts.yml` file and `charts` section in it reference charts that helm will deploy. They may produce something you could call a release, but I'm not convinced the naming change is required.

I agree that the overlap in terminology for `deployment` isn't great - if you'd like to pr a change to the docs to call it a release I'd accept that.

xeor:
I must admin I didn't even notice the `s` in your `charts.yml` file before you said it.. In my mind, it ended up being the same name except I knew the `helm` version used uppercase `C`.
Maybe it's because you really don't read the whole word, you just see that the file is there.. There are [a couple](http://www.boats.com/resources/beginners-guide-to-boat-terminology/) of boat terminologies that might work..? Or maybe just call it `Helmfile`?

`charts.yml` aint bad either, now that I know that the s is there :p

gtaylor:
`charts.yml` isn't bad, but `helmfile.yml` would be slightly more obvious as to where this file came from.

mumoshu:
@roboll First of all, thanks for sharing a great project 👍 

The OP's suggestion makes sense to me.

> The charts.yml file and charts section in it reference charts that helm will deploy. They may produce something you could call a release, but I'm not convinced the naming change is required.

Probably the description in `helm install --help` would help here.

```
$ helm install --help
This command installs a chart archive.

The install argument must be a chart reference, a path to a packaged chart,
[...]

PR #30, opened 2018-01-25
Allow running helm against a subsets of charts in a file

awwithro:
Similar to what's described in https://github.com/roboll/helmfile/issues/8

Globbing isn't useful for my particular workflow but I do have a large helmfile where its useful to run just one or two charts out of the entire file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ GLOBAL OPTIONS:
    --file FILE, -f FILE         load config from FILE (default: "helmfile.yaml")
    --quiet, -q                  silence output
    --namespace value, -n value  Set namespace. Uses the namespace set in the context by default
-   --labels value                 Only run using the releases that match labels. Tags can take the form of foo=bar or foo!=bar.
+   --selector,l value           Only run using the releases that match labels. Labels can take the form of foo=bar or foo!=bar.
                                 A release must match all labels in a group in order to be used. Multiple groups can be specified at once.
                                 --labels tier=frontend,tier!=proxy --labels tier=backend. Will match all frontend, non-proxy releases AND all backend releases.
                                 The name of a release can be used as a label. --labels name=myrelease
@@ -115,12 +115,12 @@ A few rules to clear up this ambiguity:
 
 For additional context, take a look at [paths examples](PATHS.md)
 ## Labels Overview
-Labels can be used to only use a subset of releases when running helmfile. This is useful for large helmfiles with releases that are logically grouped together.
+A selector can be used to only target a subset of releases when running helmfile. This is useful for large helmfiles with releases that are logically grouped together.
 
-Labels are simple key value pairs that are an optional field of the release spec. When filtering by label, the search can be inverted. `tier!=backend` would match all releases that do NOT have the `tier: backend` label. `tier=fronted` would only match releases with the `tier: frontend` label.
+Labels are simple key value pairs that are an optional field of the release spec. When selecting by label, the search can be inverted. `tier!=backend` would match all releases that do NOT have the `tier: backend` label. `tier=fronted` would only match releases with the `tier: frontend` label.
 
-Multiple labels can be specified using `,` as a separator. A release must match all labels in order to be selected for the final helm command. 
+Multiple labels can be specified using `,` as a separator. A release must match all selectors in order to be selected for the final helm command. 
 
-The `labels` parameter can be specified multiple times. Each parameter is resolved independently so a release that matches any parameter will be used. 
+The `selector` parameter can be specified multiple times. Each parameter is resolved independently so a release that matches any parameter will be used. 
 
 `--selector tier=frontend --selector tier=backend` will select all the charts
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ releases:
   # Published chart example
   - name: vault                            # name of this release
     namespace: vault                       # target namespace
-    tags:                                  # Arbitrary key value pairs for filtering releases
+    labels:                                  # Arbitrary key value pairs for filtering releases
       foo: bar
     chart: roboll/vault-secret-manager     # the chart being installed to create this release, referenced by `repository/chart` syntax
     values: [ vault.yaml ]                 # value files (--values)
@@ -78,10 +78,10 @@ GLOBAL OPTIONS:
    --file FILE, -f FILE         load config from FILE (default: "helmfile.yaml")
    --quiet, -q                  silence output
    --namespace value, -n value  Set namespace. Uses the namespace set in the context by default
-   --tags value                 Only run using the releases that match tags. Tags can take the form of foo=bar or foo!=bar.
-                                A release must match all tags in a group in order to be used. Multiple groups can be specified at once.
-                                --tags tier=frontend,tier!=proxy --tags tier=backend. Will match all frontend, non-proxy releases AND all backend releases.
-                                The name of a release can be used as a tag. --tags name=myrelease
+   --labels value                 Only run using the releases that match labels. Tags can take the form of foo=bar or foo!=bar.
+                                A release must match all labels in a group in order to be used. Multiple groups can be specified at once.
+                                --labels tier=frontend,tier!=proxy --labels tier=backend. Will match all frontend, non-proxy releases AND all backend releases.
+                                The name of a release can be used as a label. --labels name=myrelease
    --kube-context value         Set kubectl context. Uses current context by default
    --help, -h                   show help
    --version, -v                print the version
@@ -114,11 +114,13 @@ A few rules to clear up this ambiguity:
 - Relative paths referenced on the command line are relative to the current working directory the user is in
 
 For additional context, take a look at [paths examples](PATHS.md)
-## Tags Overview
-Tags can be used to only use a subset of releases when running helmfile. This is useful for large helmfiles with releases that are logically grouped together.
+## Labels Overview
+Labels can be used to only use a subset of releases when running helmfile. This is useful for large helmfiles with releases that are logically grouped together.
 
-Tags are simple key value pairs that are an optional field of the release spec. When filtering by tag, the search can be inverted. `tier!=backend` would match all releases that do NOT have the `tier: backend` tag. `tier=fronted` would only match releases with the `tier: frontend` tag.
+Labels are simple key value pairs that are an optional field of the release spec. When filtering by label, the search can be inverted. `tier!=backend` would match all releases that do NOT have the `tier: backend` label. `tier=fronted` would only match releases with the `tier: frontend` label.
 
-Multiple tags can be specified using `,` as a separator. A release must match all tags in order to be selected for the final helm command. 
+Multiple labels can be specified using `,` as a separator. A release must match all labels in order to be selected for the final helm command. 
 
-The `tags` parameter can be specified multiple times. Each parameter is resolved independently so a release that matches any parameter will be used. 
+The `labels` parameter can be specified multiple times. Each parameter is resolved independently so a release that matches any parameter will be used. 
+
+`--selector tier=frontend --selector tier=backend` will select all the charts
--- a/README.md
+++ b/README.md
@@ -79,9 +79,9 @@ GLOBAL OPTIONS:
    --quiet, -q                  silence output
    --namespace value, -n value  Set namespace. Uses the namespace set in the context by default
    --selector,l value           Only run using the releases that match labels. Labels can take the form of foo=bar or foo!=bar.
-                                A release must match all labels in a group in order to be used. Multiple groups can be specified at once.
-                                --labels tier=frontend,tier!=proxy --labels tier=backend. Will match all frontend, non-proxy releases AND all backend releases.
-                                The name of a release can be used as a label. --labels name=myrelease
+	                              A release must match all labels in a group in order to be used. Multiple groups can be specified at once.
+	                              --selector tier=frontend,tier!=proxy --selector tier=backend. Will match all frontend, non-proxy releases AND all backend releases.
+	                              The name of a release can be used as a label. --selector name=myrelease
    --kube-context value         Set kubectl context. Uses current context by default
    --help, -h                   show help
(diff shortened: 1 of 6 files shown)


state/state.go
```diff
@@ -12,10 +12,11 @@ import (
 
 	"github.com/roboll/helmfile/helmexec"
 
-	yaml "gopkg.in/yaml.v1"
```
awwithro: My IDE runs go fmt so there are a few formatting changes not directly related to the PR. Not sure if you have strong feelings on that.

awwithro:
@roboll ping! Any feedback on this PR?

sstarcher:
Anything needed to get this merged

roboll:
@sstarcher I haven't been using the tool lately, so just time to review and test.

If you (or anyone else) are actively using the tool and can confirm it works as expected, I'll accept the PR.

sstarcher:
I'll make some time to test it out.  Thanks

sstarcher:
@roboll are you using something else to accomplish similar functionality to helmfile?

roboll:
@sstarcher nope, tool is still relevant, just not doing anything with helm at the moment.

mumoshu:
I like this feature!

However, I'm afraid if merging this prevents a similar but a more high-level filtering feature like @xeor proposed in #28?
Can I contribute the impl for #28 in addition to this?

sstarcher:
I would not see this as mutually exclusive with #28

awwithro:
I don't see them as mutually exclusive either but personally would use
different different helmfiles over tags if the charts were that closely
related

state/state.go
```diff
@@ -218,6 +219,26 @@ func (state *HelmState) DeleteCharts(helm helmexec.Interface) []error {
...
 }
 
+// FilterCharts allows for the execution of helm commands against a subset of the charts in the helmfile.
+// returns an error a chartName doesn't exist in the list of charts
+func (state *HelmState) FilterCharts(chartNames []string) error {
+	var filteredCharts []ChartSpec
+	for _, name := range chartNames {
+		missingName := true
+		for _, chart := range state.Charts {
+			if chart.Name == name {
```
mumoshu: Nearly a nit but I wanted to confirm if, probably you're confusing release name vs chart name?
`ChartSpec.Name` stands for release name, whereas `ChartSpec.Chart` is the name of chart used to create the release.
Also see #25 for more context.

mumoshu: I'm going to fix the terminology issue starting #35. So I'd fix it on my side if necessary.

mumoshu: Your intention is to filter releases by release names rather than to filter charts by chart names, right?

awwithro: Absolutely, I see you've made things more consistent across the board (and more precise) in your PR. I'll rebase and fix my terminology.

awwithro:
@mumoshu PTAL, I've switched the terminology to reference release instead of chart which looks consistent with your rewrite.

state/state.go
```diff
@@ -213,7 +213,7 @@ func (state *HelmState) SyncReleases(helm helmexec.Interface, additonalValues []
 	return nil
 }
 
-func (state *HelmState) DiffReleases(helm helmexec.Interface, additonalValues []string) []error {
+func (state *HelmState) DiffReleases(helm helmexec.Interface, additionalValues []string) []error {
```
mumoshu: Good catch :+1:

main.go
```diff
@@ -43,6 +43,10 @@ func main() {
 			Name:  "namespace, n",
 			Usage: "Set namespace. Uses the namespace set in the context by default",
 		},
+		cli.StringSliceFlag{
+			Name:  "release-name, r",
```
mumoshu: Sorry but honestly saying, I'm not yet sure if this is the right name(if there's one) which describes the feature.
More concretely, I imagine that one would misunderstand this as a flag used for somehow overriding release names, which isn't true of course.

Instead, could this be any of the followings?

1. `helmfile sync --only myrelease1,myrelease2`
   - not very intuitive as it isn't clear that what we can filter & by what we can filter releases
2. `helmfile sync --only-releases myrelease1,myrelease2`
   - it isn't clear that by what we can filter releases
3. `helmfile sync --filter-releases myrelease1,myrelease2`
   - same as 3
4. `helmfile sync --filter-releases-by-names myrelease1,myrelease2`
   - imho, too verbose
5. `helmfile sync --filter-releases name=myrelease1,myrelease2`
   - what if we wanted to support #28? `--filter-releases tag=tier1:frontend`? not sure i could correctly differentiate `=` and `:` :)
6. `helmfile sync myrelease1,myrelease2`
7. `helmfile sync myrelease1 myrelease2`

I slightly prefer 7 as it straightly maps to my mental model that `helmfile` is working against releases in the first place. It would also look familiar to ones used to kubectl. `kubectl get po` works against all the pods, whereas `kubectl get po foo`  filters the target to a single pod.

A nice side-effect of it is that we now have a natural integration with shell brace expansion e.g. `helmfile sync myrelease{1,2}` 😃

state/state_test.go
```diff
@@ -56,3 +56,37 @@ releases:
 		t.Error("expected error")
 	}
 }
+
+func TestReleaseFilter(t *testing.T) {
```
mumoshu: ❤️

mumoshu:
@awwithro Wow, thanks for your efforts! The implementation looks awesome. I really appreciate you adding a test, too!

My final point is about how we should name the flag. WDYT?

main.go
sstarcher: I would either recommend 7 or `helmfile sync --filter name=myrelease`  Where in the future we could support more filters

awwithro: `helmfile sync --filter name=myrelease` seems the most natural to me as well considering the proposed feature to use tags as a filter. 

7. is nice and clean but I don't think would be extended with a tag structure as well. It would be less obvious what the behavior would be with names and tags. `helmfile sync --tags tier:frontend myrelease1 myrelease2` would that apply the intersection or union of the two filters? This question stands either way but I think its more clear using a `--filter` flag that either tags or names should be used. 

I suppose 7 could lead to `helmfile sync tag:tier=frontend` or similar but that doesn't seem as clean to me

sstarcher: @awwithro for supporting and/or something like
```helmfile sync --filter name=bob,tier=frontend --filter name=sam```
Any release that makes name=bob and tier=frontend AND any chart that has the name sam.

That would allow us to support and/or logic.

Comment (sstarcher) [reply main.go]: @awwithro thanks that work will help toward the full tag/filter implementation

Comment (mumoshu) [reply main.go]: @awwithro Thanks for your efforts! Please let us know if there's anything we can help. Probably questions regarding implementation, design, testing, etc?

Comment (awwithro) [reply main.go]: hey @mumoshu I've been a bit busy at work and haven't had a chance to circle back on this quite yet. I did have some questions though. I believe you're on the kubernetes slack? I can find you there

**Comment by mumoshu** [comment]
yes, lets discuss there!
2018年3月13日(火) 23:51 Alex Withrow <notifications@github.com>:

**Comment by mumoshu** [comment]
Started talking with @awwithro towards finishing this

File: README.md [hunk 6]
```diff
@@ -29,6 +29,8 @@ releases:
   # Published chart example
   - name: vault                            # name of this release
     namespace: vault                       # target namespace
+    tags:                                  # Arbitrary key value pairs for filtering releases
```
awwithro: What do you think about using `labels` instead of `tags`? would be consistent with k8s.

awwithro:
@mumoshu PTAL, I refactored thing similar to what we discussed with tags/names. What do you think? Anything missing or in need of changing?

sstarcher:
I tested running a single example and was initially confused that the globals don't get passed down to the sub functions

Works
```helmfile --tags name=filebeat diff``` 

Does not work
```helmfile diff --tags name=filebeat```

**Comment by sstarcher** [comment]
Also verified != works

File: state/state_test.go [hunk 7]
```diff
@@ -62,6 +62,73 @@ releases:
...
+    tier: frontend
+    foo: bar
+- name: myrelease2
+  chart: mychart2
+  tags:
+    tier: frontend
+- name: myrelease3
+  chart: mychart3
+  tags:
+    tier: backend
```
mumoshu: Thx for writing an effective test 👍

state/state.go
```diff
@@ -295,6 +296,39 @@ func (state *HelmState) Clean() []error {
...
+			return err
+		}
+		for _, release := range state.Releases {
+			// Allow the name of the release to be matched like a tag
+			if release.Tags == nil {
+				release.Tags = map[string]string{}
+			}
+			release.Tags["name"] = release.Name
+			if filter.Match(release) {
+				// Only add a release once
```
mumoshu: The functionality is awesome!
If I could ask more, probably you can change the nesting so that there would be less code, plus and a bit efficiency?

```golang
var filters := []ReleaseFilter{}
for _, tag := range tags {
    f, err := ParseTags(tag)
    // snip - error handling
    filters = append(filters, f)
}
for _, r := range state.Releases {
    for _, f := range filters {
        if f.Match(releases) {
            filteredReleases = append(filteredReleases, release)
            continue
        }
    }
}
```

README.md
mumoshu: Sounds good in general.

One thing I'm wondering on top of that though is - Would it also be reasonable to rename the `--tags` flag to `--selector` for more consistency with k8s'?

And in that case, I can't help expecting that `labels` could be somehow propagated as `values` to helm charts, so that we can use the same set of `labels` to filter both releases(`helmfile --selector ...`) and resulting k8s objects(`kubectl get --selector ...`). What do you think about that?

awwithro: I like the `-l`,`--selector` standardization. I'll switch to that. As far as propagating labels to the resulting k8s objects. I like the idea. I'm not sure how the implementation would work exactly since the objects would be managed via helm but is out of scope for this PR

state/state.go
awwithro: 👍

README.md
mumoshu: Note: change the remaining `tags:` to `labels:`

mumoshu: Note: change the remaining `tags:` to `labels:`

README.md
```diff
@@ -73,11 +75,16 @@ COMMANDS:
...
 GLOBAL OPTIONS:
-   --file FILE, -f FILE  load config from FILE (default: "helmfile.yaml")
-   --quiet, -q           silence output
-   --kube-context value  Set kubectl context. Uses current context by default
-   --help, -h            show help
-   --version, -v         print the version
+   --file FILE, -f FILE         load config from FILE (default: "helmfile.yaml")
+   --quiet, -q                  silence output
+   --namespace value, -n value  Set namespace. Uses the namespace set in the context by default
+   --tags value                 Only run using the releases that match tags. Tags can take the form of foo=bar or foo!=bar.
```
mumoshu: Note: change the remaining `tags:` to `labels:`

README.md
```diff
@@ -107,3 +114,11 @@ A few rules to clear up this ambiguity:
 - Relative paths referenced on the command line are relative to the current working directory the user is in
 
 For additional context, take a look at [paths examples](PATHS.md)
+## Tags Overview
```
mumoshu: Note: `Tags` -> `Labels`

README.md
```diff
@@ -114,11 +114,13 @@ A few rules to clear up this ambiguity:
...
+Labels are simple key value pairs that are an optional field of the release spec. When filtering by label, the search can be inverted. `tier!=backend` would match all releases that do NOT have the `tier: backend` label. `tier=fronted` would only match releases with the `tier: frontend` label.
 
-Multiple tags can be specified using `,` as a separator. A release must match all tags in order to be selected for the final helm command. 
+Multiple labels can be specified using `,` as a separator. A release must match all labels in order to be selected for the final helm command. 
 
-The `tags` parameter can be specified multiple times. Each parameter is resolved independently so a release that matches any parameter will be used. 
\ No newline at end of file
+The `labels` parameter can be specified multiple times. Each parameter is resolved independently so a release that matches any parameter will be used. 
+
+`--selector tier=frontend --selector tier=backend` will select all the charts
```
sstarcher: this should be --labels right?

main.go
```diff
@@ -45,6 +45,13 @@ func main() {
 			Name:  "namespace, n",
 			Usage: "Set namespace. Uses the namespace set in the context by default",
 		},
+		cli.StringSliceFlag{
+			Name: "selector, l",
+			Usage: `Only run using the releases that match labels. Labels can take the form of foo=bar or foo!=bar.
+	A release must match all labels in a group in order to be used. Multiple groups can be specified at once.
+	--selector tier=frontend,tier!=proxy --selector tier=backend. Will match all frontend, non-proxy releases AND all backend releases.
```
sstarcher: should all references to selector be changed to `labels`?

README.md
awwithro: `--selector`,`-l` is consistent with `kubectl` and how it selects objects using labels. I'd used `-l` myself previously but I think having a common operation across tools is desirable.

main.go
awwithro: similar to k8s, the idea is that labels are on objects, and a selector selects based on labels

sstarcher: ahh if that's the case the docs that are calling out `--label` should be changed right?

```--labels tier=frontend,tier!=proxy --labels tier=backend. Will match all frontend, non-proxy releases AND all backend releases.```

to 
```
--selector tier=frontend,tier!=proxy --selector tier=backend. Will match all frontend, non-proxy releases AND all backend releases.```

sstarcher: Agreed we should use --selector and -l, but we just need to make sure all the docs align.

awwithro: it should be consistent across the docs now

sstarcher:
After this is merged could we cut a release for secrets and this.

mumoshu:
@awwithro @sstarcher Thank you very much for your efforts and supports!

awwithro:
thanks @mumoshu @sstarcher for the ideas and feedback!

PR #35, opened 2018-02-27
Fix terminology

### LINKED ISSUE #25: Helm charts and releases: inconsistent terminology?

mumoshu:
In a backward compatible manner, so that (I believe) we could move the discussion made in #25 forward.

Possibly fixes #25

### DIFF
--- a/main.go
+++ b/main.go
@@ -13,7 +13,8 @@ import (
 )
 
 const (
-	helmfile = "charts.yaml"
+	DefaultHelmfile    = "helmfile.yaml"
+	DeprecatedHelmfile = "charts.yaml"
 )
 
 var Version string
@@ -21,13 +22,13 @@ var Version string
 func main() {
 
 	app := cli.NewApp()
-	app.Name = "helmfile"
+	app.Name = "DefaultHelmfile"
 	app.Usage = ""
 	app.Version = Version
 	app.Flags = []cli.Flag{
 		cli.StringFlag{
 			Name:  "file, f",
-			Value: helmfile,
+			Value: DefaultHelmfile,
 			Usage: "load config from `FILE`",
 		},
 		cli.BoolFlag{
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -221,21 +222,28 @@ func before(c *cli.Context) (*state.HelmState, helmexec.Interface, error) {
 	quiet := c.GlobalBool("quiet")
 	kubeContext := c.GlobalString("kube-context")
 
-	state, err := state.ReadFromFile(file)
+	st, err := state.ReadFromFile(file)
+	if err != nil && strings.Contains(err.Error(), fmt.Sprintf("open %s:", DefaultHelmfile)) {
+		var fallbackErr error
+		st, fallbackErr = state.ReadFromFile(DeprecatedHelmfile)
```
sstarcher: when falling back to the deprecated value should we log a message to tell them it's deprecated?

mumoshu: Good point! I believe so.

mumoshu: Added to https://github.com/roboll/helmfile/pull/35/files#diff-7ddfb3e035b42cd70649cc33393fe32cR249

mumoshu:
@sstarcher Thanks for reviewing 👍

PR #39, opened 2018-02-27
Allow specifying a default namespace used across all the releases

### LINKED ISSUE #31: Allow specifying namespace
We would like to deploy the same group of charts to the same cluster in different namespaces (branch environments)

Would there be any desire to see that in helmfile? eg `--kube-namespace`

mumoshu:
Omit `namespace` in `charts[].namespace` of charts.yaml and then run `helmfile sync -n yourns` to install your charts into `yourns`.

Resolves #31

### DIFF
--- a/main.go
+++ b/main.go
@@ -38,6 +38,10 @@ func main() {
 			Name:  "kube-context",
 			Usage: "Set kubectl context. Uses current context by default",
 		},
+		cli.StringFlag{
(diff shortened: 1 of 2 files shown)


mumoshu:
I'm willing to add some tests for this once an another PR adding `state/state_test.go` is merged. I just want to avoid a merge conflict :)

PR #47, opened 2018-03-08
Support client certs when accessing remote repos

gburton1:
The "helm repo add" command already supports the cert-file and key-file values, so we just have to pass them through. Everything works the same if you don't provide values for them. This is important for deployments where the chart repo is protected by client cert validation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,8 @@ The default helmfile is `helmfile.yaml`:
 repositories:
   - name: roboll
     url: http://roboll.io/charts
+    cert-file: optional_client_cert
+    key-file: optional_client_key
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -20,6 +20,8 @@ The default helmfile is `helmfile.yaml`:
 repositories:
   - name: roboll
     url: http://roboll.io/charts
+    cert-file: optional_client_cert
```
mumoshu: Super nit but I slightly prefer `certFile` and `keyFile` as idiomatic yaml object keys seen around the K8S world!
Would you mind if I changed these keys after merging?

mumoshu:
@gburton1 LGTM. Thanks again for your contribution 👍

PR #53, opened 2018-03-15
adding secret processing for helmfile via helm-secrets plugin

### LINKED ISSUE #41: Integration with helm-secrets
I would be interested in getting helmfile integrated with the helm-secrets project.  From looking at it, it might be possible to just change the helm binary that is called to the helm-secrets wrapper with a command line flag.

https://github.com/futuresimple/helm-secrets

How are others handling secrets?

sstarcher:


sstarcher:
@mumoshu

sstarcher:
fixes #41

helmexec/helmexec.go
```diff
@@ -9,4 +9,6 @@ type Interface interface {
 	SyncRelease(name, chart string, flags ...string) error
 	DiffRelease(name, chart string, flags ...string) error
 	DeleteRelease(name string) error
+
+	SecretDecrypt(name string) (string, error)
```
mumoshu: Very nit but I slightly prefer `DecryptSecret` so that it looks consistent with other functions!

README.md
```diff
@@ -87,6 +89,13 @@ Under the covers Helmfile is simply using the `helm diff` plugin, so that needs
 you should be able to simply execute `helm plugin install https://github.com/databus23/helm-diff`. For more details
 please look at their [documentation](https://github.com/databus23/helm-diff#helm-diff-plugin).
 
+### secrets
+
+The `secrets` parameter in a `helmfile.yaml` causes the [helm-secrets](https://github.com/futuresimple/helm-secrets) plugin to be executed to decrypt the file.
+
+Under the covers Helmfile is simply using the `helm secrets` plugin, so that needs to be installed prior.  For Helm 2.3+
```
mumoshu: Although I'm not a native English user, probably:

> Under the covers, Helmfile uses the [helm-secrets](https://github.com/futuresimple/helm-secrets) plugin and therefore you need to install it prior to start using this feature.

would sound a bit more natural(to me 😉).

state/state.go
```diff
@@ -300,6 +326,21 @@ func flagsForRelease(basePath string, release *ReleaseSpec) ([]string, error) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		valfileRendered, err := helm.SecretDecrypt(path)
+		if err != nil {
+			return nil, err
+		}
+
+		release.generatedSecrets = append(release.generatedSecrets, valfileRendered)
```
mumoshu: I prefer calling it `generatedValues` rather than `generatedSecrets`, because what you append to it here is actually a `valfile` generated from a secrets file.

README.md
sstarcher: I stole this from the previous block on helm-diff ;)

sstarcher:
@mumoshu addressed

mumoshu:
Awesome! Thank you very much for your continuous support ☺️

PR #54, opened 2018-03-15
Fix example to use expected "charts:" key.

witten:
This example uses a "releases:" key for the charts list, but when writing my own charts.yaml, only "charts:" appears to work, while "releases:" results in no charts changes when helmfile is run. All the other example files distributed with helmfile also appear to use "charts:".

Note that there appears to be some unrelated test code that also uses "releases:". You may want to fix that as well. I didn't bite that off as part of this because I don't really know Go.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ repositories:
 
 context: kube-context					 # kube-context (--kube-context)
 
-releases:
+charts:
   # Published chart example
   - name: vault                            # name of this release
     namespace: vault                       # target namespace


mumoshu:
@witten Sorry for the confusion but the README is reflecting the state of the master branch, which isn't released yet. The next release of helmfile supports both `releases` and `charts`, whereas the latter is for backward-compatibility. I'll try to release it ASAP to prevent further confusions. 

Thank you anyway for your effort!

mumoshu:
@witten [v0.10](https://github.com/roboll/helmfile/releases/tag/v0.10) has just been released with the change 🎉

witten:
Cool, thanks for clarifying!

Issue #55, opened 2018-03-16
helmfile sync never process deployment CentOS 7.3, Ubuntu 16.04.4 LTS


xjantoth:
Hi there,

I just wanted to rewrite this piece of **working code**:

    helm install stable/prometheus --name prom --set rbac.create=false --namespace=prometheus

**my simple helmfile:**

    vim charts.yaml
     
    ...
    context:  <my_context>
    releases:
      # Published chart example
      - name: prometheusnorbac                            # name of this release
        namespace: prometheus                       # target namespace
        chart: stable/prometheus     # the chart being installed to create this release, referenced by `repository/chart` syntax
        set:                                   # values (--set)
          - name: rbac.create
            value: false
    ...
    wq!

**When I run:**

```
./helmfile -f charts.yaml sync
exec: helm repo update
Hang tight while we grab the latest from your chart repositories...
...Skip local chart repository
...Successfully got an update from the "stable" chart repository
Update Complete. ⎈ Happy Helming!⎈
```

**Then chcek helm list**

```
 helm list
NAME            REVISION        UPDATED                         STATUS          CHART                   NAMESPACE
prom            1               Thu Mar 15 14:22:19 2018        DEPLOYED        prometheus-5.4.3        prometheus
promt           1               Fri Mar 16 11:50:31 2018        FAILED          prometheus-5.4.3        prometheus
promtemp        1               Fri Mar 16 10:58:31 2018        DEPLOYED        prometheus-5.4.3        prometheus

```

So from `helm list` output can be seen that deployment for `prometheusnorbac` was never finished.

**Tested OS's**
* Ubuntu 16.04.4 LTS
* Centos 7.3

mumoshu:
Hi @xjantoth Sorry for the inconvenience. We've recently changed the key from `charts:` to `releases:`, and the binary had not been released since then.

But now, we have [v0.10](https://github.com/roboll/helmfile/releases/tag/v0.10) available for you! Would you mind trying once again with the new version? Thx.

xjantoth:
Hi @mumoshu thanks for a new version **0.10**:


```

vim helmfile.yaml

...
context: <my_context>                   # kube-context (--kube-context)

releases:
  # Published chart example
  - name: prometheusnorbac                            # name of this release
    namespace: prometheus                       # target namespace
    releases: stable/prometheus     # the chart being installed to create this release, referenced by `repository/chart` syntax
    set:                                   # values (--set)
      - name: rbac.create
        value: false
...

```
[...]

PR #57, opened 2018-03-18
Improve readme for first-time users

mumoshu:
So that our user can start leveraging helmfile without too much confusion.

Includes two things:

- A "getting started" guide
- A basic description of the `sync` subcommand

cc/ @xjantoth I've improved the README according to your suggestion at https://github.com/roboll/helmfile/issues/55#issuecomment-373737432.

Would this look good to you? Thanks!


mumoshu:
@sstarcher Thx for the review :relaxed:

PR #72, opened 2018-03-25
describing current functionality of helmexec as tests

cmeury:
I was looking at the code and noticed the lack of unit tests for the `helmexec` package, so I decided to add a basic first version, bringing statement coverage to 83.7%. There's still a lot of test cases missing (like empty or nil arguments), but they would have required actual code changes to support the edge cases. I wanted to just describe existing, happy-path functionality first.

Writing the tests required me to factor out the actual command-line execution bit into the `Runner`.

Beware, I'm a go beginner. I'd appreciate a critical look at this, especially the package-local functions in the interface in `helmexec.go`. Go crazy with critique :P

### DIFF
--- a/helmexec/helmexec.go
+++ b/helmexec/helmexec.go
@@ -1,5 +1,7 @@
 package helmexec
 
+import "io"
+
 type Interface interface {
 	SetExtraArgs(args ...string)
 
@@ -11,4 +13,11 @@ type Interface interface {
 	DeleteRelease(name string) error
 
 	DecryptSecret(name string) (string, error)
+
+	// unit testing
+	exec(args ...string) ([]byte, error)
+	setRunner(runner Runner)
+	getExtra() []string
+	getKubeContent() string
+	getWriter() io.Writer
 }
--- a/helmexec/helmexec.go
+++ b/helmexec/helmexec.go
@@ -1,7 +1,5 @@
 package helmexec
 
-import "io"
-
 type Interface interface {
 	SetExtraArgs(args ...string)
 
@@ -13,11 +11,4 @@ type Interface interface {
 	DeleteRelease(name string) error
 
 	DecryptSecret(name string) (string, error)
-
-	// unit testing
-	exec(args ...string) ([]byte, error)
-	setRunner(runner Runner)
-	getExtra() []string
-	getKubeContent() string
-	getWriter() io.Writer
 }
--- a/helmexec/runner.go
+++ b/helmexec/runner.go
@@ -15,10 +15,10 @@ type Runner interface {
 	Execute(cmd string, args []string) ([]byte, error)
 }
 
-type CliRunner struct {}
+type CliRunner struct{}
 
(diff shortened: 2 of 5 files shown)


helmexec/helmexec.go
```diff
@@ -11,4 +13,11 @@ type Interface interface {
 	DeleteRelease(name string) error
 
 	DecryptSecret(name string) (string, error)
+
+	// unit testing
+	exec(args ...string) ([]byte, error)
+	setRunner(runner Runner)
+	getExtra() []string
+	getKubeContent() string
+	getWriter() io.Writer
```
mumoshu: You've made me realize a good point!

TL;DR; For this case, `func NewHelmExec(writer io.Writer, kubeContext string) Interface {` should return the concrete type, rather than the interface so that it looks `func NewHelmExec(writer io.Writer, kubeContext string) execer {`.

Doing so would probably make `getExtra`, `setRunner`, `getKubeContent`, `getWriter` unnecessary, because then you can just reference the fields directly in tests you've added.

`type Interface` should be declared in the `main` package hence `./helmexec/helmexec.go` should be moved to `./helmexec.go`.

I can't say this pattern should be enforced everywhere, but you can consult [this reddit and the article](https://www.reddit.com/r/golang/comments/5cos3x/accept_interfaces_return_structs/) for other pros of doing it, too.

helmexec/runner.go
```diff
@@ -0,0 +1,30 @@
...
+	"os"
+	"os/exec"
+)
+
+const (
+	tmpPrefix = "helmfile-"
+	tmpSuffix = "-exec"
+)
+
+type Runner interface {
```
mumoshu: Can we name `CliRunner` as `shellExecer` and omit this interface? The rationale is that, at this point, we don't have any concrete benefit by factoring out an interface.

cmeury:
Thanks for the detailed review @mumoshu, really helpful. I'll work that into the branch next time I got the chance to work on this.

helmexec/helmexec.go
cmeury: I was thinking the same thing about the return time, changed. However, I am not sure I follow how the `Interface` should now go into the main package. The `state` package needs access to it and it seems to also be more at home in the `helmexec` package since it describes its functionality.

helmexec/runner.go
cmeury: I tried to do this, but ultimately failed to see how I can use a mocked runner without an interface. We need to set the runner explicitly in `helmexec.New()`, so it needs to be an interface. I renamed it to `ShellRunner` though, to honour your sentiment :)

Maybe I need to go back to the books about Go, I feel I am not seeing these things clearly.

mumoshu: Good point! I had definitely overlooked that. Thx for the rename, too 😄

mumoshu:
This is really a great step forward. Thank you very much for your contribution @cmeury!

PR #74, opened 2018-03-26
doc: Add a basic example for managing oneshot jobs

### LINKED ISSUE #49: Oneshot jobs
I'm a bit hesitant to propose this but how about supporting oneshot jobs in helmfile?

An oneshot job can be either a K8S job or a pod created via `helm upgrade --install` to accomplish a job like db migration, enqueueing a large job to a kind of workflow system, etc.

As of today, we have to resort a variant of `kubectl run --image $img` to run an oneshot job, even in a situation that our regular app are managed by helm and shared the image, envvars with oneshot jobs.

AFAIK, there's no such tool to support this use-case. And supporting it in helmfile allows us to provide a more docker-compose-like U/X to helm users, but the docker-build feature.

Thoughts?

cc/ @roboll @sstarcher

mumoshu:
Resolves #49

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,35 @@
+# Helmfile practical examples and advanced usages
+
+## Managing oneshot-jobs with Helmfile
+
+In case you want to manage oneshot-jobs like a report generation or a database migration and so on, add a dedicated release spec for the job in `helmfile.yaml` like:
+
+```yaml
+repositories:
+  - name: yourorg
+    url: https://yourorg.example.com/charts
+
+releases:
+  - name: dbmigrator
+    labels:
+      job: dbmigrator
+    chart: ./dbmigrator
+    # DB host, port, and connection opts for the environment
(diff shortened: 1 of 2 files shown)


examples/README.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
+You would then start a database migration job by executing:
+
+```console
+# Start a database migration for the prod environment
+$ RAILS_ENV=prod helmfile sync --selector job=dbmigrator
+
+# Tail log until you are satisfied
+$ stern -l job=dbmigrator
```
cmeury: May I suggest using standard `kubectl` or adding a reference to the [stern](https://github.com/wercker/stern) tool.

mumoshu: Sure thing! I'll make it just refer `kubectl logs -l`.

mumoshu:
@cmeury Thanks for the review!

PR #76, opened 2018-03-27
Upgrade container to alpine 3.7 and Helm 2.8.2 & check sha256

cmeury:
Small improvements to the container setup:

* Upgrade to latest alpine version
* Install Helm 2.8.2
* Check SHA256 of Helm
* Add a run command to the Makefile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,14 @@
-FROM alpine:3.4
+FROM alpine:3.7
 
+ENV HELM_VERSION=v2.8.2
+ENV HELM_LOCATION="https://kubernetes-helm.storage.googleapis.com"
+ENV HELM_FILENAME="helm-${HELM_VERSION}-linux-amd64.tar.gz"
+ENV HELM_SHA256="614b5ac79de4336b37c9b26d528c6f2b94ee6ccacb94b0f4b8d9583a8dd122d3"
 RUN apk add --update curl && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,8 +1,14 @@
...
+FROM alpine:3.7
 
+ENV HELM_VERSION=v2.8.2
+ENV HELM_LOCATION="https://kubernetes-helm.storage.googleapis.com"
+ENV HELM_FILENAME="helm-${HELM_VERSION}-linux-amd64.tar.gz"
+ENV HELM_SHA256="614b5ac79de4336b37c9b26d528c6f2b94ee6ccacb94b0f4b8d9583a8dd122d3"
 RUN apk add --update curl && \
-	curl -L https://kubernetes-helm.storage.googleapis.com/helm-v2.0.0-linux-amd64.tar.gz | \
-	tar zxf - && mv /linux-amd64/helm /usr/local/bin/ && rm -r /linux-amd64 && apk del curl
+    curl -L ${HELM_LOCATION}/${HELM_FILENAME} -o ${HELM_FILENAME} && \
```
andytom: Might be outside of the scope of this PR but you could use the busybox `wget` command that is bundled with the Alpine image and then there would be no need to add and remove curl.

cmeury: Done.

mumoshu:
Thanks for your supports, everyone ☺️

PR #81, opened 2018-04-03
Run simple integration test against minikube (local & CI)

cmeury:
## Purpose
When working on the code, I missed an automated integration test to verify helmfile still works as expected, but not on the unit level but actually installing charts. I tried to come up with a bash script (`integration/run.sh`) for my own testing purposes and wanted to share it here. I don't necessarily see this being merged, but some of you could give it a whirl, give some feedback or post alternative suggestions.

## Problems

Problems I see with the simplistic approach in this PR:

* ~~~Bash, instead of Golang. We're already steering a command-line tool using Go, why not use Go for the integration test as well.~~~ Makes sense because end-users likely use a shell to run helmfile.
* Self-baked testing setup. This makes it (hopefully) simple to reason and extend, but will quickly become unwieldy.
* ~~~Test setup and cleanup will quickly become time-consuming and error-prone, no real sand-boxing (maybe using unique namespaces or release names with timestamps?).~~~ Resolved
* ~~~The general approach towards integration testing may not be aligned with the maintainer's ideas. Do we even want something like that?~~~ Positive feedback

## Example Output

[asciicast](https://asciinema.org/a/nzgwFY8f7KQC6iwbyj6Gcqxyq)


mumoshu:
> Do we even want something like that?

Definitely! Thank you very much for starting the work.

> Bash, instead of Golang. We're already steering a command-line tool using Go, why not use Go for the integration test as well.

I'm happy with integration testing through a bash script because I, as helmfile user, usually run and test helmfile via a shell like bash. An integration test suite which just runs commands I use daily for manual testing is very straight-forward to understand.

If I could ask for more, ideally, we should run an ephemeral K8s cluster in Travis to run this integration test automatically in CI. [minikube helps that](https://blog.travis-ci.com/2017-10-26-running-kubernetes-on-travis-ci-with-minikube).

integration/run-tests.sh
```diff
@@ -0,0 +1,92 @@
...
+test "happypath: rollout of httpbin chart"
+$helmfile -f ${test_dir}/happypath.yaml sync
+wait_deploy_ready httpbin-httpbin
+curl --fail $(minikube service --url --namespace=${test_ns} httpbin-httpbin)/status/200 || fail "httpbin failed to return 200 OK"
+$helmfile -f ${test_dir}/happypath.yaml delete
+$helm status --namespace=${test_ns} httpbin &> /dev/null && fail "release should not exist anymore after a delete"
+pass "happypath.yaml"
+
+# If we reach here, no command exited with a non-zero exit code and we passed all tests.
+$kubectl delete namespace ${test_ns}
```
mumoshu: Almost nit but trap EXIT and run this cleanup inside the trap func would be a better approach because we then won't leave unnecessary resources when the script failed in the middle of process.

An example snippet(completely untested!):
```bash
on_exit() {
  $kubectl delete namespace ${test_ns}
}

trap 'on_exit' EXIT
```

Makefile
```diff
@@ -21,10 +21,18 @@ test:
 	go test -v ${PKGS} -cover -race -p=1
 .PHONY: test
 
+integration:
+	bash ./integration/run-tests.sh
```
mumoshu: Nit but I'd slightly prefer `./test/integration/run`

integration/run-tests.sh
```diff
@@ -0,0 +1,92 @@
...
+    fail "Could not connect to Docker daemon"
+fi
+info "Connection to minikube docker daemon successful"
+
+if ! hash gox 1>/dev/null; then
+    fail "Gox needs to be installed. Run: make tools"
+fi
+
+if ! hash helm 1>/dev/null; then
+    fail "Helm needs to be installed."
```
mumoshu: Very helpful validations for test env. Awesome!

cmeury:
Thanks for that generous review, I have incorporated your feedback and also polished the setup a bit more.

cmeury:
And it's running on CircleCI.

mumoshu:
Just finished reviewing your work. YOU ROCK. Great job 👍

PR #82, opened 2018-04-03
delete decrypted secret on sync command

sstarcher:
due to not passing around the pointer into channels and through the functions we were not getting back the generatedValues results.  This fixes that.


sstarcher:
@mumoshu fully agreed relying on the mutability is not the preferred mechanism, but we need some global state to keep track of files created.  We could use a general global generatedFiles vs using it at the release level.

PR #77, opened 2018-03-31
Update dependencies of local charts when running 'sync'

cmeury:
As discussed in #26.

Issue #26, opened 2018-01-09
Missing command - dependency update


xeor:
I've started using `helmfile`, and after a couple of days of use, I am missing a way to do `helm dependency update`. Should it be a part of helmfile? And included in `sync`

mumoshu:
@xeor Sounds reasonable!

But out of curiosity, would you mind sharing me how do you separate uses between `helm` and `helmfile` when you are able to use the [helm's umbrella chart pattern](https://github.com/kubernetes/helm/blob/master/docs/charts_tips_and_tricks.md#complex-charts-with-many-dependencies) like you do?

Can't you just transform your helmfile into a single umbrella helm chart so that perhaps you can just use `helm status` against it?

xeor:
This was only for my personal charts.. The umbrella is not that big, but I got multiple "umbrellacharts".
I don't really like to put all the variables inside the helmfile. The helmfile for me is more or less a collection of charts. If it's an upstream chart, and it got just a couple of values to change. I might put the values in the helmfile. Otherwise, I like not to..

mumoshu:
@xeor Thanks for the clarification!

I've iterated on your explanation and my current/possible use-cases of helmfile.
And now I believe your use-case is ABSOLUTELY valid!

Would it be good for you if `helmfile sync` just runs `helm dependency update` before updating the releases? Would there be any bad side-effect of doing so?

mumoshu:
cc/ @sstarcher Would you also be happy with this change?

mumoshu:
@xeor FWIW, I also believe a small enhancement to helmfile like #28(which would also support #49) would allow helmfile to support what we can't easily do with umbrella charts. Appreciate your comments on those two issues, too!

sstarcher:
If you wanted to include it I don't think you would want it part of sync and it should only be run in charts with local file paths.

In the grand scheme of things using local charts that are not versioned is a bad practice, but I'm not going to stop you :)

If you go this route with adding a command that works against local charts are we going to end up adding support for every helm command that runs against local charts?

mumoshu:
@sstarcher Thx for the response!

How about just improving `helmfile sync` to also run `helm dep update` when the release's `chart:` references a local chart?

I'm curious about the usefulness of only running `dep update` or even `repo update`(already implemented as `helmfile repos` but I have never got to use that myself) through `helmfile`.

cmeury:
@xeor Would you mind giving this a test run?

mumoshu:
[...]

PR #85, opened 2018-04-04
Sync command: change the way we apply args

fsilberstein:
As discussed in #26, the previous PR #77 prevent applying some args like `"--wait"` for the sync command.

The purpose of this PR is to apply args flag only to sync command and not `UpdateDeps`


mumoshu:
@fsilberstein Thank you so much for the good catch! LGTM.

Issue #86, opened 2018-04-05
Env var interpolation in name and namespace


danielcb:
Hi,
since it was done in #36 for repositories: Would it be possible to extend the interpolation to name and namespaces for automated deployments?

For namespace it looks like it can be easily extended in flagsForRelease(). For name it's currently scattered between SyncReleases and DiffReleases.. but generally just wondering: Any reason why it's not (yet) possible / why it should not be possible for those two fields?

danielcb:
Added a PR to illustrate what I was trying to say/ask.

manics:
Could this be extended to other (or all?) fields? I'm currently using YAML anchors and references to deploy the same chart with multiple configurations, e.g.
```
RELEASE_COMMON: &RELEASE_COMMON
  chart: jupyterhub/jupyterhub
  version: v0.7-4104389

releases:

  - <<: *RELEASE_COMMON
    name: jupyterhub-1
    namespace: jupyterhub-1
    ...
  - <<: *RELEASE_COMMON
    name: jupyterhub-2
    namespace: jupyterhub-2
    ...
```
But interpolating chart and version would be more readable.

danielcb:
Yes. Which fields? Chart/version?

manics:
Those are the two I'm interested in, but for consistency perhaps it's easier to say which fields shouldn't be interpolated?

mumoshu:
@manics @danielcb I rather want to start by limiting helmfile's dynamism to be necessary and sufficient for intended use-cases, so that hopefully helmfile.yml doesn't get mess.
So, would you mind sharing us your exact use-cases of wanting templates everywhere for further discussion? Thx!

mumoshu:
Anyway, closing this as the original issue is resolved via #90 and #87. Please feel free to submit separate feature requests for another places you want templating!

cmeury:
Somehow I think using YAML anchors is not that bad. Just using the capabilities of the format.

But I had a similar request in #59, using a `defaults` section. How do you think these two compare?

PR #87, opened 2018-04-05
Add envvar interpolation for names and namespaces

danielcb:
As proposed in #86.

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -165,6 +165,11 @@ func (state *HelmState) SyncReleases(helm helmexec.Interface, additionalValues [
 	for w := 1; w <= workerLimit; w++ {
 		go func() {
 			for release := range jobQueue {
+				nameRendered, err := renderTemplateString(release.Name)
+				if err != nil {
+					errQueue <- err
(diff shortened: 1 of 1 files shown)


cmeury:
Seems reasonable to me. No specific reason why those fields should not be interpolatable(?) as well. I just could not think of a good reason *for* it.

danielcb:
Cool. Yep, as said in the issue, we would like to use it in our automated deployment chain. And since everything else is already open for templating via envvars these changes would avoid us having to add an extra layer of templating around helmfile.

Let me know if there's anything missing/wrong in the PR.

mumoshu:
Merging this following #90. But beware that we shouldn't lose benefits of declarative nature provided helmfile.yml by making it too dynamic! And I'm still exploring about how the scope of helmfile should be.

But anyway, this doesn't affect existing users so let's go ahead and see how this works in the wild!
I appreciate it if you could share us your experiences after using this feature seriously in prod 🙏

danielcb:
Will do. Thanks!

Issue #88, opened 2018-04-05
env in 'set' does not interpolate unless quoted


maratoid:
this works:

```
set:
      - name: vault.replicas 
        value: "{{ env \"NUM_REPLICAS\" }}"
```

This does not:

```
set:
      - name: vault.replicas 
        value: {{ env "NUM_REPLICAS" }}
```

second example results in:

```
exec: helm upgrade --install vault ... --namespace mynamespace --values .../vault.yaml --values .../secrets.yaml.dec --set vault.replicas=
```

cmeury:
`{` and `}` have a special meaning in the YAML format, see here: http://yaml.org/spec/1.2/spec.html#id2790832

So, your second example without quotes seems to be invalid YAML (checked with a [YAML validator](https://codebeautify.org/yaml-validator)).

maratoid:
Fair enough. In that case the readme's should be fixed up. Also, first example should probably result in error (as described in the 'configuration syntax' section of the readme).

As is helmfile accepts the input and just sets the relevant key to nothing.

cmeury:
> first example should probably result in error

I assume you mean the second example?

Yeah, agreed, that should be handled more robustly.

maratoid:
yes, my apologies. edited.

mumoshu:
@maratoid Thanks for the feedback and the report 👍 

Oh, I have never realized that the second example doesn't result in any error!
It should definitely error out.
Would you mind submitting a PR or an dedicated github issue for that?

maratoid:
@mumoshu I'll try my best, thanks :)

PR #89, opened 2018-04-06
Add quotes to envvar example in readme

danielcb:
Fix readme file. Envvar interpolation is not working without quotes. See #88.


mumoshu:
Good catch. Thanks again for your contribution 👍

Issue #94, opened 2018-04-06
Race condition in DiffReleases


danielcb:
There's a race condition in DiffReleases (multiple goroutines trying to concurrently write to the errs slice).

go run -race output:
```go
err: template: stringTemplate:1: function "ev" not defined
==================
WARNING: DATA RACE
Read at 0x00c4200f2000 by goroutine 14:
  github.com/roboll/helmfile/state.(*HelmState).DiffReleases.func1()
      /home/danielb/go/src/github.com/roboll/helmfile/state/state.go:263 +0x603

Previous write at 0x00c4200f2000 by goroutine 10:
  github.com/roboll/helmfile/state.(*HelmState).DiffReleases.func1()
      /home/danielb/go/src/github.com/roboll/helmfile/state/state.go:249 +0x940
  text/template/parse.(*Tree).textOrAction()
      /usr/local/go/src/text/template/parse/parse.go:348 +0x2cb
  text/template/parse.(*Tree).parse()
      /usr/local/go/src/text/template/parse/parse.go:294 +0x41b
  text/template/parse.(*Tree).Parse()
      /usr/local/go/src/text/template/parse/parse.go:233 +0x33c
  text/template/parse.Parse()
      /usr/local/go/src/text/template/parse/parse.go:55 +0x13e
  text/template.(*Template).Parse()
      /usr/local/go/src/text/template/template.go:198 +0x214
  github.com/roboll/helmfile/state.renderTemplateString()
      /home/danielb/go/src/github.com/roboll/helmfile/state/state.go:111 +0x63
  github.com/roboll/helmfile/state.flagsForRelease()
      /home/danielb/go/src/github.com/roboll/helmfile/state/state.go:439 +0x180d
  github.com/roboll/helmfile/state.(*HelmState).DiffReleases.func1()
      /home/danielb/go/src/github.com/roboll/helmfile/state/state.go:247 +0xd7

Goroutine 14 (running) created at:
  github.com/roboll/helmfile/state.(*HelmState).DiffReleases()
      /home/danielb/go/src/github.com/roboll/helmfile/state/state.go:244 +0x1b4
  main.main.func3()
      /home/danielb/go/src/github.com/roboll/helmfile/main.go:160 +0x328
  runtime.call32()
      /usr/local/go/src/runtime/asm_amd64.s:573 +0x3a
  reflect.Value.Call()
      /usr/local/go/src/reflect/value.go:308 +0xc0
[...]

PR #95, opened 2018-04-06
Fix race condition. Add concurrency option similiar to DiffReleases

danielcb:
This fixes the race condition mentioned in #94. I tried to implement it similar to DiffReleases, but used wait groups instead of the done channel. Also I added the same concurrency option that DiffReleases uses. Default is as much workers as releases, same as in DiffReleases.

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -230,45 +230,75 @@ func (state *HelmState) SyncReleases(helm helmexec.Interface, additionalValues [
 	return nil
 }
 
-func (state *HelmState) DiffReleases(helm helmexec.Interface, additionalValues []string) []error {
-	var wg sync.WaitGroup
+func (state *HelmState) DiffReleases(helm helmexec.Interface, additionalValues []string, workerLimit int) []error {
+	var wgRelease sync.WaitGroup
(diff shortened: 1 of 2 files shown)


state/state.go
```diff
@@ -230,45 +230,75 @@ func (state *HelmState) SyncReleases(helm helmexec.Interface, additionalValues [
...
+				wgRelease.Done()
 			}
-			wg.Done()
-		}(&wg, release)
+		}()
 	}
-	wg.Wait()
+	wgError.Add(1)
+	go func() {
+		for err := range errQueue {
```
mumoshu: Sorry if I missed something here but how this doesn't end up waiting for infinite num of errors?
I see `close(errQueue)` nowhere. When does `errQueue` gets closed so that this range loop finishes?

danielcb: Hi, not a problem. Have a look at the next few lines. The loop over the errQueue happens in a goroutine. The part beginning on line 297 first closes and waits for the release jobs and then does the same for the error jobs.

mumoshu: Thx for the clarification! You are certainly correct. I somehow missed that.

Issue #52, opened 2018-03-12
Note about production users


mumoshu:
Similar to [the one for kubernetes-on-aws](https://github.com/hjacobs/kubernetes-on-aws-users).

It will help users and maintainers have another aspect of confidence in their project and tooling, while allowing maintainers to try to contact prod users in case it seems necessary(like before making a breaking change, scope-creep, etc.

roboll:
I’m fine with this if people want to volunteer their info, but I’m not sure if there are other implications. 

There should certainly be no sponsorship or anything like that, no money changing hands.

mumoshu:
> no money changing hands

I have never heard of this great word! Yes, that's definitely what I expect for a healthy OSS object 👍

cmeury:
Information from my side, which can be used if anyone is in the mood for a PR.

* **Link:** https://www.ricardo.ch/
* **Description:** ricardo.ch is the largest online marketplace in Switzerland - the well-known service provides customers the best possible e-commerce experience since 1999.
* **Logo:** https://style-6a38.kxcdn.com/12.2.4/lib/images/ric-logo__core.svg

cmeury:
@mumoshu This one can possibly be closed.

mumoshu:
@cmeury Yes. Thanks for reminding me 👍

PR #98, opened 2018-04-07
templating the entire helmfile and adding some documentation

sstarcher:


state/state.go
```diff
@@ -129,16 +133,12 @@ func (state *HelmState) applyDefaultsTo(spec *ReleaseSpec) {
 	}
 }
 
+// SyncRepos will update the given helm releases
 func (state *HelmState) SyncRepos(helm helmexec.Interface) []error {
 	errs := []error{}
 
 	for _, repo := range state.Repositories {
-		url, err := renderTemplateString(repo.URL)
```
mumoshu: Could you try to preserve backward-compatibility here? E.g. try to leave legacy templating as-is.

I believe existing template expressions fail due to syntax error around escaped quotes so we need to somehow resolve that…

sstarcher: When templating the entire file this will cause it to error
```
      - name: ingress.enabled
        value: "{{ env \"HOME\" }}"
```

And after the entire file is templated no templates will be left for the `renderTemplateString` to function against.

I believe the best course of action is to release a major version and specify migration steps.  It would be nice if helmfile would transition to a semantic versioning schema.

mumoshu: Thanks as always for the thoughtful feedback!

Definitely. That's what I meant where we break backward-compatibility in this PR. I'd like to care existing prod helmfile users not to be forced to fix every helmfile.yaml due to backward-imcompatibilities introduced by a helmfile upgrade.

On the other hand, I want to merge PRs like this quickly for our productivity :)

How can we accomplish both?

And yes, semver and a major version
bump for helmfile would be the future, but that's not what we can do immeadiately! Also, it does't necessary mean we can easily drop existing users with short notice.

So, alternatively, can you introduce a apiVersion key to helmfile.yaml so that we can version current and the default to apiVersion to v1alpha1?

Then, the new templating semantics you introduce via this PR could be enabled via `apiVersion: v1beta1`.

sstarcher: I'm not prepared to implement that complexity into the helmfile logic.  I would propose we switch our versioning scheme to semantic version and release a 0.12.0 that contains a breaking change and in the change log we document that upgrading requires making modifications.

I have no issue with someone else implementing the versioning you are requesting, but the migration is fairly simple for anyone wanting to use the updated features, but I don't have the bandwidth to do so.

mumoshu: @sstarcher Thanks for the clarification and the feedback. Noted. I'll do my best as a maintainer 😉 

Regarding my last suggestion of introducing `apiVersion` to helmfile, it won't work as I expect because theoretically the vanilla helmfile.yml before applying the entire-file templating can be an invalid yaml.

mumoshu: @sstarcher @cmeury Would you mind if I introduced a new convention to helmfile that you MUST name your helmfile.yml suffixed with `.tmpl` so that it looks like `helmfile.yml.tmpl` to invoke the new entire-file templating? Slightly off-topic but it would also reflect the fact that the former helmfile.yml is a valid yaml but the new helmfile.yml introduced after this PR is not.

Very sorry for rushing with many questions. I need feedbacks from experienced helm/helmfile users to shape how the next helmfile should look like 🙏

mumoshu:
Hi @gtaylor, would u be ok with rewriting all your helmfile.yaml's to replace `{{ env \"FOO\" }}` to `{{ "FOO" }}` after this change is merged? Just asking. I'm just wondering about how everyone feels it.

state/state.go
cmeury: Having a different file extensions seems weird to me, I'd rather handle the "versioning" in the application logic. Could we detect the escaped quotes, warn about the upcoming breaking change and revert to the previous functionality?
Introducing a different version of helmfile seems way too "big" for such a minor change. We're not fundamentally altering the hierarchy of keys.

mumoshu: > We're not fundamentally altering the hierarchy of keys.

That's true. I'm still wondering though, if the template syntax is a significant part of the whole helmfile.yml syntax which justifies versioning. But maybe your sentiment is that it is still a minor change regardless of my point? (Just trying to sync up with you

manics: > Would you mind if I introduced a new convention to helmfile that you MUST name your helmfile.yml suffixed with .tmpl

If the templated file is invalid yaml I think this is a good idea, otherwise it breaks people's expectations of what the `.yml` file extension means.

cmeury: But isn't Helm doing the same thing? Its templates are not valid YAML but they're still using the `.yaml` extension. Irrespective of whether that's a good or bad thing, I would argue we should follow Helm's practice to meet Helm user's expectations as much as possible.

manics: Good point!

mumoshu: Good point 😄

sstarcher:
@mumoshu I think this is a good time to talk about what people currently expect from the stability of the helmfile format and how far we should go for backwards compatibility and for how long.  We currently support multi deprecated options.  This will continue to build on technical debt unless we remove them overtime.

mumoshu:
@sstarcher Definitely. I agree with your concerns on technical debt. But I don't actually have good communication channels with prod users but @gtaylor. Also, I wished existing prod users will chime-in to #52 but I had no luck so far.

@roboll Could you share me list of existing prod users of helmfile for contact?

I'll try to keep backward-compatibility whereas possible, as long as we have some hands to do so, as a maintainer. But it doesn't necessarily mean I can do everything 😉 

So,  unless we could reach someone, the next best thing we can practically do is introduce a semantic versioning to helmfile, as you had suggested, as early as possible, and prevent further confusions like this!

mumoshu:
Hoping my attempt to resolve the conflict between this and the master succeeded 🤔

mumoshu:
Thank you so much for all your support!
Let's keep discussion in relevant issues.

PR #100, opened 2018-04-09
Redo release automation via CircleCI

### LINKED ISSUE #91: CircleCI job "deployment" might not work
In PR #81 we left out the `deployment` job in `.circleci/config.yml`. It is neither included in the `jobs` section nor is it contained in a [workflow](https://circleci.com/docs/2.0/workflows/). I suspect the command will currently not work. We need to test this and potentially correct this.

https://github.com/roboll/helmfile/blob/master/.circleci/config.yml#L76

mumoshu:
Resolves #91

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,13 +73,19 @@ jobs:
             make integration
 
 # GITHUB_TOKEN env var must be setup in circleci console
-deployment:
+
   release:
-    tag: /v.*/
-    commands:
-      - docker login -e="." -u="$DOCKER_USER" -p="$DOCKER_PASS" quay.io
-      - cd "$WORK" && make tools
-      - cd "$WORK" && BUILD_URL="$CIRCLE_BUILD_URL" make push release
+    docker:
+      - image: circleci/golang:1.7
+    working_directory: /go/src/github.com/roboll/helmfile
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -94,4 +100,10 @@ workflows:
             branches:
               only:
                 - master
-                - /pull.*/
\ No newline at end of file
+                - /pull.*/
+      - release:
+          filters:
+            branches:
+              ignore: /.*/
```
mumoshu: It looked unnecessary at glance but then turned out necessary - I had to add this branch ignore pattern to actually avoid `release` jobs being built against untagged commits from any branch.

Makefile
```diff
@@ -1,4 +1,4 @@
-ORG     := $(shell basename $(realpath ..))
+ORG     ?= $(shell basename $(realpath ..))
```
mumoshu: Set `ORG` env var in your CircleCI build environment so that you can actually test this workflow on your GitHub/Quay/CircleCI account.

mumoshu:
Verified this to actually work on my fork: https://github.com/mumoshu/helmfile/releases

mumoshu:
Merging this as required to cut the next release.

PR #106, opened 2018-04-11
Update README not to confuse new-comers due to breaking change in master

mumoshu:


mumoshu:
@sstarcher Thanks for the review 👍

PR #110, opened 2018-04-17
support selector that is empty

### LINKED ISSUE #99: Label syntax for missing values
Lets say I have 10 charts.  1 chart has a label `stage: pre` and another has the label `stage: post`

When I do
`helmfile -l stage=pre sync`  I get the expected 1 chart from pre and the same for post.

When I do
`helmfile -l stage!=pre,stage!=post sync` I would expect to get the other 8 charts, but I get 0 charts as the charts without the label are not selectable.

When I do 
`helmfile -l stage!='' sync` I would expect to also get the other 8 charts, but I get an error.
```
2018/04/07 16:45:35 Malformed label: stage!=. Expected label in form k=v or k!=v
```

sstarcher:
Currently if I have the following charts
```yaml
  - name: pre
    labels:
      stage: pre
  - name: post
    labels:
      stage: post
  - name: other
  - name: other2
  - name: other3
```

`helmfile -l stage=pre` will install chart pre as expected
`helmfile -l stage!=` will fail with a malformed label selector
`helmfile -l stage!=pre` will install post

After these changes
`helmfile -l stage=pre` will install pre
`helmfile -l stage!=` will install other, other2, other3
`helmfile -l stage!=pre` will install post

Resolves #99

My specific use-case is to be able to do
```bash
helmfile -l stage=pre
helmfile -l stage!=
helmfile -l stage=post
```

Without needing to label all of the non pre/post charts with a stage.

### DIFF
--- a/state/release_filters.go
+++ b/state/release_filters.go
@@ -15,14 +15,16 @@ type ReleaseFilter interface {
 // LabelFilter matches a release with the given positive lables. Negative labels
 // invert the match for cases such as tier!=backend
 type LabelFilter struct {
-	positiveLabels map[string]string
-	negativeLabels map[string]string
+	positiveLabels [][]string
+	negativeLabels [][]string
 }
 
 // Match will match a release that has the same labels as the filter
 func (l LabelFilter) Match(r ReleaseSpec) bool {
 	if len(l.positiveLabels) > 0 {
-		for k, v := range l.positiveLabels {
+		for _, element := range l.positiveLabels {
+			k := element[0]
+			v := element[1]
 			if rVal, ok := r.Labels[k]; !ok {
 				return false
 			} else if rVal != v {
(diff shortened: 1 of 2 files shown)


mumoshu:
@sstarcher Hi. Thanks as always for your contribution!

> helmfile -l stage!= will install other, other2, other3

I would prefer it to install everything for consistency with `kubectl`'s behavior on label selector.

Even with my preference applied, your use-case can be achieved like:

```
helmfile -l stage=pre
helmfile -l stage= # This installs everything but stage=pre and stage=post
helmfile -l stage=post
```

sstarcher:
@mumoshu added an additional test case to specifically test for.  Due to the usage of a map anything that was `stage=blah,stage=other` would not work as it would overwrite the same key.

```
helmfile -l stage=pre
# This installs everything but stage=pre and stage=post
helmfile -l stage!=pre,stage!=post
helmfile -l stage=post
```

mumoshu:
@sstarcher LGTM. Thanks again for your awesome work 🎉

Issue #112, opened 2018-04-21
Docker image does not have root CAs


gavvvr:
If you run Docker image and try to `helm init`, there is an error:

> Error: Looks like "https://kubernetes-charts.storage.googleapis.com" is not a valid chart repository or cannot be reached: Get https://kubernetes-charts.storage.googleapis.com/index.yaml: x509: failed to load system roots and no roots provided

Command to reproduce: docker run --rm -it --entrypoint=sh --entrypoint=helm quay.io/roboll/helmfile:v0.12.0 init

mumoshu:
Closing this as resolved via #113. Thanks for the report and the fix @gavvvr!

PR #113, opened 2018-04-21
Add root CAs to docker image

gavvvr:
See #112


gavvvr:
How to trigger a new build for quay.io? I thought it will be available as soon as the code is merged

PR #115, opened 2018-04-24
support relative path for is local check

### LINKED ISSUE #108: local chart detection assume relative path
Currently we detect local charts for dependency updates using a relative path regex.  It instead might be better to just check for file existence to support fully qualified paths.

https://github.com/roboll/helmfile/commit/1af8743bb6cdaca8b94db1a9009357df5ab27ac5#diff-f6be34387a141f652301c16cf7a5e36cR363

sstarcher:
Fixes #108

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -399,15 +399,16 @@ func (state *HelmState) UpdateDeps(helm helmexec.Interface) []error {
 // 	 be constructed relative to the `base path`.
 // - Everything else is assumed to be an absolute path or an actual <repository>/<chart> reference.
 func normalizeChart(basePath, chart string) string {
-	if !isLocalChart(chart) {
+	regex, _ := regexp.Compile("^[.]?./")
+	if !regex.MatchString(chart) {
 		return chart
 	}
 	return filepath.Join(basePath, chart)
 }
 
 func isLocalChart(chart string) bool {
-	regex, _ := regexp.Compile("^[.]?./")
-	return regex.MatchString(chart)
+	_, err := os.Stat(chart)
+	return err == nil
 }
 
 func flagsForRelease(helm helmexec.Interface, basePath string, release *ReleaseSpec) ([]string, error) {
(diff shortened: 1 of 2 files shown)


state/state.go
```diff
@@ -399,15 +399,16 @@ func (state *HelmState) UpdateDeps(helm helmexec.Interface) []error {
...
+	if !regex.MatchString(chart) {
 		return chart
 	}
 	return filepath.Join(basePath, chart)
 }
 
 func isLocalChart(chart string) bool {
-	regex, _ := regexp.Compile("^[.]?./")
-	return regex.MatchString(chart)
+	_, err := os.Stat(chart)
```
mumoshu: Just curious but doesn't this result in `chart: stable/mysql` without `./` to be considered a local chart when a directory named `stable/mysql` exists? My example isn't practical but it still seems slightly better to leave the regex matching along with the new file existence check you've added.

sstarcher: I don't really agree as the regex matching is limited and does not support all situations.  Like fully qualified paths.

mumoshu: That's also true!

Certainly I'm not aware of every case. But anyway, do we need to support fully qualified paths?

To me, it seems like the opposite of helmfile's primary motivation of declaratility :thinking:

sstarcher: In that case why support in local files?  My only comment on this is a file existence check is less error prone compared to the regex.  Unless we used a protocol like...

`chart: stable/blah` means a chart in a repository
`chart: file://stable/blah` means a local file.

mumoshu: The protocol is interesting!

Anyway, local charts are declarative enough to me as I can normally spot
which version of local chart is used by seeing e.g. git commit's id. I can't expect that for absolute paths, right?

The existence check is definitely nice! I just thought that, assuming we'd need support for relative but absolute path, we can have both, or no necessity to avoid - regex + existence check :thinking:

sstarcher: absolute paths are certainly just as viable as relative paths if not more so.  If things are running in a docker container and you are mounting a directory or if you are using a path such as `/opt/blah`

I would lean toward the removal of the regex and using the file existance.

mumoshu: Makes sense! Thx.

mumoshu:
＠sstarcher Thanks as always for your contribution!

PR #124, opened 2018-04-27
build consistent statically linked binary for docker image

### LINKED ISSUE #123: Issues with v.0.13.0 - Dynamically linked binary in alpine
v0.12.0 runs fine:
```
$ docker run --rm -it quay.io/roboll/helmfile:v0.12.0 helmfile
NAME:
   helmfile -

USAGE:
   helmfile [global options] command [command options] [arguments...]

VERSION:
   v0.12.0

COMMANDS:
     repos   sync repositories from state file (helm repo add && helm repo update)
     charts  sync charts from state file (helm upgrade --install)
     diff    diff charts from state file against env (helm diff)
     sync    sync all resources from state file (repos, charts and local chart deps)
     delete  delete charts from state file (helm delete)

GLOBAL OPTIONS:
   --file FILE, -f FILE         load config from FILE (default: "helmfile.yaml")
   --quiet, -q                  silence output
   --kube-context value         Set kubectl context. Uses current context by default
   --namespace value, -n value  Set namespace. Uses the namespace set in the context by default
   --selector value, -l value   Only run using the releases that match labels. Labels can take the form of foo=bar or foo!=bar.
                                A release must match all labels in a group in order to be used. Multiple groups can be specified at once.
                                --selector tier=frontend,tier!=proxy --selector tier=backend. Will match all frontend, non-proxy releases AND all backend releases.
                                The name of a release can be used as a label. --selector name=myrelease
   --help, -h                   show help
   --version, -v                print the version
```

v0.13.0 does not:
```
$ docker run --rm -it quay.io/roboll/helmfile:v0.13.0 helmfile
standard_init_linux.go:190: exec user process caused "no such file or directory"
```

v0.12.0 (statically linked):
```
docker run --rm -it quay.io/roboll/helmfile:v0.12.0 sh -c 'apk add file -q --update && file /usr/local/bin/helmfile'
/usr/local/bin/helmfile: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, with debug_info, not stripped
```

v0.13.0 (dynamically linked):
```
docker run --rm -it quay.io/roboll/helmfile:v0.13.0 sh -c 'apk add file -q --update && file /usr/local/bin/helmfile'
/usr/local/bin/helmfile: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, with debug_info, not stripped
```

cmeury:
* use golang 1.10.1 images everywhere
* introduce a `static-linux` target to build a statically linked 64-bit binary only for the docker image
* use a multi-stage dockerfile for a consistent build environment

This resolves #123.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
     working_directory: /go/src/github.com/roboll/helmfile
     steps:
       - checkout
-      - run: make static-linux
+      - run: make build
       - persist_to_workspace:
           root: /go/src/github.com/roboll/helmfile
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,11 +4,11 @@ jobs:
 
   build:
     docker:
-      - image: circleci/golang:1.10
+      - image: circleci/golang:1.10.1
     working_directory: /go/src/github.com/roboll/helmfile
     steps:
       - checkout
-      - run: make build
+      - run: make static-linux
       - persist_to_workspace:
           root: /go/src/github.com/roboll/helmfile
           paths:
@@ -16,7 +16,7 @@ jobs:
 
   test:
     docker:
-      - image: circleci/golang:1.10
+      - image: circleci/golang:1.10.1
     working_directory: /go/src/github.com/roboll/helmfile
     steps:
       - checkout
@@ -76,7 +76,7 @@ jobs:
 
   release:
     docker:
-      - image: circleci/golang:1.10
+      - image: circleci/golang:1.10.1
     working_directory: /go/src/github.com/roboll/helmfile
     steps:
     - checkout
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -76,15 +76,14 @@ jobs:
...
-      - image: circleci/golang:1.10
+      - image: circleci/golang:1.10.1
     working_directory: /go/src/github.com/roboll/helmfile
     steps:
     - checkout
     - setup_remote_docker
     - run:
         command: |
           docker login -u="$DOCKER_USER" -p="$DOCKER_PASS" quay.io
-          make tools
```
mumoshu: `make tools` seems to be still required for `make release`. It depends on `ghr` for managing gh releases.

mumoshu:
@cmeury Thanks as always for your contribution! Just a single comment. LGTM overall 👍 
Probably I can make time to address it on my own.

.circleci/config.yml
cmeury: Good point: https://github.com/roboll/helmfile/pull/129

PR #125, opened 2018-04-27
add users file to track production use

cmeury:
Created a Google Doc and a markdown file to track production users, resolving #52.


mumoshu:
This is super helpful! Thanks as always for your support 🙏

Issue #56, opened 2018-03-16
Missing `releases[].chart` results in an unintuitive error


mumoshu:
As seen in https://github.com/roboll/helmfile/issues/55#issuecomment-373735128

We should add a validation for `releases[].chart`!

cmeury:
There's a new version of `gopkg.in/yaml` wich supports strict marshalling. I thought this was a nice approach, but it goes much further than simply checking for the name of `releases[].chart`. Let me know what you think in #126.

mumoshu:
Sounds good! AFAIK, go-yaml v2 has been working fine for 2 years for kube-aws. So I have no doubt on its reliability.

cmeury:
Error is now:

```
error: yaml: unmarshal errors:
  line 4: field releases not found in type state.ReleaseSpec
```

Is that clear enough, or should we identify the error and report to the user in more direct language?

PR #126, opened 2018-04-27
fail fast on syntax error by strictly unmarshalling

cmeury:
* upgrade `gopkg.in/yaml` from `v1` to `v2` to have [`UnmarshalStrict()`](https://godoc.org/gopkg.in/yaml.v2#UnmarshalStrict) verify the YAML syntax
* use minikube `v0.25.2`, kubernetes `v1.9.0` and helm `v2.9.0` in integration tests
* [unrelated] adding simple happy path test for `HelmState.SyncReleases()`

Fixing #56


cmeury:
Had a lot of trouble with the minikube setup, in the end imitated [habitat's way of doing things](https://github.com/kinvolk/habitat-operator/blob/master/.circleci/config.yml). Will squash this if approved.

.circleci/config.yml
```diff
@@ -35,34 +35,30 @@ jobs:
...
-            sudo minikube start --vm-driver=none
-            sudo chown -R $USER.$USER ~/.minikube
-            sudo chown -R $USER.$USER ~/.kube
-            minikube update-context
+            curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/${K8S_VERSION}/bin/linux/amd64/kubectl
+            chmod +x kubectl && sudo mv kubectl /usr/local/bin/
+            curl -Lo minikube https://storage.googleapis.com/minikube/releases/${MINIKUBE_VERSION}/minikube-linux-amd64
+            chmod +x minikube && sudo mv minikube /usr/local/bin/
+            sudo minikube config set WantReportErrorPrompt false
+            sudo -E minikube start --vm-driver=none --bootstrapper=localkube --kubernetes-version=${K8S_VERSION}
```
mumoshu: Ah. I heard that recent minikube started to rely on kubeadm to bootstrap k8s by default. Perhaps you reverted it to the old behavior with `--bootstrapper=localkube`.

cmeury: @mumoshu Yes, it's a remnant of trying to run the tests with minikube `v0.26.1`. It looks like they switched to kubeadm in 0.26.x and require `systemd` to be installed on the host for `vm-driver=none` to work. The CircleCI image we use does not come with it, so I had to use a different bootstrapper, but I could not get it to work. So I reverted back to 0.25.x, which makes the localkube setting redundant. I removed it now, good catch.

mumoshu:
@cmeury Thank you so much for your contribution 🍻

Issue #103, opened 2018-04-10
Integration test flakiness?


mumoshu:
```
==> v1beta1/Deployment
NAME             DESIRED  CURRENT  UP-TO-DATE  AVAILABLE  AGE
httpbin-httpbin  1        1        1           0          0s

==> v1/Pod(related)
NAME                              READY  STATUS             RESTARTS  AGE
httpbin-httpbin-7777dd96b5-d2sxw  0/1    ContainerCreating  0         0s


Waiting for rollout to finish: 0 of 1 updated replicas are available...
deployment "httpbin-httpbin" successfully rolled out
curl: (7) Failed to connect to 10.142.15.211 port 31556: Connection refused
FAIL: (Bhttpbin failed to return 200 OK
namespace "helmfile-tests-20180409-132441" deleted
make: *** [integration] Error 1
Exited with code 2
```

mumoshu:
Didn't dig deep but the error can be seen at https://circleci.com/gh/roboll/helmfile/285?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

cmeury:
#128 might help to avoid this failure mode by retrying upon the initial curl exiting with code <> 0.

cmeury:
Another example: https://circleci.com/gh/roboll/helmfile/466

```
deployment "httpbin-httpbin" successfully rolled out
curl: (7) Failed to connect to 10.142.0.114 port 32753: Connection refused
FAIL: (Bhttpbin failed to return 200 OK
```

mumoshu:
Another 2 examples:

https://circleci.com/gh/roboll/helmfile/tree/pull%2F171

PR #128, opened 2018-04-27
retry curl to reduce risk of race condition

cmeury:
I am not sure why #103 happens, but I suspect it might be a rare occasion of the pod reporting healthy but the virtual network or service discovery has not fully been initialized. This PR retries the `curl` call to the service to minimize this risk. It's more of a prophylactic measure "into the dark", but it might help.


cmeury:
Conflict resolved, ready to be merged.

mumoshu:
LGTM. Thanks again for your continuous effort 🙏

Issue #59, opened 2018-03-22
Specify default values for values files


cmeury:
To simplify building a hierarchy of values files "outside" of the helmfile, it would be useful for deduplication to specify a set of default files to always load. On top of that, the key-value map of the releases could be made available for interpolation:

```yaml
...

defaults:
  values:
      - "./values/all/traefik.yaml"
      - "./values/{{ release.name }}/traefik.yaml"

releases:
  - name: traefik
    namespace: platform
    chart: stable/traefik
    version: 1.24.1
```  

What do you guys think about this?

mumoshu:
Hi @cmeury, thanks for the feedback.

Sounds interesting to me!

But if I could ask for more, would you mind providing a more large, or maybe serious example?
I couldn't figure out the exact benefit of the feature only by reading the example you've provided above.

Especially, `- "./values/{{ release.name }}/traefik.yaml"` seems to fit best under `releases[].values`.

cmeury:
Mistake on my end, there old have been no mention of traefik in the defaults. It's a means of deduplication:

```yaml
...

defaults:
  values:
      - "./values/default/{{ release.name }}.yaml"

      - "./values/{{ context.name }}/{{ release.name }}.yaml"

releases:
  - name: traefik
    namespace: platform
    chart: stable/traefik
    version: 1.24.1 
```  

With a directory structure:

```
values/default/traefik.yaml
values/minikube/traefik.yaml
values/dev/traefik.yaml
etc.
```

Or vice versa with having folders similar to the t
release name and files according to context/environment.

manics:
Following up from https://github.com/roboll/helmfile/issues/86#issuecomment-379218999
The use case I have is deploying the same chart multiple times with different configs. This can be achieved with YAML anchors and references, but I personally I think it detracts from the clarity of YAML:
```
RELEASE_COMMON: &RELEASE_COMMON
  chart: jupyterhub/jupyterhub
  version: v0.7-4104389

releases:

  - <<: *RELEASE_COMMON
    name: jupyterhub-1
    namespace: jupyterhub-1
    ...
  - <<: *RELEASE_COMMON
    name: jupyterhub-2
    namespace: jupyterhub-2
    ...
```
I like the suggestion here of `defaults` since it keeps everything inside the helmfile instead of relying on external env-vars.

mumoshu:
[...]

PR #135, opened 2018-05-04
Introduce template function `required_env`

AndreasBieber:
The `required_env` function allows you to declare a particular environment variable as required for template rendering.
If the environment variable is unset or empty, the template rendering will fail with an error message.

Example:

```yaml
releases:
  - name: {{ required_env "NAME" }}-vault
    namespace: {{ required_env "NAME" }}
    chart: roboll/vault-secret-manager
    set:
      - name: proxy.domain
        value: {{ required_env "PLATFORM_ID" }}.my-domain.com
```

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -107,7 +107,21 @@ func readFromYaml(content []byte, file string) (*HelmState, error) {
 }
 
 func stringTemplate() *template.Template {
-	return template.New("stringTemplate").Funcs(sprig.TxtFuncMap())
+	funcMap := sprig.TxtFuncMap()
+	alterFuncMap(&funcMap)
+	return template.New("stringTemplate").Funcs(funcMap)
+}
+
+func alterFuncMap(funcMap *template.FuncMap) {
+	(*funcMap)["requiredEnv"] = getRequiredEnv
+}
+
+func getRequiredEnv(name string) (string, error) {
+	if val, exists := os.LookupEnv(name); exists && len(val) > 0 {
+		return val, nil
+	}
+
+	return "", fmt.Errorf("required env var `%s` is not set", name)
 }
 
 func renderTemplateString(s string) (string, error) {
--- a/README.md
+++ b/README.md
@@ -52,9 +52,11 @@ releases:
       - name: address
         value: https://vault.example.com
       - name: db.password
-        value: {{ env "DB_PASSWORD" }}    # value taken from environment variable. Quotes are necessary. Will throw an error if the environment variable is not set. $DB_PASSWORD needs to be set in the calling environment ex: export DB_PASSWORD='password1'
+        value: {{ requiredEnv "DB_PASSWORD" }}    # value taken from environment variable. Quotes are necessary. Will throw an error if the environment variable is not set. $DB_PASSWORD needs to be set in the calling environment ex: export DB_PASSWORD='password1'
       - name: proxy.domain
-        value: {{ env "PLATFORM_ID" }}.my-domain.com # Interpolate environment variable with a fixed string
+        value: {{ requiredEnv "PLATFORM_ID" }}.my-domain.com # Interpolate environment variable with a fixed string
+      - name: proxy.scheme
+        value: {{ env "SCHEME" | default "https" }}
 
   # Local chart example
   - name: grafana                            # name of this release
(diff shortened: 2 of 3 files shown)


mumoshu:
@sstarcher Hey. I guess we intentionally changed the `env`'s behavior, not to fail when the env var isn't set in https://github.com/roboll/helmfile/commit/71fa693997ed56108272219c34fad0427f8bdc97#diff-f6be34387a141f652301c16cf7a5e36cL92?
How about merging this PR with changing the proposed func name to `env`, to override the one [provided by sprig](http://masterminds.github.io/sprig/os.html)?

AndreasBieber:
I think it's good to have both. One for optional envs and one for required. What do you think?

sstarcher:
Any reason to not do what helm does and have a generic require function?

mumoshu:
@sstarcher That's a good point! But the answer is yes for me at the moment.

helm doesn't actually have `env`, so I believe we don't have a momentum to conform helmfile to helm, in this specific case.

If we had introduced an alternative of `env`, something like a values file for helmfile itself as per #59 , it would make sense for me to do something like `{{ required "Bar is required!" .HelmfileValues.Foo.Bar }}` in your `helmfile.yml`.

However, for now, `{{ required "Bar is required!" ( env "BAR" ) }}` as you might have suggested, seems just unintuitive compared to `{{ env "BAR" }}`(w/ previous behavior of failing on missing value) or even `{{ required_env "BAR" }}`.

WDYT?

AndreasBieber:
This was my first thought on implementing a `required` func like from helm. But i don't like the fact, that i have to define an error message for each `required`.

sstarcher:
I would rather go with `required_env` so we don't deviate from helms default usage of `env`.  It might make sense to see if `sprig` would take an addition so we could get it both into helm and sprig, but they may not be interested.

mumoshu:
@AndreasBieber @sstarcher Thanks both for the feedbacks!

@sstarcher Maybe I'm missing something but AFAIK helm doesn't allow `env` at all, not to leak tiller env vars to helm templates. So we don't need to worry about the deviation in this specific case?

state/state.go
```diff
@@ -107,7 +107,21 @@ func readFromYaml(content []byte, file string) (*HelmState, error) {
...
 
 func stringTemplate() *template.Template {
-	return template.New("stringTemplate").Funcs(sprig.TxtFuncMap())
+	funcMap := sprig.TxtFuncMap()
+	alterFuncMap(&funcMap)
+	return template.New("stringTemplate").Funcs(funcMap)
+}
+
+func alterFuncMap(funcMap *template.FuncMap) {
+	(*funcMap)["required_env"] = requiredEnv
```
mumoshu: According the discussion made in this PR, I'd like to make it assigned to the key`"env"`

sstarcher: sounds fine, but lets make sure we document it as a difference from sprig

mumoshu: Good point! That's definitely helpful and valuable.

AndreasBieber:
I would prefer `required_env` because otherwise strigs `env` would have another behavior than in helmfile. But if you want to stick with `env` than i'll change it.

mumoshu:
@AndreasBieber Thanks. That's also good point! I take your feedback as an user's  - which should be considered very carefully.

For me, maybe it is just that:

1. I still doubt if users read the doc for sprig's own `env` when `helmfile` provided `env` itself.
2. The `env` without the validation sounds just confusing. Why do we need it? Can't we just introduce `required_env` only in README.md?

That being said, if you think `required_env` is the way to go as an user's perspective, it will address 1. above - I like it!

In that case, regarding the 2. above, how about and would you mind:

- Renaming `required_env` to `requiredEnv` according to sprig's function naming of lowerCamelCase
- Replacing every occurrence of `env` to `requiredEnv` you introduce, in README.md

Then, it makes sense from every aspect I can think of.

bastianeicher:
Being able to make some env vars optional and specify default values for them while making others as required seems very useful to me.

Keeping sprig's `env`, combined with `default` and adding a new `requiredEnv` would allow this nicely. For example:

```yaml
releases:
- name: myservice
  chart: somewhere/myservice
  set:
  - name: scheme
    value: {{ env "MYSERVICE_SCHEME" | default "https" }}
  - name: host
    value: {{ requiredEnv "MYSERVICE_HOST" }}
```

Replacing all the `env` instances in the readme with `requiredEnv` sounds like a good idea, but why not also include a sample for the `env` + `default` combo?

sstarcher:
@bastianeicher I'm onboard with your proposal for requiredEnv, env and a env+default combo example.

mumoshu:
Thanks everyone for all your support and the awesome work! Everything LGTM now 👍

Issue #119, opened 2018-04-26
Feature Request: Add support for `helm --set-string`


osterman:
## what
* add support for `--set-string`

## why
* Helm will cast certain values to integers (e.g. `true`)
* Certain fields need to be strings

## use-case

Certain services like `kube-lego` expect a boolean string.

e.g.
```
ingress.annotations.kubernetes\.io/tls-acme=true
```
When passed with `--set`, this gets cast to an integer and causes the error:
```
failed to create patch: failed to get versionedObject: unable to convert unstructured object to extensions/v1beta1, Kind=Ingress: cannot convert int64 to string
```

Using `--set-string` everything works. 

## references
https://github.com/kubernetes/helm/pull/3599

## what does not work
* `--set ingress.annotations.kubernetes\.io/tls-acme="true"`
* `--set ingress.annotations.kubernetes\.io/tls-acme=\"true\"`
* `--set ingress.annotations.kubernetes\.io/tls-acme="!!bool true"`

osterman:
Here are a few possible interfaces:

How it works today:
```
set:
  name: "foo"
  value: "true"
```

## option 1
Introduce a top-level `set_string` arg
```
set_string:
  name: "foo"
  value: "true"
```

## option 2
Introduce a `string` parameter.
```
set_string:
  name: "foo"
  string: "true"
```

## option 3 (preferred)
Introduce a `type` parameter.
```
set_string:
  name: "foo"
  value: "true"
  type: "string"
```

mumoshu:
@osterman Sounds like a good idea 👍 

Nit but you want `set_string` to be just `set` in option 2 and 3, right?

mumoshu:
I "slightly" prefer option 1 over 3 just because it's easier to add validation for unexpected keys at go-yaml-level vs adding our own validation for `type`.

From another point of view, I slightly prefer option 1 over 2 and 3 because (1) it leaves possibility to add any value source at helmfile-level regardless of `--set` `--set-string` and [`--set-file`](https://github.com/kubernetes/helm/pull/3758/files) and (2) it translates directly to respective helm flags.

So, probably I'd expect this in the future:

```yaml
releases:
  - name: myapp
    set:
    - name: val1
      name: foo
    setString:
    - name: val2
      value: bar
    setFile:
    - name: val3
      value: file1
```

which is translated to:

```console
helm upgrade --install --set val1=foo --set-string val2=bar --set-file val3=file1
```

WDYT?

[...]

PR #136, opened 2018-05-07
Escape names and values before passing to Helm

bastianeicher:
Calling `helmfile charts` with a `helmfile.yaml` like this:

```yaml
releases:
  - name: myrelease
    chart: somechart
    set:
      - name: someList
        value: 'a,b,c'
      - name: otherList
        value: '1,2,3'
```

leads to a `helm` invocation like this:

    helm upgrade --install myrelease somechart --set someList=a,b,c,otherList=1,2,3

The commas in the values break the comma separation of `--set` values.

This Pull Request fixes the issue by performing escaping:

    helm upgrade --install myrelease somechart --set someList=a\,b\,c,otherList=1\,2\,3

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -503,7 +503,7 @@ func flagsForRelease(helm helmexec.Interface, basePath string, release *ReleaseS
 	if len(release.SetValues) > 0 {
 		val := []string{}
 		for _, set := range release.SetValues {
-			val = append(val, fmt.Sprintf("%s=%s", set.Name, set.Value))
+			val = append(val, fmt.Sprintf("%s=%s", escape(set.Name), escape(set.Value)))
 		}
 		flags = append(flags, "--set", strings.Join(val, ","))
 	}
@@ -518,7 +518,7 @@ func flagsForRelease(helm helmexec.Interface, basePath string, release *ReleaseS
 		for _, set := range release.EnvValues {
 			value, isSet := os.LookupEnv(set.Value)
 			if isSet {
-				val = append(val, fmt.Sprintf("%s=%s", set.Name, value))
+				val = append(val, fmt.Sprintf("%s=%s", escape(set.Name), escape(value)))
 			} else {
 				errMsg := fmt.Sprintf("\t%s", set.Value)
 				envValErrs = append(envValErrs, errMsg)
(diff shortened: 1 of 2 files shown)


mumoshu:
@bastianeicher  LGTM. Thank you so much for the contribution!

Just waned to note - we will probably avoid relying on the `--set` family in the future. Instead, we will use a temporary generated values.yaml and `-f`. It will solve this kind of problems altogether. See #119 for more context.

So, stay tuned for the future change! Also, please feel free to leave any comments regarding that.

Thanks

mumoshu:
I'll cut a new release asap for your convenience.

teejaded:
This breaks the `array={item1,item2}` syntax.  Also, as far as I can tell it isn't necessary because you can just add "\\" in front of special characters.

https://github.com/kubernetes/helm/blob/master/docs/using_helm.md#the-format-and-limitations-of---set

```
--- #chart template
apiVersion: v1
kind: ConfigMap
metadata:
  name: arrtest
data:
  arrtest: |
    {{- range .Values.array }}
    {{ . }}
    {{- end}}
--- #helmfile
releases:
  - name: test
    chart: ./arrtest
    set:
    - name: array
      value: "{item1,item2}"
```

output:

```
/usr/local/bin/helmfile_darwin_amd64_0_15_0 sync --args "--dry-run --debug"
exec: helm repo update
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "stable" chart repository
Update Complete. ⎈ Happy Helming!⎈
exec: helm dependency update /tmp/test_chart/arrtest
No requirements found in /tmp/test_chart/arrtest/charts.
exec: helm upgrade --install --reset-values test /tmp/test_chart/arrtest --set array=\{item1\,item2\} --dry-run --debug
[debug] Created tunnel using local port: '53227'

[debug] SERVER: "127.0.0.1:53227"

Error: UPGRADE FAILED: render error in "arrtest/templates/arrtest.yaml": template: arrtest/templates/arrtest.yaml:8:21: executing "arrtest/templates/arrtest.yaml" at <.Values.array>: range can't iterate over {item1,item2}
err: exit status 1
```

mumoshu:
Hi! Yes, you are correct. Anyway, thanks to everyone's feedback including
this PR, my understanding is now that we should avoid relying on `--set` at
all.

Helmfile recently introduced inline values in `values:` key. The usage is
basically same and it doesn't suffer from this escaping issue. Would you
mind trying it? #139

Also, probably we should deprecate the `set:` key asap not to confuse
people?

2018年5月17日(木) 8:34 TJ Miller <notifications@github.com>:

Issue #40, opened 2018-02-28
Feature request: Inlined values


hlascelles:
Would there be any appetite to allow inlining values in their natural nested form (instead of name/value pair). I imagine it would require temporary file creation during execution, unless there are libs to convert nested yaml to json dot notation. Not sure if that is beyond helmfile's remit.

eg:

```
charts:
  - name: vault                      
    namespace: vault             
    chart: roboll/vault-secret-manager
    inlined_values: 
      some:
        quite:
        - name: deep
          value: 
            variable: {{ env "VAL" }}
```

mumoshu:
@hlascelles Hi! Thanks for the request.

> I imagine it would require temporary file creation during execution, unless there are libs to convert nested yaml to json dot notation

Would it work for you if helmfile translated your example to:

```yaml
set:
- name: some.quite.deep
   value: "{{ env \"VAL\" }}"
```

Is this what you're looking for?

And also, I'm still unable see a concrete benefit from achieving what you suggest 🤔 

Just curious, but what's your goal of having the another way to specify deeply nested values?

hlascelles:
You're right, there's nothing we can't do with setting name value pairs. I just realised that we had a large number of values files, some complex, some not.

Some were essentially one liners, eg heapster was just:
```
rbac:
  create: true
```

... whereas the complex ones also have ENV values that we need to pass in (so we're currently templating those in temporary files).

The heapster example could easily go into a name/value pair, but then it would be different to the complex charts which need to remain in nested form if we are to understand them properly. Or we split the complex values files to have name/value pairs for ENV *and* a complex values file.

I imagined being able to glance at one (large :smiley:) file, which would show all the values in their nested form, next to the chart that is consuming them, with ENV templating.

This is a "nice to have" (for us) feature request - we're having good success as it is.  Yes, if there is a "yaml to json dot notation" converter helmfile could use internally (post ENV templating), that would do the trick nicely.

maver1ck:
I think the goal is to specify values in helmfile the same way we're specifying them in charts.

mumoshu:
Hey @hlascelles @maver1ck, we have an another thread which is hot about (1) templating entire `values.yaml` before passing it to helm for processing and (2) implement/encourage inlining values into `values:` and prefer doing so over `set:` in `helmfile.yaml`

https://github.com/roboll/helmfile/issues/119#issuecomment-384830608

Would the discussed feature address your use-cases?

mumoshu:
[...]

PR #139, opened 2018-05-10
Inline values support

int128:
This pull request allows the feature #40, that is, `values` accepts the following types:
- A string for values file path.
- A `key: value` map for inline values.

I have tested this feature with the following YAML:

```yaml
releases:
  - name: nginx-ingress
    namespace: kube-system
    chart: stable/nginx-ingress
    values:
      - another.yaml
      - foo:
          bar: {{ env "FOO_BAR" }}
```

and result is:

```
% FOO_BAR=1 ./helmfile -f fixture/helmfile.yaml diff
exec: helm diff nginx-ingress stable/nginx-ingress --values /...snip.../another.yaml --values /var/folders/26/mzllyxzd199ggjmq0lvp0bkr0000gn/T/values395419091
...

% cat /var/folders/26/mzllyxzd199ggjmq0lvp0bkr0000gn/T/values161641152
foo:
  bar: 1
```

Please let me know if you have any suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,11 @@ releases:
       foo: bar
     chart: roboll/vault-secret-manager     # the chart being installed to create this release, referenced by `repository/chart` syntax
     version: ~1.24.1                       # the semver of the chart. range constraint is supported
-    values: [ vault.yaml ]                 # value files (--values)
+    values:
+      - vault.yaml                         # value files (--values)
+      - db:                                # inline values. Passed via a temporary values file (--values)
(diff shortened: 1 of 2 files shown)


sstarcher:
Currently it looks like the temporary file is not cleaned up.  If you add it to the list of `generatedValues` it should get deleted on exit.

int128:
Thank you for your suggestion!
I added cleaning up the temporary file. Also added an example in README.md.

PR #142, opened 2018-05-11
Make purge optional when running "helmfile delete"

### LINKED ISSUE #71: Make purge optional when running "helmfile delete"
Currently, `helmfile delete` runs `helm delete --purge`. This might be unexpected for helm users. Shouldn't purge be an optional flag?

Alternatively, we could have two actual commands for convenience and to make it crystal clear:

    helmfile delete
    helmfile purge

mumoshu:
`helmfile delete` has been implying `--purge` but it is not the case since this change.

The new behavior is:
- `helmfile delete` to delete releases, but without purging them
- `helmfile delete --purge` to actually purge releases.

Resolves #71


mumoshu:
Let's see how this works in the wild

PR #143, opened 2018-05-11
Fix the helm-diff deprecation warning

### LINKED ISSUE #130: Catch up the recent change in helm-diff
We're seeing a confusing deprecation warning message `Command "helm diff" is deprecated, use "helm diff upgrade" instead` from helm-diff, which has troubled users like @rendhalver in e.g. #127.

As the "deprecation" implies, this is caused by the fact that helmfile is calling helm-diff by running `helm diff` but the command is deprecated in favor of `helm diff upgrade`.
We should be able to suppress that by just appending `upgrade` when calling helm-diff.

mumoshu:
Resolves #130

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ dependencies of any referenced local charts.
 The `helmfile diff` sub-command executes the [helm-diff](https://github.com/databus23/helm-diff) plugin across all of
 the charts/releases defined in the manifest.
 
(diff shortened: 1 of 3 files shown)


sstarcher:
Might want to update the docs and specify that a version greater than X is required for helm diff to work

mumoshu:
@sstarcher Thanks for the comment! Added it.

mumoshu:
I appreciate your review ☺️

Issue #144, opened 2018-05-14
Feature Request: helmfile test option


Evesy:
**Feature**
An additional directive to the helmfile binary; `helmfile test`

**Why**
Helm provides `helm test` (https://github.com/kubernetes/helm/blob/master/docs/chart_tests.md) as a method to run automated tests post chart install to ensure things are working as they should be.

It would be nice to be able to run something like `helmfile test` against a particular helmfile in order to run helm tests against all charts/releases defined in the file. Either as part of the sync (i.e. `helmfile sync --test`) to be ran immediately after the corresponding chart is installed, or as a separate command ran after a sync (i.e. `helmfile test`).

A chart without tests will exit with a 0 status, so it can be safely ran against any charts.

mumoshu:
@Evesy Thanks for the awesome feature request. I have never noticed about that helm feature. Interesting.

Can we just start by adding `helmfile test [--selector k1=v1,k2=v2,...] [--cleanup] [--timeout int] [--args ...]`?

Reasons:

For dev use-case, I prefer helmfile to do just one thing very well. In this specific case, I'd expect helmfile to just fan-out to helm, rather than doing workflow thing for me. So I'd write `helmfile sync && helmfile test` or `SELECTOR='tier=backend' bash -c 'helmfile sync -l $SELECTOR && helmfile test -l $SELECTOR` instead of introducing `helmfile sync --test`.

For production use-case, I prefer delegating anything other than just fan-out of helm commands to external workflow system like argo and brigade.
More concretely, I'd run `helmfile sync` and then `helmfile test` in separate but dependent jobs in the workflow system, so that I can benefit from reliability and durability from the system.

Evesy:
@mumoshu `helmfile test` would probably be my preference over bringing it into the scope of `helmfile sync` for the reasons you listed above. It also gives the opportunity to run any tests on a scheduled basis without the need to kick off a helm upgrade

Evesy:
FYI I've made a start at https://github.com/Evesy/helmfile however I've never touched any Golang before so it'll be slow progress (and a good chance I get stuck at some point!)

PR #150, opened 2018-05-17
Add helmfile test sub-command

Evesy:
Ref: https://github.com/roboll/helmfile/issues/144

**Feature**
An additional sub-command to the helmfile binary; helmfile test

**Why**
Helm provides helm test (https://github.com/kubernetes/helm/blob/master/docs/chart_tests.md) as a method to run automated tests post chart install to ensure things are working as they should be.

It would be nice to be able to run something like helmfile test against a particular helmfile in order to run helm tests against all charts/releases defined in the file. Either as part of the sync (i.e. helmfile sync --test) to be ran immediately after the corresponding chart is installed, or as a separate command ran after a sync (i.e. helmfile test).

A chart without tests will exit with a 0 status, so it can be safely ran against any charts.

**Notes**

`--cleanup` (bool) & `--timeout` (default: 300) are available as first class arguments. Additional arguments can be passed to the helm binary as with other sub commands using `--args=`

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"text/template"
@@ -382,7 +383,7 @@ func (state *HelmState) DeleteReleases(helm helmexec.Interface, purge bool) []er
 	return nil
 }
 
-func (state *HelmState) TestReleases(helm helmexec.Interface, cleanup bool) []error {
+func (state *HelmState) TestReleases(helm helmexec.Interface, cleanup bool, timeout int) []error {
 	var wg sync.WaitGroup
 	errs := []error{}
 
@@ -393,6 +394,7 @@ func (state *HelmState) TestReleases(helm helmexec.Interface, cleanup bool) []er
 			if cleanup {
 				flags = append(flags, "--cleanup")
 			}
+			flags = append(flags, "--timeout", strconv.Itoa(timeout))
 			if err := helm.TestRelease(release.Name, flags...); err != nil {
 				errs = append(errs, err)
 			}
--- a/state/state.go
+++ b/state/state.go
@@ -382,7 +382,7 @@ func (state *HelmState) DeleteReleases(helm helmexec.Interface, purge bool) []er
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


state/state.go
```diff
@@ -382,6 +383,34 @@ func (state *HelmState) DeleteReleases(helm helmexec.Interface, purge bool) []er
 	return nil
 }
 
+// TestReleases wrapper for executing helm test on the releases
+func (state *HelmState) TestReleases(helm helmexec.Interface, cleanup bool, timeout int) []error {
+	var wg sync.WaitGroup
+	errs := []error{}
+
+	for _, release := range state.Releases {
```
mumoshu: Probably we want concurrency control in near future. We do that in e.g. [`ReleaseStatuses`](https://github.com/roboll/helmfile/pull/150/files#diff-f6be34387a141f652301c16cf7a5e36cR316).

But that's an another story and I'm ok with it isn't covered in this PR.
Just wanted to make it clear.

state/state_test.go
```diff
@@ -552,6 +552,9 @@ func (helm *mockHelmExec) DeleteRelease(name string, flags ...string) error {
 func (helm *mockHelmExec) DecryptSecret(name string) (string, error) {
 	return "", nil
 }
+func (helm *mockHelmExec) TestRelease(name string, flags ...string) error {
+	return nil
```
mumoshu: If I could ask for more - Would you mind adding a test for `TestReleases`?

Probably you'd make this func look like the sub [ReleaseStatus](https://github.com/roboll/helmfile/pull/150/files#diff-a1153b96e079da8da43231ad48de57a0R542) func:

```golang
func (helm *mockHelmExec) TestRelease(name string, flags ...string) error {
 	if strings.Contains(name, "error") {
 		return errors.New("error")
 	}
 	helm.releases = append(helm.releases, mockRelease{name: name, flags: flags})
 	return nil
 }
```

And write a test involves the stub like https://github.com/roboll/helmfile/pull/150/files#diff-a1153b96e079da8da43231ad48de57a0R700

Evesy:
@mumoshu If you want to run it home with some tests that'd be great. I've had a look but I'd need to take some time to figure it out, and I'm also eager to get this feature out :)

Cheers

mumoshu:
Here it is 😄#152

mumoshu:
@Evesy Thank you so much for your contribution 👍

Issue #93, opened 2018-04-06
More control over log verbosity


manics:
Follow-up from https://github.com/roboll/helmfile/issues/79#issuecomment-378544353
The use case in that comment could be solved by defining and hiding secret variables, but @mumoshu suggested more control over logging such as `trace`, `debug`, `info`, `warn`, `error` would be useful.

For example, when creating or troubleshooting a helmfile having more verbose output is useful. In contrast if you're deploying a large production system with many charts you might prefer reduced output to summarise the deployment status of each chart.

At present there are two logging levels, normal which prints the helm command and all output, and `quiet` which hides everything including the output of helm.

osterman:
Yes, this is an issue for me too. I need to pass `--quiet`, otherwise, all secrets are output in the logs due to the output of `exec: ....` with `--set` parameters.  But what I really want is to see the output from `helm` without the `exec`.  Thus, I support some sort of `--log-level` argument of which the `exec: ...` should be `debug`.

manics:
I've looked into possible libraries for this. [glog](https://github.com/golang/glog) seems overly focussed on low-level logging, for example it defaults to logging to a file and there's no way to send logs to stdout, only stderr.

[go.uber.org/zap](https://github.com/uber-go/zap) looked more promising, you can configure stdout or stderr logs, and it provides an optional global logger `zap.S().Info(...)`. See https://github.com/manics/helmfile/commit/a515de458171757d866bdc7a3503664f8ddd2847 for my an initial example.

manics:
@mumoshu in principle does using go.uber.org/zap sound OK to you? If so I'll do more investigation/coding.

mumoshu:
@manics Sorry for the delayed response! Yes, I'm fine with any logger as long as it is "leveled", and I have no desire to introduce loggers with rich sets of outputs(like logrus). That being said, I'm fine with zap.

PR #154, opened 2018-05-29
Add basic repository authentication

InTheCloudDan:
We have a remote repository that requires username and password to authenticate to. This PR enables that basic functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,8 @@ repositories:
     url: http://roboll.io/charts
     certFile: optional_client_cert
     keyFile: optional_client_key
+    username: optional_username
+    password: optional_password
 
 context: kube-context					 # kube-context (--kube-context)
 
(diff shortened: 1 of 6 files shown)


InTheCloudDan:
One thing I did notice is that with the way I handled it, even if it's an environment variable it prints out the value to screen as part of the helm exec. I'm not sure if it'd be worth hiding the password somehow.

mumoshu:
Hi, thank you so much for this!

The proposed feature and code LGTM, but yes, we should passwords somehow.
Otherwise one needs to hide it by piping a helmfile run to `sed` to mask passwords? 😃 

You can browse #93 and 79 for more context. I'm considering to introduce log levels. How about `helm repo add` is logged only when the log level is greater than or equal to `DEBUG`?

InTheCloudDan:
That would work for me, possibly log "Adding Repositories" as a generic message to know what stage the exec is at.

If that's going to be handled outside of this PR, would you like me to update the README with a temporary warning so this can be merged in, then the warning could be removed when the final decision is made/implemented?

mumoshu:
@InTheCloudDan Thanks again! The generic message would be a great idea.

A warning in README would also be nice. I'll merge this asap after you added an additional commit for that 👍

InTheCloudDan:
Added a basic warning. I didn't want to go to overboard with it and figured makes sense next to info about the helm version. If there's any changes you want please let me know.

Issue #137, opened 2018-05-10
Ordering


Stono:
Firstly, 
Thanks for this project, it's awesome - you've saved me a big headache.

However, quick Q.   I need to control the order in which the charts are deployed.

Is there any way (hacky or not) that I can do that, without using labels and having two helmfile runs?

sstarcher:
not currently

mumoshu:
@Stono Hi! Just curious but why do you want the ordering?

Stono:
Hello @mumoshu!
Great question, let me try and explain...

We have a helmfile which can take an empty kubernetes cluster and turn it into what we call the "delivery platform", however the components need to be deployed in a certain order, for a variety of reasons:

  - We need to deploy a chart which contains CustomResourceDefinitions **before** we deploy a chart which creates instances of those CustomResources
  - We use a service mesh, which uses a MutatingWebhookHandler in order to inject the sidecar into Pods as they're deployed.  This needs to be configured before any Deployments, ReplicaSets or Pods are deployed.

As a result, we've ended up with a bit of a hacky script that looks like this:

```
echo "Creating MutatingWebhookHandler certificates..."
cd manual
./install-sidecar-injection.sh
cd ..

echo "Deploying istio (first phase)..."
helmfile --selector component=istio,order=1 sync --args="--wait --timeout=300"

echo "Adding the MutatingWebhookHandler..."
kubectl apply -f manual/istio-sidecar-injector-with-ca-bundle.yaml

echo "Deploying istio (second phase)..."
helmfile --selector component=istio,order!=1 sync --args="--wait --timeout=300"

echo "Deploying everything else..."
helmfile --selector component!=istio sync --args="--wait --timeout=300"
```

<img width="969" alt="screen shot 2018-05-11 at 09 22 26" src="https://user-images.githubusercontent.com/1486729/39914585-5c385656-54fd-11e8-83b6-7c9a776a11d3.png">

mumoshu:
@Stono Hi. Thank you so much for the detailed explanation!
Makes sense.

Btw, would this suggested feature https://github.com/roboll/helmfile/issues/151#issuecomment-391165364 help your use-case, too?

Stono:
Yes it would so long as the files were processed in order 👌

gtaylor:
To throw in another usage case, we have Helm charts for Roles and RoleBindings. IIRC, the latter will fail if you try to bind a Role that doesn't exist yet.

[...]

PR #160, opened 2018-06-06
Ordered helmfile(s) from a directory

### LINKED ISSUE #151: Feature Request: Load helmfiles from a directory
## what
* Add support for a `--dir` option which will process each file as a *standalone* `helmfile.yaml`
* Maybe add a `--recursive` option to scan directories recursively 
e.g. 

```
helmfile --dir conf.d --selector default=true sync
```

 
## why

* It should be much simpler to iterate over files in a directory than figuring out how to merge multiple templatized YAML files
* Compatibility with [`CODEOWNERS`](https://help.github.com/articles/about-codeowners/)

## use-case

Our [helmfile](https://github.com/cloudposse/geodesic/blob/master/rootfs/conf/kops/helmfile.yaml) is too big to manage (1330 lines). 

## additional context

All we're asking is for the equivalent functionality of this:
```
for conf in conf.d/*; do
  helmfile --file=$conf $@
done
```

## related
* https://github.com/roboll/helmfile/issues/96
* https://github.com/roboll/helmfile/issues/96#issuecomment-379483155

mumoshu:
Resolves #151 #137

### DIFF
--- a/main.go
+++ b/main.go
@@ -302,27 +302,41 @@ func eachDesiredStateDo(c *cli.Context, converge func(*state.HelmState, helmexec
 	return nil
 }
 
-func findDesiredStateFiles(fileOrDirPath string) ([]string, error) {
-	var dir string
-
-	if fileOrDirPath != "" {
-		if !fileExists(fileOrDirPath) {
-			return []string{}, fmt.Errorf("state file named %s is not found", fileOrDirPath)
+func findDesiredStateFiles(specifiedPath string) ([]string, error) {
+	var existingDir string
+	if specifiedPath != "" {
+		if !pathExists(specifiedPath) {
+			return []string{}, fmt.Errorf("specified state file %s is not found", specifiedPath)
 		}
 
-		if !directoryExists(fileOrDirPath) {
-			return []string{fileOrDirPath}, nil
+		if !directoryExists(specifiedPath) {
+			return []string{specifiedPath}, nil
 		}
 
-		dir = fileOrDirPath
+		existingDir = specifiedPath
+	} else if pathExists(DefaultHelmfileDirectory) && directoryExists(DefaultHelmfileDirectory) {
+		existingDir = DefaultHelmfileDirectory
 	}
 
-	if dir == "" && fileExists(DefaultHelmfileDirectory) && directoryExists(DefaultHelmfileDirectory) {
-		dir = DefaultHelmfileDirectory
+	var existingFile string
+	if pathExists(DefaultHelmfile) {
+		existingFile = DefaultHelmfile
+	} else if pathExists(DeprecatedHelmfile) {
+		log.Printf(
+			"warn: %s is being loaded: %s is deprecated in favor of %s. See https://github.com/roboll/helmfile/issues/25 for more information",
+			DeprecatedHelmfile,
+			DeprecatedHelmfile,
+			DefaultHelmfile,
+		)
+		existingFile = DeprecatedHelmfile
 	}
 
-	if dir != "" {
-		files, err := filepath.Glob(filepath.Join(dir, "*.yaml"))
+	if specifiedPath == "" && existingFile != "" && existingDir != "" {
+		return []string{}, fmt.Errorf("configuration conlict error: you can have either %s or %s, but not both", existingFile, existingDir)
+	}
+
+	if existingDir != "" {
+		files, err := filepath.Glob(filepath.Join(existingDir, "*.yaml"))
 		if err != nil {
 			return []string{}, err
 		}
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -317,8 +283,77 @@ func main() {
...
+		}
+
+		dir = fileOrDirPath
+	}
+
+	if dir == "" && fileExists(DefaultHelmfileDirectory) && directoryExists(DefaultHelmfileDirectory) {
+		dir = DefaultHelmfileDirectory
+	}
+
+	if dir != "" {
```
sstarcher: I think this is clearer
```
if dir != "" { 
...
else if  fileExists(DefaultHelmfileDirectory) && directoryExists(DefaultHelmfileDirectory) {
...
}
```

main.go
```diff
@@ -317,8 +283,77 @@ func main() {
...
+		if err != nil {
+			return []string{}, err
+		}
+		sort.Slice(files, func(i, j int) bool {
+			return files[i] < files[j]
+		})
+		return files, nil
+	}
+
+	if fileExists(DefaultHelmfile) {
```
sstarcher: Should we complain if we have both a helmfile dir and a helmfile?  Or should we load both?  Currently it would ignore the helmfile and use the helmdir.

gtaylor: That's a sneaky kind of behavior, so whatever we do, I hope we make it explicit and/or noisy. Maybe a warning + using the helmdir?

osterman: Typically with apps that have a `conf.d` structure, they also support a main config file. That main config file then will then have an option to point to a `conf.d` folder. For example `systemd`. Now, I'm *not* advocating adding this to this PR. As far as I'm concerned, executing one or the other (or even both) is fine by me.

mumoshu: Thanks for the feedbacks! Let me go forward with the immediate exit for now. I want to make it very explicit.

main.go
mumoshu: What would be included in `if` and `else if` bodies respectively? We need `dir` to be set for use in https://github.com/roboll/helmfile/pull/160/files#diff-7ddfb3e035b42cd70649cc33393fe32cR325.

mumoshu: Would it look clearer now?

main.go
mumoshu: Addressed in the last commit

main.go
```diff
@@ -317,25 +283,93 @@ func main() {
...
+		log.Printf(
+			"warn: %s is being loaded: %s is deprecated in favor of %s. See https://github.com/roboll/helmfile/issues/25 for more information",
+			DeprecatedHelmfile,
+			DeprecatedHelmfile,
+			DefaultHelmfile,
+		)
+		existingFile = DeprecatedHelmfile
+	}
+
+	if specifiedPath == "" && existingFile != "" && existingDir != "" {
```
mumoshu: I believe that `specifiedPath == ""` is needed in order to allow the user to selectively run `helmfile sync` for the default helmfile.d or helmfile.yaml, and `helmfile sync -f $path` for a specific helmfile

sstarcher: The logic looks correct after stepping through a bunch of options, but it's very complicated.  I think I would move the DefaultHelmfile/DeprecatedHelmfile check as part of the first if else.  I also think you can drop the `specifiedPath==""` from this if statement as the first function should always set existingDir if specifiedPath is set or return.

And of course I could be missing something on my recommendations.

mumoshu: Ah, so this block should be nested inside the `} else if pathExists(DefaultHelmfileDirectory) && directoryExists(DefaultHelmfileDirectory) {` block in the first if-else?
Sounds good! I'll address it asap. Thanks 🙏

mumoshu:
@sstarcher Thanks for the review. I wonder if it's a bit more readable now :)

osterman:
Itching to use this!!

osterman:
Can we get a new release of `helmfile` so we can download/install the binary in our scripts?

alebabai:
@mumoshu could you explain a bit how to use `helmfile.d`?
Because it's seems it doesn't work for me, i'm using `Helmfile v0.20.0`

So i have only `helmfile.d` dir in current directory.
I executed `$ helmfile sync` and get 
> 2018/07/04 14:29:52 err: specified state file helmfile.yaml is not found.

Another question is about path to the values-files if `helmfile.d` is used.
Will path to the values-files be resolved relatively to helmfile.d folder or execution folder?

alebabai:
It works for `helmfile -f helmfile.d/`.
By the way in this case path to the values-files be resolved relatively to helmfile.d folder and to the current execution folder.

PR #162, opened 2018-06-08
Add helmfile lint support

jlyheden:
Our use case is to have a list of helmfile releases version controlled together with all settings and have a CI pipeline that will lint all releases with settings before running sync. The new functionality was mostly copy pasted from the Diff implementation with some extra handling for fetching remote charts so open for any feedback.

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -37,6 +37,8 @@ type RepositorySpec struct {
 	URL      string `yaml:"url"`
 	CertFile string `yaml:"certFile"`
 	KeyFile  string `yaml:"keyFile"`
+	Username string `yaml:"username"`
+	Password string `yaml:"password"`
 }
 
 // ReleaseSpec defines the structure of a helm release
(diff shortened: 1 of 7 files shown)


state/state.go
```diff
@@ -315,6 +316,122 @@ func (state *HelmState) DiffReleases(helm helmexec.Interface, additionalValues [
...
+		return errs
+	}
+	defer os.RemoveAll(dir)
+
+	for w := 1; w <= workerLimit; w++ {
+		go func() {
+			for release := range jobQueue {
+				errs := []error{}
+				// Plugin command doesn't support explicit namespace
+				release.Namespace = ""
```
mumoshu: I guess `helm list` supports namespacing, right? 🤔 Could you remove this line?

jlyheden: You are absolutely right, removed it!

jlyheden:
I'll look into rebasing the branch. I also noticed that there was a potential race condition when running lint with concurrency > 1 as multiple go funcs could fetch to the same temp location in parallell, which will be avoided by separating each of the releases fetched chart by release name.

PR #166, opened 2018-06-12
display hint to user about sync before diff

cmeury:
Example output:

```
$ helmfile  diff
exec: helm diff ...
Error: This command needs 2 arguments: release name, chart path
Usage:
...
Error: plugin "diff" exited with error
err: exit status 1
err: Please make sure that you have run 'helmfile charts' at least once before running 'helmfile diff'.
```

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -281,6 +281,9 @@ func (state *HelmState) DiffReleases(helm helmexec.Interface, additionalValues [
 				if len(errs) == 0 {
 					if err := helm.DiffRelease(release.Name, normalizeChart(state.BaseChartPath, release.Chart), flags...); err != nil {
 						errs = append(errs, err)
+						// display hint to user until https://github.com/databus23/helm-diff/pull/15 is merged
+						hint := "Please make sure that you have run 'helmfile charts' at least once before running 'helmfile diff'."
+						errs = append(errs, errors.New(hint))
 					}
 				}
 				for _, err := range errs {


state/state.go
```diff
@@ -281,6 +281,9 @@ func (state *HelmState) DiffReleases(helm helmexec.Interface, additionalValues [
 				if len(errs) == 0 {
 					if err := helm.DiffRelease(release.Name, normalizeChart(state.BaseChartPath, release.Chart), flags...); err != nil {
 						errs = append(errs, err)
+						// display hint to user until https://github.com/databus23/helm-diff/pull/15 is merged
+						hint := "Please make sure that you have run 'helmfile charts' at least once before running 'helmfile diff'."
```
mumoshu: `helmfile sync` rather than `helmfile charts`?

cmeury: `sync` is a superset of `charts`, isn't `charts` the minimum to run?

mumoshu: It seems like I'm pretty confused with the naming of `helmfile charts` after all. Because it actually deal with releases, rather than charts.

Anyway, I guess `helmfile charts` is a subset of `sync`, because it [sync releases only](https://github.com/roboll/helmfile/blob/df264e273630c7e7af58721d75ba447bc9097964/main.go#L116). If the user ran `helmfile diff` before `helmfile repos`, the subsequent `helmfile charts` run would just fail due to missing chart deps? Then, I think that we'd better to suggest `helmfile sync` rather than `helmfile charts`. Please correct me if I'm missing something!

sstarcher:
Running `charts` would cause an update.  Why would we want to recommend they run that before diff?  If anything `repos` needs to be ran first.

cmeury:
The thing is, you cannot diff without an already deployed release. So the command only works if at minimum we did `charts` (= helm upgrade) or `sync` (= repo update & helm upgrade). It is not necessary to update the repositories if they were already updated previously. But the release to be diff'd needs to be around. Hence, we say you need to have at least once deployed the release for diff to work.

sstarcher:
I'm on board with discussing it until helm-diff gets updated ;)

cmeury:
Let's close this in favour of #172 :)

Issue #114, opened 2018-04-23
Add ability to specify tiller namespace


rmartinez3:
I want to be able to use helmfile to talk to different instances of tillers in different namespace.  I would like to add a global flag  --tiller-namespace , to specify the namespace where tiller is located. Also if possible to have a tillerNamespace section in helmfile. That way can specify either via flag or helmfile what tiller to talk .

mumoshu:
@rmartinez3 Hi! Thanks for the feature request 👍 

Probably, adding `--tiller-namespace` to `sync`, `diff`, `delete` makes sense to me.
But FYI, I guess you can do that today with e.g. `helmfile sync --args "--tiller-namespace=yourns"`.

I'd rather want to avoid enchancing helmfile.yml here.
The context is that, I basically like to include only declarative parts of helm releases in helmfile.yml.
`--tiller-namespace` seems like a part of operational knowledge - on developer's minikube you'd just want to omit `tiller-namespace` but in staging and prod clusters you'd want to specify the tiller-ns.

WDYT?

rmartinez3:
Hello thanks for  responding and thanks for the workaround. 

Yes it  makes sense to have it added to sync , diff and delete. 


As for tiller-namespace being set via helmfile. The reason I would like to have it set. Its to minimize mistakes , currently we have our own maintained helmfile repo  and will like to move back to main stream helmfile. We set the kube-context and tiller-namespace in helmfile.yaml, so that a install or upgrade via helmfile will be just referencing helmfile and values.  That way the person using helmfile wouldn't have to worry about setting tiller-namespace flag. We currently have tiller instances deployed for different teams in each cluster and helmfile helps us distinguish what to deploy to. That with rbac, having teams with their own namespaces. Setting tiller-namespace via helmfile helps out.

I did see now that labels have been added via helmfile. So something i would look into as well.

mumoshu:
@rmartinez3 Hi. Thank you so much for the detailed explanation!
I have not yet encountered advanced use-cases like yours myself, but I believe your points are valid and very important.

Looking back, we already have an operational knowledge of "which kube-context to be used by helm` in our `helmfile.yaml`. It is named `context:`, which you can see from the README of this project.

That being said, how about enhancing helmfile.yaml to add more operational context to helmfile?

Currently we have:

```yaml
context: kubectl-context-name
```

How about enhancing it so that we can pack all the operational knowledge into it:

```yaml
context:
  # What exists today
  kubeContext: default
[...]

PR #171, opened 2018-06-13
adding args to be added in helmfile along with context to be set in helmspec

rmartinez3:
This is regarding issue #114. This PR allows for args to be set in helmfile. Added section helm which 
 args can be set ,as well as context will be set and other contributed key values later on when contributed.

### DIFF
--- a/main.go
+++ b/main.go
@@ -71,9 +71,10 @@ func main() {
 					return err
 				}
 
-				args := c.String("args")
+				args := getArgs(c, state)
+
 				if len(args) > 0 {
-					helm.SetExtraArgs(strings.Split(args, " ")...)
+					helm.SetExtraArgs(args...)
 				}
 
 				errs := state.SyncRepos(helm)
@@ -105,9 +106,10 @@ func main() {
 					return err
 				}
 
-				args := c.String("args")
+				args := getArgs(c, state)
+
 				if len(args) > 0 {
-					helm.SetExtraArgs(strings.Split(args, " ")...)
+					helm.SetExtraArgs(args...)
 				}
 
 				values := c.StringSlice("values")
@@ -146,9 +148,10 @@ func main() {
 					return err
 				}
 
-				args := c.String("args")
+				args := getArgs(c, state)
+
 				if len(args) > 0 {
-					helm.SetExtraArgs(strings.Split(args, " ")...)
+					helm.SetExtraArgs(args...)
 				}
 
 				if c.Bool("sync-repos") {
@@ -206,9 +209,10 @@ func main() {
 					os.Exit(1)
 				}
 
-				args := c.String("args")
+				args := getArgs(c, state)
+
 				if len(args) > 0 {
-					helm.SetExtraArgs(strings.Split(args, " ")...)
+					helm.SetExtraArgs(args...)
 				}
 
 				values := c.StringSlice("values")
@@ -241,9 +245,10 @@ func main() {
 
 				workers := c.Int("concurrency")
 
-				args := c.String("args")
+				args := getArgs(c, state)
+
 				if len(args) > 0 {
-					helm.SetExtraArgs(strings.Split(args, " ")...)
+					helm.SetExtraArgs(args...)
 				}
 
 				errs := state.ReleaseStatuses(helm, workers)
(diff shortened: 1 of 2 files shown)


state/state.go
```diff
@@ -24,13 +24,19 @@ import (
 // HelmState structure for the helmfile
 type HelmState struct {
 	BaseChartPath      string
+	Helm               HelmSpec         `yaml:"helm"`
```
mumoshu: I'm still thinking about what would be the best key name here. Assuming it is possible to override what are specified under the `helm:` setting, how about naming it `helmDefaults:` so that it would be clearer that you can override?

main.go
```diff
@@ -343,7 +349,19 @@ func before(c *cli.Context) (*state.HelmState, helmexec.Interface, error) {
...
+			log.Printf("err: Cannot set atrribute context on both helmspec and context.")
+			os.Exit(1)
+		}
+
 		kubeContext = st.Context
+	} else if st.Helm.Context != "" {
+		if kubeContext != "" {
+			log.Printf("err: Cannot use option --kube-context and set attribute context.")
+			os.Exit(1)
+		}
```
mumoshu: Similarty to https://github.com/roboll/helmfile/pull/171/files#r195421646, I'm fine with this situation, but I prefer the value from the command-line takes precedence

state/state.go
rmartinez3: helmDefaults sounds good to me. Will make changes

main.go
```diff
@@ -394,3 +413,16 @@ func clean(state *state.HelmState, errs []error) error {
...
 	return nil
 }
+
+func getArgs(c *cli.Context, state *state.HelmState) []string {
+	args := c.String("args")
+	if len(args) > 0 {
+		if len(state.Helm.Args) > 0 {
+			log.Printf("err: Cannot use option --args and set attribute args.")
+			os.Exit(1)
+		}
```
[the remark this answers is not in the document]
rmartinez3: sounds good. will have --args take precedence over args in helmfile

rmartinez3:
So just to clarify . Command line args will take precedence over args set on helmfile. Im thinking of adding a map check for arguments. Should checking of --kube-context <context> be checked in the args section as well ? If so, should the  flags take precedence ?

helmfile -f helmfile.yaml --kube-context test   diff --args --kube-context=test1

kube-context test will take precedence

Not sure if I'm overthinking this.
Thanks

schollii:
Can't tell if this is still the case, but looking at this: 
```
helm:
  args:
  - --kube-context=default
  # Your requested feature
  - --tiller-namespace=yourns
```
Having to specify "--" before a helm option in the yaml file is unnecessary, noisy, and unconventional. The reason we use -- on command line is to make it clear that these are optional settings. Here the user is looking at a settings file, and clearly the args section is entirely optional, so why clutter the file with "- --"? the option name "kube-context" should be sufficient. Less change for typos, easier to read.

mumoshu:
@schollii Hi, thanks for the feedback!

Actually I'm fine with including `--` in the array items, because it looks natural to me after `args:` in the k8s pod spec.

Anyway, we(including you and me) should be always open to add dedicated keys later. See my first proposal in https://github.com/roboll/helmfile/issues/114#issuecomment-384851216. Take `helmDefaults.args` as a dirty but a cover-all setting to get your work done today. We can later enhance helmfile to support keys like `tillerNamespace` which translates to `--tiller-namespace=yourns` as I wrote in the linked comment.

WDYT?

mumoshu:
@rmartinez3 Hey, thanks for your effort!

> Command line args will take precedence over args set on helmfile.

Yes. So `--args` overrides `helmDefaults.args`.

> Im thinking of adding a map check for arguments. Should checking of --kube-context be checked in the args section as well ? If so, should the flags take precedence ?
> 
> helmfile -f helmfile.yaml --kube-context test diff --args --kube-context=test1
> 
> kube-context test will take precedence

I'm fine with no taking that into account at all. Just append the `args` to the helm command without validation. `helmDefaults.args` is basically the default value for `--args`. It is an array of cmdline args and merging it in a reliable way doesn't worth the effort imho.

If we want the merging facility, we should implement it along with the dedicated key like `tillerNamespace` as I commented above.

Thoughts?

rmartinez3:
For now i can just leave it as command --args override helmDefaults.args. Then i can afterwards make a merge request to include the arg merging and tillernamespace key. Thanks, will work on this later today.

rmartinez3:
I made changes for command line arg to take precedence over helm default args.

osterman:
Just in case it helps others... the syntax as of release `0.20.0` looks like this:

```
helmDefaults:
  args:
    - "--wait"
    - "--recreate-pods"
    - "--timeout=600"
    - "--force"
    - "--reset-values"
```

PR #181, opened 2018-06-29
Add the option to specify an alternative helm binary

hori-ryota:
Helm may not be version compatible, how about option to overwrite command for switching other version client?

incompatible versions error sample is followings
```
Error: UPGRADE FAILED: incompatible versions client[v2.9.1] server[v2.8.2]
```

### DIFF
--- a/helmexec/exec.go
+++ b/helmexec/exec.go
@@ -11,7 +11,7 @@ const (
 )
 
 type execer struct {
-	command     string
+	helmBinary  string
 	runner      Runner
 	writer      io.Writer
 	kubeContext string
@@ -21,7 +21,7 @@ type execer struct {
 // New for running helm commands
 func New(writer io.Writer, kubeContext string) *execer {
 	return &execer{
-		command:     command,
+		helmBinary:  command,
 		writer:      writer,
 		kubeContext: kubeContext,
 		runner:      &ShellRunner{},
@@ -33,7 +33,7 @@ func (helm *execer) SetExtraArgs(args ...string) {
 }
 
 func (helm *execer) SetHelmBinary(bin string) {
-	helm.command = bin
+	helm.helmBinary = bin
 }
 
 func (helm *execer) AddRepo(name, repository, certfile, keyfile, username, password string) error {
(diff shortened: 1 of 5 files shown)


helmexec/exec.go
```diff
@@ -20,6 +21,7 @@ type execer struct {
 // New for running helm commands
 func New(writer io.Writer, kubeContext string) *execer {
 	return &execer{
+		command:     command,
```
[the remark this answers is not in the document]
mumoshu: You can also name it `binary` but I'm pretty sure we need to add another `binary` for `helm-diff` to call it directly to work-around https://github.com/databus23/helm-diff/issues/54 and https://github.com/kubernetes/helm/issues/4170. In that regard, I'd still prefer `helmBinary` over `binary`.

main.go
```diff
@@ -31,6 +31,10 @@ func main() {
 	app.Usage = ""
 	app.Version = Version
 	app.Flags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "helmcommand, c",
```
mumoshu: I'd prefer `helm-binary` as per other comments!

main.go
```diff
@@ -31,6 +31,10 @@ func main() {
...
+			Usage: "overwrite helm command. Uses 'helm by default",
```
mumoshu: `path to helm binary`, and no need to note about the default value. AFAIK cobra include it in the usage when you set the default value via the `Value` field.

hori-ryota:
replaced 😋

main.go
```diff
@@ -31,6 +31,10 @@ func main() {
 	app.Usage = ""
 	app.Version = Version
 	app.Flags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "helm-binary, b",
+			Usage: "path to helm binary",
```
mumoshu: Oh, I somehow missed about this, but do you have any specific reason not to add `Value: "helm",` instead of putting this in many places?

```
if c.String("helm-binary") != "" {
    helm.SetHelmBinary(c.String("helm-binary"))
}
```

Just asking. I'd address it myself even when it turned out we'd better use the default value instead.

hori-ryota: I don't have a clear reason, but in other words I thought that it would be bad for the default value definition to be scattered in two places.

If we include default values when parsing command arguments, it may be more appropriate to pass them as arguments to execer's constructor.

mumoshu: Yep, made a lot of sense! I'll definitely consider doing so. Thanks for your confirmation.

mumoshu:
Squashed commits while tweaking the commit message a bit

PR #185, opened 2018-07-09
Multi-level logging

manics:
Closes https://github.com/roboll/helmfile/issues/93

Summary of changes:
- Adds the [uber-go-/zap](https://github.com/uber-go/zap) library for levelled logging (69040996e0bcbd2f211d0d8d576147027e928e3b)
- A new CLI arg `--log-level` which can take the values `debug` `info` `warn` `error` `fatal` `panic`. The default is `info`, and the existing `--quiet` flag is treated as `--log-level warn`
- All `helm.exec` calls are preceded by a one-line summary. The current information on the full helm command line arguments `exec: helm exec ... ` is only output at `debug` level. This means sensitive command line arguments which may include passwords and `--set` arguments should be hidden by default.


sstarcher:
This looks like a nice change.  I'll try and get some time to test it out.

PR #186, opened 2018-07-11
[WIP] adding merging for helm defaults args and --args, adding tillernamspace and kubecontext to helmdefaults

rmartinez3:
this pul request adds tillernamspace and kubecontext to helmdefaults also added merging of args capability. 

made it possible if more people want to add key defaults they can add field to helmdefaults and add arg tag so merger can append to args

### DIFF
--- a/main.go
+++ b/main.go
@@ -6,14 +6,15 @@ import (
 	"log"
 	"os"
 	"os/signal"
+	"path/filepath"
+	"reflect"
+	"sort"
 	"strings"
 	"syscall"
 
 	"github.com/roboll/helmfile/helmexec"
 	"github.com/roboll/helmfile/state"
 	"github.com/urfave/cli"
-	"path/filepath"
-	"sort"
 )
 
 const (
@@ -189,9 +190,9 @@ func main() {
 			},
 			Action: func(c *cli.Context) error {
 				return eachDesiredStateDo(c, func(state *state.HelmState, helm helmexec.Interface) []error {
-					args := c.String("args")
+					args := getArgs(c, state)
 					if len(args) > 0 {
-						helm.SetExtraArgs(strings.Split(args, " ")...)
+						helm.SetExtraArgs(args...)
 					}
 					if c.GlobalString("helm-binary") != "" {
 						helm.SetHelmBinary(c.GlobalString("helm-binary"))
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -493,9 +494,69 @@ func clean(state *state.HelmState, errs []error) error {
...
+				if len(argVal) > 1 {
+					argsMap[arg] = argVal[1]
+				} else {
+					argsMap[arg] = ""
+				}
+			}
+		}
+	}
+
+	setDedicatedKeys(state, argsMap)
```
mumoshu: Shouldn't we move this to line 498? Otherwise the `--args` can be overwrote by `helmDefaults`, which seems to be the opposite of what I expect? 🤔

main.go
```diff
@@ -493,9 +494,69 @@ func clean(state *state.HelmState, errs []error) error {
...
+	}
+
+	state.HelmDefaults.Args = argArr
 
 	return state.HelmDefaults.Args
 }
+
+func setDedicatedKeys(state *state.HelmState, argsMap map[string]string) {
+
+	helmdefaults := reflect.Indirect(reflect.ValueOf(state.HelmDefaults))
```
mumoshu: Nice try! But would you mind if I suggest something like a chain of verbose code like:

```go
if _, exists := argsMap["--tiller-namespace"]; ok && state.HelmDefaults.TillerNamespace != "" {
  argsMap["--tiller-namspace"] = state.HelmDefaults.TillerNamespace
}
```

Or wrap it into a func like:

```golang
setDefaultValue(argsMap, "--tiller-namespace", state.HelmDefaults.TillerNamespace)
setDefaultValue(argsMap, "--kube-context", state.HelmDefaults.KubeContext)
```

The reflection way is definitely DRY, but it seems to require too much engineering effort to maintain compared to the verbose alternatives above, especially when we are to support `--tiller-connection-timeout int` which results in an another field tag (=`arg-type`?) or an another `if`.

main.go
mumoshu: Never mind. You do guard against it at https://github.com/roboll/helmfile/pull/186/files#diff-7ddfb3e035b42cd70649cc33393fe32cR557!

main.go
rmartinez3: Sure ill switch from reflection to more verbose checks. Thanks for looking into pull request. will make updates

rmartinez3:
Made changes. Let me know if others are needed as well. Thanks.

README.md
```diff
@@ -39,6 +39,17 @@ repositories:
...
+#default values to set for args along with dedicated keys that can be set by contributers, cli args take precedence overe these
+helmDefaults:           
+  tillerNamespace: tiller-namespace  #dedicated default key for tiller-namespace
+  kubeContext: kube-context          #dedicated default key for kube-context
+  args:
+    - "--wait"
+    - "--recreate-pods"
+    - "--timeout=600"
+    - "--force"
+    - "--reset-values"
```
mumoshu: I appreciate this detailed documentation ❤️

PR #199, opened 2018-07-25
feat: Optional detailed exitcodes for `helmfile diff`

mumoshu:
Adds the `--detailed-exitcode` to the `helmfile diff` command to return `1` on failure, and `2` when no error but diff is seen.

This feature requires the latest `helm-diff` containing https://github.com/databus23/helm-diff/pull/78, and `helm` containing https://github.com/helm/helm/pull/4367.

This is verified to work by manually running commands like the followings:

```bash
./helmfile --helm-binary helm211dev -f ./examples/helmfile.d diff --detailed-exitcode; echo $?
./helmfile --helm-binary helm211dev -f ./examples/helmfile.d diff; echo $?
```

Note that, in above example commands, `helm211dev` is a custom `helm` binary that is built from helm's master branch containing [the necessary enhancement to allow propagate non-zero plugin exit code](https://github.com/helm/helm/pull/4367).

### DIFF
--- a/main.go
+++ b/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/urfave/cli"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
+	"os/exec"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


mumoshu:
For anyone interested, this also works with [`helm-tiller`](https://github.com/rimusz/helm-tiller) when you run it this way 😉 

```
helm211dev tiller run -- $(pwd)/helmfile --log-level debug --helm-binary helm211dev -f $(pwd)/examples/helmfile.d diff --detailed-exitcode; echo $?
# echoes "2" when there's diff
```

Note that https://github.com/helm/helm/pull/4367 is also required for `helm-tiller`, in order to propagate the exit code `2` from` helmfile diff --detailed-exitcode`.

PR #203, opened 2018-08-01
Load environment variables from .env file

kjdev:
* vendor: add dep (github.com/subosito/gotenv)


sstarcher:
* What's the benefit of supporting .env loading outside of helmfile itself?
* How does/would we want this to work with `helmfile.d`?

kjdev:
We wanted to be able to insert into containers.env via values  ​​(env or requiedEnv) of helmfile as we wanted to set environment from .env in docker-compose (or docker)

> .env:
> ```
> TEST=test
> ```
> __docker-compose__
> docker-compose.yaml:
> ```
> services:
>   test:
>     image: 'image'
>     environment:
>       TEST: "${TEST}"  # docker-compose up ==> test
> ```
>
> __helmfile__
> helmfile.yaml:
> ```
> releases:
> - name: test
>   charts: ./charts
>   values:
>   - env:
>       test: {{ env "TEST" }}
> ```
> charts/templates/deployment.yaml:
> ```
> spec:
>   template:
>     spec:
>       containers:
>       - name: test
>         image: test
>         env:
>         - name: TEST
>           value: {{ $.Values.env.test }} # helmfile sync ==> test
> ```

I wanted to be able to read environment variables from an external file when executing helmfile.

You may be able to specify a file with '--env-file [FILE]' instead of an .env file.

sstarcher:
With something like helmfile it might make more sense to specify a env file.

kjdev:
I tried creating an env-file option to specify a file.

https://github.com/kjdev/helmfile/commit/d0d077d8097d5a2f4ad84bc3066f69606f15aa64

mumoshu:
@kjdev Hey. Thanks for your efforts!
I believe that I understand usefulness of .env files, but what would be difference(s) between this feature and `. .env; helmfile sync` or `bash -c '. .env; helmfile sync`?

kjdev:
@mumoshu
At first I thought about loading .env implicitly like docker(-composer), but if you specify it with `--env-file` then it's the same as`. .env; helmfile sync` It may not be.

mumoshu:
@kjdev Would you mind trying `set -a; . .env; set +a; helmfile sync`?

mumoshu:
@kjdev Thanks for your confirmation 👍

Would you mind if I close this pull request after documenting the `-set -a; ...` thing in README, also with a big acknowledgement and a reference to your awesome work?

aweis89:
@maximmi I think it would be nice to be able specify a specific env file per environment. I've created a wrapper script that intercepts the `--environment` param and uses that load an `<env>.env`. But I'd love to be able create a pull request for this feature. One of the things I love about using `helmfile` is not having to have all these scripts loading env vars etc. Thoughts?

osterman:
We use [direnv](https://direnv.net) with helmfile and kops for this purpose. My preference is a generalized solution than a one-off for `helmfile`. My opinion is based on the observaton that "dot env" file formats vary between authors. I like that `direnv` is consistent and works with most shells.

PR #217, opened 2018-08-22
fix: Make logging on unmatched selectors less verbose

### LINKED ISSUE #200: Change Log Level for Unmatched Selectors
## what
* The current log level for `specified selector did not match any releases` is too verbose (IMO)
* Propose moving it to `debug` or `warn` log levels

```
2018/07/26 01:28:44 specified selector did not match any releases in helmfile.d/0610.kubernetes-dashboard.yaml
```

## why
* If you have a directory with a couple dozen helmfiles, selectors might not apply to most files
* The fact that the selector did not match becomes noisy and is alarming to those unfamiliar

mumoshu:
Fixes #200
Depends on #216 because this feature requires zap.SuggeredLogger to be passed to HelmState, which is done as part of #200

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -607,7 +607,8 @@ func (state *HelmState) FilterReleases(labels []string) error {
 		filteredReleases = append(filteredReleases, r)
 	}
 	if len(filteredReleases) == 0 {
-		return fmt.Errorf("specified selector did not match any releases in %s\n", state.file)
+		state.logger.Debugf("specified selector did not match any releases in %s\n", state.file)
+		return nil
 	}
 	state.Releases = filteredReleases
 	return nil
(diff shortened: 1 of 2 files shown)


state/state.go
```diff
@@ -607,7 +607,8 @@ func (state *HelmState) FilterReleases(labels []string) error {
 		filteredReleases = append(filteredReleases, r)
 	}
 	if len(filteredReleases) == 0 {
-		return fmt.Errorf("specified selector did not match any releases in %s\n", state.file)
+		state.logger.Debugf("specified selector did not match any releases in %s\n", state.file)
+		return nil
```
mumoshu: This broke helmfile's ability to filter the unmatched releases. Returning `nil` here means that you filter nothing.

#225 includes the fix for it.

PR #218, opened 2018-08-22
fix: helmfile should fail on duplicate release name

### LINKED ISSUE #193: Warn upon duplicate selector labels
To avoid human error, we should warn on duplicate labels for identical releases. Give this helmfile:

```yaml
  - name: traefik-ingress
    namespace: platform
    labels:
      minikube: deploy
    chart: stable/traefik
    version: 1.24.1

  - name: traefik-ingress
    namespace: platform
    labels:
      minikube: deploy
    chart: stable/traefik
    version: 1.35.0
```

Executing `helmfile --selector minikube=deploy` will just silently pick the second release (1.35.0). Could we fail or warn?

mumoshu:
Fixes #193

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -109,6 +109,16 @@ func readFromYaml(content []byte, file string, logger *zap.SugaredLogger) (*Helm
 		state.DeprecatedReleases = []ReleaseSpec{}
 	}
 
+	releaseNameCounts := map[string]int{}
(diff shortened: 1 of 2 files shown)


osterman:
https://github.com/roboll/helmfile/issues/193#issuecomment-415296723

PR #222, opened 2018-08-23
feat: `wait: true` in each release

### LINKED ISSUE #65: Declaring a subset of releases for unnecessity of `--wait`
Extracted from https://github.com/roboll/helmfile/pull/60#issuecomment-375728487

Please see the discussion made there for more context.

mumoshu:
To instruct `helm upgrade` to wait for resources via `--wait`

Resolves #65

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -60,6 +60,7 @@ type ReleaseSpec struct {
 	Chart   string `yaml:"chart"`
 	Version string `yaml:"version"`
 	Verify  bool   `yaml:"verify"`
+	Wait    bool   `yaml:"wait"`
 
 	// Name is the name of this release
 	Name      string            `yaml:"name"`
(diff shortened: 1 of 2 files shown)


state/state.go
```diff
@@ -663,14 +651,42 @@ func chartNameWithoutRepository(chart string) string {
...
+	if release.Verify {
+		flags = append(flags, "--verify")
+	}
+	if release.Wait {
+		flags = append(flags, "--wait")
+	}
 	if release.Version != "" {
 		flags = append(flags, "--version", release.Version)
 	}
-	if release.Verify {
```
mumoshu: `verify: true` in helmfile.yaml should have been failing on `helmfile lint` and `helmfile diff` due to that the unsupported `--verify` flag is passed.

This PR also fix that.

Issue #212, opened 2018-08-18
0.21.0 args being reordered


sstarcher:
When running with `--args` the arguments are being non-deterministic rearranged.
* `helmfile sync --args "--timeout=3600 --set image.tag=blah"`

Sometimes out result is
```
exec: helm upgrade --install --reset-values blah stable/chart --version 0.1.0 --namespace=blah image.tag=blah --timeout=3600 --set
Error: flag needs an argument: --set
```

Other times
```
exec: helm upgrade --install --reset-values blah stable/chart --version 0.1.0 --namespace=blah --set image.tag=blah --timeout=3600
```

rmartinez3:
This is caused by the getArgs function 

https://github.com/roboll/helmfile/blob/a5d42e741ba402924017ad7934943b400a705e90/main.go#L553-L570

which can be fixed  error by taking in account of string array flags 

```
        args := c.String("args")
	argsMap := map[string]string{}
	spaceflagArg := map[string]bool{}

	if len(args) > 0 {
		argsVals := strings.Split(args, " ")
		prevFlag := ""
		for _, arg := range argsVals {
			if strings.HasPrefix(arg, "--") {
				argVal := strings.SplitN(arg, "=", 2)
				if len(argVal) > 1 {
					arg := argVal[0]
					value := argVal[1]
					argsMap[arg] = value
				} else {
					argsMap[arg] = ""
				}
				prevFlag = arg
			} else {
				spaceflagArg[prevFlag] = true
				argsMap[prevFlag] = arg
			}
		}
	}
```



```
       var argArr []string

	for key, val := range argsMap {
		if val != "" {
			if spaceflagArg[key] {
				argArr = append(argArr, fmt.Sprintf("%s %s", key, val))
			} else {
				argArr = append(argArr, fmt.Sprintf("%s=%s", key, val))
			} // or

		} else {
			argArr = append(argArr, fmt.Sprintf("%s", key))
		}
	}
```

arguments will be rearranged but wont cause any errors with theses changes. Cause of this is flags are kept in a map to keep track of cli args  and helm defaults args.  So from the map the arguments are build.

I do know there is talk about changing the GetArgs function . Changes can be made in order to capture flags better and merge helm default args.

Stono:
We've just been hit by this non-deterministic behaviour in 0.22.0 too, and have had to roll back to 0.20.0

mumoshu:
[...]

PR #224, opened 2018-08-23
fix for args being reordered

rmartinez3:
This pull requests takes in account of order for cli provided args
- Separates args to its own package and test for args.
- fixes issue https://github.com/roboll/helmfile/issues/212

### DIFF
--- a/args/args.go
+++ b/args/args.go
@@ -1,0 +1,100 @@
+package args
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/roboll/helmfile/state"
+)
+
+type argMap struct {
+	m     map[string]string
+	flags []string
+}
+
+func (a *argMap) SetArg(flag, arg string) {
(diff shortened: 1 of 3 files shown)


args/args.go
```diff
@@ -0,0 +1,100 @@
...
+	if state.HelmDefaults.TillerNamespace != "" {
+		argsMap.SetArg("--tiller-namespace", state.HelmDefaults.TillerNamespace)
+	}
+	if state.HelmDefaults.KubeContext != "" {
+		argsMap.SetArg("--kube-context", state.HelmDefaults.KubeContext)
+	}
+
+	var argArr []string
+
+	for _, flag := range argsMap.flags {
```
mumoshu: Nit, but can we probably just write `range argsMap.m {`, which may indicate that we can use a vanilla golang map instead of `ArgsMap`?

PR #231, opened 2018-08-24
fix: helmfile should fail on duplicate release name after filtered by labels

mumoshu:
This is a follow-up for #218, fixes the unintentional degradation that broken the use-case described in https://github.com/roboll/helmfile/issues/193#issuecomment-415434408


mumoshu:
@osterman You can now safely define multiple releases sharing the same release name. An duplication error occurs when e.g. `helmfile sync -l foo=bar` finds two or more releases named the same, but that's after the label selector was applied to filter releases.

Issue #233, opened 2018-08-26
template fails cause crash


sstarcher:
This happens when using requiredEnv inside of a values file and the env variable is not set.

```
2018/08/26 08:33:34 [bug] this code path shouldn't be arrived: helmfile is expected to exit from within the `cleanup` func in main.go: template: stringTemplate:41:41: executing "stringTemplate" at <requiredEnv "HELM_AC...>: error calling requiredEnv: required env var `HELM_ACCOUNT_ID` is not set
panic: [bug] this code path shouldn't be arrived: helmfile is expected to exit from within the `cleanup` func in main.go: template: stringTemplate:41:41: executing "stringTemplate" at <requiredEnv "HELM_AC...>: error calling requiredEnv: required env var `HELM_ACCOUNT_ID` is not set

goroutine 1 [running]:
log.Panicf(0x144bde3, 0x75, 0xc42015b728, 0x1, 0x1)
	/usr/local/Cellar/go/1.10.3/libexec/src/log/log.go:333 +0xda
main.main()
	/Users/sstarcher/go/src/github.com/roboll/helmfile/main.go:388 +0x17d8
```

sstarcher:
using latest master as of today

PR #235, opened 2018-08-26
improve error handling for value rendering

sstarcher:
fixes - https://github.com/roboll/helmfile/issues/233

Output on values render error
```
err: failed to render [/Users/sstarcher/xxx/xxx/values.yaml], because of template: stringTemplate:10:18: executing "stringTemplate" at <requiredEnv "HELM_AC...>: error calling requiredEnv: required env var `HELM_ACCOUNT` is not set
```

Also removes panic and sets the output as `apps.Run()` can and will return errors.  Panic makes no sense.

### DIFF
--- a/main.go
+++ b/main.go
@@ -384,8 +384,9 @@ func main() {
 	}
 
 	err := app.Run(os.Args)
-	logger.Errorf("%v",err)
-
+	if err != nil {
+		logger.Errorf("%v",err)
+	}
 }
 
 func eachDesiredStateDo(c *cli.Context, converge func(*state.HelmState, helmexec.Interface) []error) error {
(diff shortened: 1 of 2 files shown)


sstarcher:
Not sure what your test is doing that caused the failure.  No functional changes were made just logging.

sstarcher:
ahh pristine is formatting

main.go
```diff
@@ -385,7 +385,7 @@ func main() {
 
 	err := app.Run(os.Args)
 	if err != nil {
-		log.Panicf("[bug] this code path shouldn't be arrived: helmfile is expected to exit from within the `cleanup` func in main.go: %v", err)
+		logger.Errorf("%v", err)
```
mumoshu: Should we `os.Exit(1)` here? Otherwise this seems to return 0 even when there's an error.

PR #242, opened 2018-08-28
feat: Template functions to replace `set`s

### LINKED ISSUE #227: feat: Template functions to replace `set`s
In #207, I and @sstarcher have discussed about if it is a good idea to leverage `--set`, `--set-string` and `--set-file` flags to make it easy to keep feature parity between helm and helmfile.

Conclusion at the moment is to avoid relying on `sets` as much as possible. Instead, we should enhance #216 originally proposed in #97.

In theory, it is matter of adding following template functions so that we can completely avoid using `set`s:

- `readFile` reads the specified local file and generate a golang string
- `unmarshalYaml` reads a golang string and generates a map
- `setValuleAtPath PATH NEW_VALUE` traverses a golang map, replaces the value at the PATH with NEW_VALUE
  - Originally this was name `set` but renamed as such to avoid conflict with `set` from sprig
- `marshalYaml` marshals a map into a string
   - `unmarshalYaml` and `marshalYaml` were named `yamlToMap` and `mapToYaml` respectively. But renamed to respect the golang's convention to use marshal/unmarshal

Expected use-cases of these functions are:

- Instead of `--set foo[0]={bar,baz}`, write a `values.yaml` containing:

```
{{ readFile "another.values.yaml" | unmarshalYaml | setValuleAtPath "foo[0]" (array "bar", "baz") | marshalYaml }}
```

- Instead of `--set-file foo=path/to/file`, write:

```
foo: |
{{ readFile "path/to/file" | indent 2 }}
```

mumoshu:
Resolves #227

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,6 +270,96 @@ The `selector` parameter can be specified multiple times. Each parameter is reso
 
 `--selector tier=frontend --selector tier=backend` will select all the charts
 
+## Templates
+
+You can use go's text/template expressions in `helmfile.yaml` and `values.yaml`(helm values files).
+
+In addition to built-in ones, the following custom template functions are available:
+
+- `readFile` reads the specified local file and generate a golang string
+- `fromYaml` reads a golang string and generates a map
+- `setValuleAtPath PATH NEW_VALUE` traverses a golang map, replaces the value at the PATH with NEW_VALUE
+- `toYaml` marshals a map into a string
+
+### Values Files Templates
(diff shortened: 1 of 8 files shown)


mumoshu:
Releasing this in v0.25.0. Enjoy, and feel free to provide feedbacks on github issues 😉

sstarcher:
@mumoshu thanks for the great work

PR #261, opened 2018-08-30
feat: All the paths are relative to helmfile.yaml

### LINKED ISSUE #246: Paths should be relative to that of helmfile.yaml
In the example below:

```yaml
releases:
- name: myapp
  chart: ./myapp/charts/myapp
  values:
  - ./myapp/values.yaml
  secrets:
  - ./myapp/secrets.yaml
```

I would expect `helmfile -f path/to/helmfile.yaml` to resolve all the paths included in the file to be under the directory `path/to`. As of today helmfile resolves it relative to CWD.

### LINKED ISSUE #252: helmfile lint breaks when local charts are referenced from helmfile.d/
## Problem

When a local chart is referenced in the state file, `helmfile lint` tries to fetch the chart anyway.

This can be observed by using this simple helmfile (local-chart-helmfile.yaml):

```
releases:
- name: my-release
  namespace: default
  chart: ../local-charts/my-local-chart # this 
```

Here is the file layout to reproduce the problem:

```
root
  | local-charts
  `--| my-local-chart
  | helmfile.d
  `--| local-chart-hemlfile.yaml
```
Other helmfile operations seem to work as expected (I tested sync).

## Analysis

After some investigation, this is probably related to a wrong working directory, for the `LintRelease()` function as `isLocalChart()` (a simple os.Stat() with no root dir decoration) called from [LintReleases()](https://github.com/roboll/helmfile/blob/822cc13e72c560bc8ee989cfcbee7d8f98498a7a/state/state.go#L376) returns false for a relative path. 

If I change the `chart:` path in local-chart-helmfile.yaml to be rooted at the working directory, linting completes without error. But this is not a viable solution because other helmfile operations (sync) treat relative paths relative to the helmfile.yaml (or helmfile.d/).

If I change the working directory to `helmfile.d/` and issue `helmfile -f . lint` this also completes without error, which is a temporary workaround.

Note that `helmfile -f helmfile.d/ sync` works as intended as the chart can be reached relative to manifest file, as stated in the README documentation.

Related issue #246

I think a concept of root dir as determined by the `-f` argument, and propagated to every command would solve this problem.

mumoshu:
`helmfile lint` works with relative chart reference (#252)
The tempalte function `readFile` accepts the path relative to helmfile.yaml

Resolves #246
Fixes #252

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -349,7 +349,7 @@ func (state *HelmState) LintReleases(helm helmexec.Interface, additionalValues [
 				}
 
 				chartPath := ""
-				if isLocalChart(release.Chart) {
+				if pathExists(normalizeChart(state.BaseChartPath, release.Chart)) {
 					chartPath = normalizeChart(state.BaseChartPath, release.Chart)
 				} else {
 					fetchFlags := []string{}
@@ -586,14 +586,18 @@ func (state *HelmState) UpdateDeps(helm helmexec.Interface) []error {
 // 	 be constructed relative to the `base path`.
 // - Everything else is assumed to be an absolute path or an actual <repository>/<chart> reference.
 func normalizeChart(basePath, chart string) string {
-	regex, _ := regexp.Compile("^[.]?./")
-	if !regex.MatchString(chart) {
+	if !isLocalChart(chart) {
 		return chart
 	}
 	return filepath.Join(basePath, chart)
 }
 
 func isLocalChart(chart string) bool {
+	regex, _ := regexp.Compile("^[.]?./")
+	return regex.MatchString(chart)
+}
+
+func pathExists(chart string) bool {
 	_, err := os.Stat(chart)
 	return err == nil
 }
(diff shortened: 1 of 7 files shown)


state/state.go
```diff
@@ -586,14 +586,18 @@ func (state *HelmState) UpdateDeps(helm helmexec.Interface) []error {
...
-	regex, _ := regexp.Compile("^[.]?./")
-	if !regex.MatchString(chart) {
+	if !isLocalChart(chart) {
 		return chart
 	}
 	return filepath.Join(basePath, chart)
 }
 
 func isLocalChart(chart string) bool {
+	regex, _ := regexp.Compile("^[.]?./")
```
dtshepherd: @mumoshu Is there a reason a `/` is required to be a local chart?  Its seems valid to have the helmfile.yaml in the same folder as a local chart:

```
├── Chart.yaml
├── charts
│   ├── grafana-1.14.7.tgz
│   ├── influxdb-0.11.0.tgz
│   ├── kapacitor-1.1.0.tgz
│   ├── telegraf-ds-1.7.4.tgz
│   └── telegraf-s-1.7.4.tgz
├── helmfile.d
│   └── helmfile.yaml
```
And the helmfile.yaml:
```
releases:
- name: myrelease
  chart: ..
```
I'd expect the regex to just be `^[.]`.

dtshepherd: For now, I can work around this by adding a trailing slash `/` to my `chart` path.

PR #263, opened 2018-08-30
feat: helmfile apply [--auto-approve]

### LINKED ISSUE #205: Only sync release if there is a difference
Any possibility of adding a feature to run the sync for releases in the helmfile that have differences.

With the feature of `helm diff` I would imagine this would be something that would fit?

mumoshu:
This command syncs releases only if there is any difference between the desired and the current state. It asks for an confirmation by default. Provide `--auto-approve` flag after the `apply` command to skip it.

Resolves #205

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,25 +185,28 @@ USAGE:
 
 COMMANDS:
      repos   sync repositories from state file (helm repo add && helm repo update)
-     charts  sync charts from state file (helm upgrade --install)
-     diff    diff charts from state file against env (helm diff)
+     charts  sync releases from state file (helm upgrade --install)
+     diff    diff releases from state file against env (helm diff)
      lint    lint charts from state file (helm lint)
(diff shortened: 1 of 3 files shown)


kody-abe:
@mumoshu Probably need a README update as well? Thanks for getting this in!

mumoshu:
@kody-abe Thanks for your review! I added some doc updates now.

blissd:
This broke local testing deployments for us as we run `helmfile` non-interactively from scripts. I think this is a bit backwards. After all, I just ran a command to delete stuff, don't make me confirm that I really mean it.

An approach I would prefer would be to restore the original behaviour, but add a `--dry-run` flag that would simply print the commands that would execute instead of actually executing them.

mumoshu:
@blissd See https://github.com/roboll/helmfile/issues/354#issuecomment-423220668 - you aren't forced to use `apply`!

Issue #248, opened 2018-08-29
unknown command --set


Stono:
Hey folks, 
I don't know if this is a follow on from all the arguments issues there have been recently, but building from `master`, a command which should work (and does work on 0.20.0), doesn't:

```
❯ /Users/karl.stoney/go/src/github.com/roboll/helmfile/helmfile sync --args="--set fqdn=.testing.k8.atcloud.io --set global.name=shippr-simple --set global.version=latest --set global.env=testing" .
Updating repo
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "autotrader" chart repository
...Successfully got an update from the "stable" chart repository
...Successfully got an update from the "aws" chart repository
Update Complete. ⎈ Happy Helming!⎈

Decrypting secret /Users/karl.stoney/git/autotrader/conference-application/helmfile/environments/preprod/secrets.yaml
Decrypting /Users/karl.stoney/git/autotrader/conference-application/helmfile/environments/preprod/secrets.yaml

Upgrading aws/platform-helm-at-service
Error: unknown flag: --set global.env
```

Any ideas before I go down a rabbit hole of debugging?

rmartinez3:
You can comma seperate the values,
--args="--set fqdn=.testing.k8.atcloud.io,global.name=shippr-simple,global.version=latest, global.env=testing"

I'll work on supporting multiple flags that are supported on helm 

Thanks for catching this.

Stono:
Hey @rmartinez3 - it doesn't work even with a single argument though:

```
❯ /Users/karl.stoney/go/src/github.com/roboll/helmfile/helmfile sync --args="--set fqdn=.testing.k8.atcloud.io" .
Updating repo
exec: helm repo update
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "aws" chart repository
...Successfully got an update from the "stable" chart repository
...Successfully got an update from the "autotrader" chart repository
Update Complete. ⎈ Happy Helming!⎈

Decrypting secret /Users/karl.stoney/git/autotrader/conference-application/helmfile/environments/preprod/secrets.yaml
--set fqdn=.testing.k8.atcloud.io
exec: helm secrets dec /Users/karl.stoney/git/autotrader/conference-application/helmfile/environments/preprod/secrets.yaml --set fqdn=.testing.k8.atcloud.io
Decrypting /Users/karl.stoney/git/autotrader/conference-application/helmfile/environments/preprod/secrets.yaml

[...]

PR #268, opened 2018-08-31
fix: `--args "--set foo=bar"` always result in "unknown flag: --set foo" error

mumoshu:
Ref #248


mumoshu:
Merging this anyway to move #248 forward

PR #270, opened 2018-08-31
feat: Merge multiple docs inside a single helmfile.yaml

### LINKED ISSUE #254: feat: Merge multiple docs inside a single helmfile.yaml
Extracted from https://github.com/roboll/helmfile/issues/59#issuecomment-416608106

This feature would allow us to template your `helmfile.yaml` according to environment values #253.

An example helmfile.yaml would look like this. Notice the document separator `---` in the middle of the yaml file:

```yaml
environments:
  default:
    values:
    - environments/default.yaml
---
respositories:
- name: your-private-git-repo-hosted-charts
  url: https://{{ .Environment.Values.githubToken }}@raw.githubusercontent.com/kmzfs/helm-repo-in-github/master/

releases:
- name: myapp
  # snip
```

#253 alone doesn't allow you to template your `helmfile.yaml` according to the environment values due to the chicken-and-egg problem. That is, helmfile.yaml needs to be templated before parsed as yaml. But templating the helmfile.yaml referencing `{{ .Environment.Values }}` requires you to load `environments:` form yaml.

This also helps the use-case of implementing ["base" and "layer" helmfiles](https://github.com/roboll/helmfile/issues/240#issuecomment-416886474).

mumoshu:
One caveat is that it doesn't merge arrays, but rather the latter overwrites the former.

Resolves #254

### DIFF
--- a/docs/writing-helmfile.md
+++ b/docs/writing-helmfile.md
@@ -1,0 +1,72 @@
+# The Helmfile Best Practices Guide
+
+This guide covers the Helmfile’s considered patterns for writing advanced helmfiles. It focuses on how helmfile should be structured and executed.
+
+## Layering
+
+You may occasionally end up with many helmfiles that shares common parts like which repositories to use, and whichi release to be bundled by default.
+
+Use Layering to extract te common parts into a dedicated *library helmfile*s, so that each helmfile becomes DRY.
+
+Let's assume that your `helmfile.yaml` looks like:
+
+```
+{ readFile "commons.yaml" }}
+---
+{{ readFile "environments.yaml" }}
+---
+releases:
+- name: myapp
+  chart: mychart
+```
+
+Whereas `commons.yaml` contained a monitoring agent:
+
+```yaml
+releases:
+- name: metricbaet
+  chart: stable/metricbeat
+```
+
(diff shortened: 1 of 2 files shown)


luisdavim:
`imdario/mergo` has options to merge arrays aren't you using that?

mumoshu:
@luisdavim Hey! Actually we're using mergo elsewhere in the helmfile codebase. But AFAICS, `yaml.Unmarshal` merges consumed yaml docs into the targeted desired state in the same way as `mergo`. So, I had no specific reason to use merge here.

I'd appreciate it if could inform me any specific benefit of mergo for this use, if any. Thanks!

luisdavim:
The benefit would be that you can control wither the arrays are merged, appended or overwritten. I think that the current implementation is fine as long as it's clear to the user that the arrays are overwritten, if there's a need to change that mergo could be the solution.

mumoshu:
@luisdavim Makes sense! But you say that mergo supports various strategies to merge arrays? All I know is `merge.WithOverride` which seems to merge arrays by replacing the one in dst with the one in src(=override).

mumoshu:
Also, I'm not sure how everyone wants to merge arrays. Do you need appends, overrides, or both?

PR #278, opened 2018-09-02
WIP: adding to be albe to set multiple flags for --set, -f, --value in args

rmartinez3:


args/args_test.go
```diff
@@ -9,15 +9,15 @@ import (
 )
 
 func TestGetArgs(t *testing.T) {
-	args := "--timeout=3600 --set app1.bootstrap=true,app2.bootstrap=false --tiller-namespace ns"
+	args := "--timeout=3600 --set app1.bootstrap=true --set app2.bootstrap=false --tiller-namespace ns"
```
mumoshu: Does it still work even when a single `--set` had two or more key-values like `--set app1.bootstrap=true --set app2.bootstrap=false,app3.bootstrap=true`?

rmartinez3: go test shows that this test case works. Let me know if theres anything else that needs to be checked.

Issue #283, opened 2018-09-03
Add the ability to do `helm template`


Stono:
It'd be nice to be able to do helm template on the helmfile (pr incoming).

mumoshu:
Interesting! What do you use it for? Example use-cases are welcomed.

Stono:
Hey @mumoshu - so we use helmfile+helm to do all of our deployments to kubernetes in our org.  It's actually wrapped by our deployment API.

We currently use `helmfile sync` with a `--dry-run` and `--debug` to generate the yaml files to do some basically validation in advance of it being actually deployed, for example:

  - Detect known "gotchas" within our platform.  For example, your ports need to be named `http-*` etc
  - Enforce certain standards (labels, etc)

However, this has some downsides; primarily that it depends on having a network connection to kubernetes available, plus it's a bit slow.

We are perfectly happy doing a helm template and then performing our validation on those results instead, does that make sense?

Stono:
Thinking about it too, it also opens up the opportunity for me to extend https://github.com/Stono/helm-test to work with helmfile too (as it parses and asserts on helm template)

mumoshu:
I'm in favor of your use-case https://github.com/roboll/helmfile/issues/283#issuecomment-418077609. I'd appreciate it if you could contribute a README improvement that covers the topic 👍

rplute:
+1

PR #284, opened 2018-09-03
feat: `helmfile template`

Stono:
Hey, 
This PR will do a `helm template` over releases within the helmfile.

It addresses: https://github.com/roboll/helmfile/issues/283

### DIFF
--- a/main.go
+++ b/main.go
@@ -11,6 +11,8 @@ import (
 
 	"os/exec"
 
+	"io/ioutil"
+
 	"github.com/roboll/helmfile/args"
 	"github.com/roboll/helmfile/environment"
 	"github.com/roboll/helmfile/helmexec"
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -449,6 +475,26 @@ func executeSyncCommand(c *cli.Context, state *state.HelmState, helm helmexec.In
...
+
+	if errs := state.UpdateDeps(helm); errs != nil && len(errs) > 0 {
+		return errs
+	}
+
+	if c.GlobalString("helm-binary") != "" {
+		helm.SetHelmBinary(c.GlobalString("helm-binary"))
+	}
+
+	args := args.GetArgs(c.String("args"), state)
```
Stono: You'll notice that unlike the `sync` function, I pass args to `TemplateReleases`, and apply them to `helm` _after_ the fetch has taken place.

This is because if you're using `--set` in your `--args`, they get applied to the `helm fetch` too, which is incorrect and results in an `unknown flag --set` error.

I believe this to be a bug in `--sync` too but I haven't fixed it under this PR

mumoshu:
@Stono LGTM. Thanks for contributing the awesome feature!

Issue #303, opened 2018-09-05
helmfile lint does not add configured repositories


mhulscher:
Unlike `helmfile sync`, `helmfile lint` does not add missing repositories to local Helm. This will cause linting to fail unless the repository was previously configured.

```
$ helmfile --file helmfile.yaml lint
Error: repo incubator not found

err: exit status 1
```

`helmfile.yaml`
```
repositories:
- name: stable
  url: https://kubernetes-charts.storage.googleapis.com
- name: incubator
  url: https://kubernetes-charts-incubator.storage.googleapis.com

releases:
- name: vault
  namespace: vault
  chart: incubator/vault
  version: 0.12.0
```

mhulscher:
This is not fixed in version v0.33.1

mumoshu:
@mhulscher Would you mind sharing a full output of `helmfile --log-level debug lint`? (Beware not to expose repository passwords though!

mhulscher:
This was indeed broken on v0.33.1 but seems to have been fixed in v0.33.2 👍

mumoshu:
@mhulscher Yay! Thanks for your confirmation!

PR #305, opened 2018-09-05
Add missing repos on lint command.

Stono:
Fixes https://github.com/roboll/helmfile/issues/303

### DIFF
--- a/main.go
+++ b/main.go
@@ -13,6 +13,8 @@ import (
 
 	"io/ioutil"
 
+	"strings"
+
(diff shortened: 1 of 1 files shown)


mumoshu:
Awesome! Thanks for your contribution.

@sstarcher @osterman We're changing the default behavior of `helmfile lint` to run `repos` beforehand. Are you ok with that, or perhaps you need to add `--skip-repo-update` like we did for `helmfile  apply`?

PR #307, opened 2018-09-05
ensure helmfiles are searched relative to base helmfile

### LINKED ISSUE #306: using helmfiles: key with relative paths is broken
When you run helmfile on a content like below from a foreign working directory, helmfile fails to find the globbed files.

helmfile.yaml
```
environments:
  staging:
  production:


helmfiles:
    - services/*/helmfile.yaml
```

davidovich:
This allows calls from an unrelated directory:

```
cd
helmfile -f ~/project/helmfile.yaml
```

Fixes #306


mumoshu:
Ah, good catch! Thank you so much for your contribution @davidovich.

Issue #297, opened 2018-09-05
Referencing environment values from within helmfile.yaml


mumoshu:
#267 supports referencing `.Environment.Name` from within helmfile.yaml, but `.Environment.Values`. The latter can be referenced from within values file templates only as of today.

Looking into use-cases like @Stono's https://github.com/roboll/helmfile/issues/243#issuecomment-418386220, I can't stop imagining how helmfile could reduce envvar references, replacing with  helmfile's env values.

For example, how about exploiting [```{{` STR `}}``` expressions](https://golang.org/pkg/text/template/#hdr-Examples) to nest templates?

The below example adds `template:` under each item in `releases`, so that the yaml string value for `template` is considered gotmpl that is rendered after `.Environment.Values` are loaded.

```yaml
releases:
- template: |{{`
    name: sauron-web
    namespace: sauron-web
    chart: aws/platform-helm-at-service
    version: 0.1.{{ .Environment.Values.goDependencyLabelPlatform" }}
    values:
    - environments/values.yaml
    - environments/{{ .Environment.Name }}/values.yaml
    secrets:
    - environments/{{ .Environment.Name }}/secrets.yaml
    set:
    - name: deployment.containers[0].image.tag
      value: {{ requiredEnv "GO_PIPELINE_COUNTER" }}
`}}
```

You many prefer accepting the release template without introducing the new key `template:`:

```yaml
releases:
- |{{`
  name: sauron-web
  namespace: sauron-web
  chart: aws/platform-helm-at-service
  version: 0.1.{{ .Environment.Values.goDependencyLabelPlatform" }}
  values:
  - environments/values.yaml
  - environments/{{ .Environment.Name }}/values.yaml
  secrets:
  - environments/{{ .Environment.Name }}/secrets.yaml
  set:
  - name: deployment.containers[0].image.tag
    value: {{ requiredEnv "GO_PIPELINE_COUNTER" }}
`}}
```

Or maybe just allow templating per value?
In the below example, only the value for `version` is templated.

```yaml
releases:
- name: sauron-web
  namespace: sauron-web
  chart: aws/platform-helm-at-service
  version: 0.1.{{` {{.Environment.Values.goDependencyLabelPlatform}} `}}
  values:
  - environments/values.yaml
  - environments/{{ .Environment.Name }}/values.yaml
  secrets:
  - environments/{{ .Environment.Name }}/secrets.yaml
  set:
  - name: deployment.containers[0].image.tag
[...]

PR #308, opened 2018-09-05
double render the helmfile

davidovich:
This allows using the values defined in the environments: section.

This is done by using a buffer instead of re-reading the file twice.

see #254

@mumoshu can you weigh in on the approach ?

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"fmt"
 	"log"
 	"os"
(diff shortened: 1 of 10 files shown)


sstarcher:
Seems like a plausible way to support the values directly.

main.go
```diff
@@ -554,10 +554,22 @@ func findAndIterateOverDesiredStates(fileOrDir string, converge func(*state.Helm
 	noTargetFoundForAllHelmfiles := true
 	for _, f := range desiredStateFiles {
 		logger.Debugf("Processing %s", f)
-		yamlBuf, err := tmpl.NewFileRenderer(ioutil.ReadFile, "", environment.EmptyEnvironment).RenderTemplateFileToBuffer(f)
+
+		fileRenderer := tmpl.NewFileRenderer(nil, "", environment.EmptyEnvironment)
```
mumoshu: I suspect if the `nil` as the first argument here makes any `readFile` call inside helmfile.yaml fail at first pass.

mumoshu: Fixed it in 2072ede

mumoshu:
I've added some debug logs while fixing the `readFile` issue that would also help users.

mumoshu:
One gotcha is that something like the below blows up helmfile on first-pass render with `error calling eq: invalid type for comparison`:

```
{{ if eq .Environment.Values.foo "bar" }}
...
{{ end }}
```

because the environment value isn't loaded yet hence `nil`.

A work-around is to modify it by adding a nil-check:

```
{{ if (.Environment.Values.foo) and eq .Environment.Values.foo "bar" }}
...
{{ end }}
```

elemental-lf:
>  I thought that helmfile would end up executing {{ readFile }} and {{ exec }} expressions in the template twice, that is costly in computation and io (in theory), but it doesn't sound critical to me.
Could this lead to a difference in rendering if readFile or exec would return different information on the second call (think about a timestamp or a randomly generated password)? Some parts of the template using the value from the first call and some parts of the template using the value from the second call for example?

davidovich:
Yes, if template functions have side effects, they would be executed twice. I am searching for a way to ignore, or make null operations for the first pass, as this first pass is only meant to parse the environment which is to be re-injected. 

I was not aware of the new execution functions in template rendering, I will try to find a solution.

sstarcher:
ya, due to the nature of certain functions this might not be a workable solution, unless for the first pass you overrode what the template functions do with mocks.

davidovich:
@mumoshu I resolved errors and side-effects happening on the initial render. Also, 

```
{{ if (.Environment.Values.foo) and eq .Environment.Values.foo "bar" }}
...
{{ end }}
```
is resolved. You can see this in main_test.go

mumoshu:
Thanks @davidovich. It looks awesome!

One nit I found is that whether you should also stub `readFile` and `exec` for environment values file templates or not. For example, `foo.yaml.gotmpl` inside the `environments.ENV.values` array allows you to template environment values.

Not sure what's the less confusing way yet.

sstarcher:
My overall concern is all of the edge cases this might bring up and any confusion it might cause people with how helmfile works.  If everyone is comfortable with those things then I'm ok.

mumoshu:
> Also, 
> 
> ```
> {{ if (.Environment.Values.foo) and eq .Environment.Values.foo "bar" }}
> ...
> {{ end }}
> ```
> is resolved. You can see this in main_test.go

@davidovich Hey! I took sometime to review it again, and noticed that you're testing the opposite of the issue I found.

It was that, the template execution fails when you had no values set in [`default/values.yaml` ](https://github.com/roboll/helmfile/pull/308/files#diff-53618cae32e23ae47e1c35f10da92654R107). It happens because any missing value is considered a specially typed "<no value>" value in templates, so for example a comparison between the specially typed value and a string fails with type error `error calling eq: invalid type for comparison`.

You'll probably want to define a `emptyValuesYaml := []byte{}` instead of `defaultValuesYalm` and feed it to the test.

davidovich:
> You'll probably want to define a emptyValuesYaml := []byte{} instead of defaultValuesYalm and feed it to cover this case in your test.

You are correct, this fails, but wouldn't it fail even if we didn't have a double render ? Is the concern related to less obvious failure mode ? I have added an `err` evaluation after the second render. we get the correct error there.

mumoshu:
@davidovich Thanks for your continuous effort!

> but wouldn't it fail even if we didn't have a double render ? Is the concern related to less obvious failure mode 

Yeah, I agree. To provide you more context, @sstarcher and I had also concluded so #317.
Setting `missingkey=error` for every template used in helmfile makes sense to me now, so that the above example fails on missing key, rather than the non-obvious type error.

Does that sound good to you, too?

mumoshu:
> Setting missingkey=error for every template used in helmfile makes sense to me

Oops. In the context of double rendering, I suppose, this should be "set `missingkey=zero` for the first render, and `missingkey=error` for the second render".

Otherwise the first render against a helmfile that contains something like `{{ if (.Environment.Values.foo) and eq .Environment.Values.foo "bar" }}` fails due to a missing key error.

davidovich:
@mumoshu I have implemented the `missingkey=error` as you suggested.

I came back to this comment: 
> One nit I found is that whether you should also stub readFile and exec for environment values file templates or not.

and added a test to see the behavior. The current code as you wrote it supports nicely this usecase. If you have a .gotmpl file, all render functions are upheld and allow you to use readFile and exec in the environment section ! I think this is aligned with your expectations, and it makes a lot of sense.

davidovich:
Tell me if everything is ok, and I will rebase on master.

sstarcher:
I tested this off your branch and I had no issue with the double rendering.  I'm having another issue with helmfile, but I don't think it's related to your rendering.

my rending in a helmfile is very simple so I'm confident your implementation will work for me.

mumoshu:
This resolves #297

tmpl/tmpl.go
```diff
@@ -11,7 +11,13 @@ func (c *Context) stringTemplate() *template.Template {
 	for name, f := range c.createFuncMap() {
 		funcMap[name] = f
 	}
-	return template.New("stringTemplate").Funcs(funcMap)
+	tmpl := template.New("stringTemplate").Funcs(funcMap)
+	if c.preRender {
+		tmpl.Option("missingkey=zero")
+	} else {
+		tmpl.Option("missingkey=error")
```
mumoshu: This seems to resolve #317, too.

PR #313, opened 2018-09-07
add devel flag support

sstarcher:


sstarcher:
This needs - https://github.com/databus23/helm-diff/pull/92/

state/state.go
```diff
@@ -798,6 +803,10 @@ func (state *HelmState) flagsForUpgrade(helm helmexec.Interface, release *Releas
 		flags = append(flags, "--version", release.Version)
 	}
 
+	if release.Devel != nil && *release.Devel || state.HelmDefaults.Devel {
```
mumoshu: I didn't notice it when I've written similar code for Verify below, but what happens when `helmDefaults.devel` is set to `true` but `releases[].devel` is explicitly set to `false` rather than omitted?

Should we write this `if release.Devel != nil && *release.Devel || ((release.Devel == nil || release.Devel != nil && !*release.Devel) && state.HelmDefaults.Devel)`?

sstarcher:
@mumoshu added a test specifically for that and fixed the logic.

PR #339, opened 2018-09-14
feat: delete in reverse order

### LINKED ISSUE #312: Ordering of charts on delete: List should be reversed
When using ``helmfile.d`` and multiple helmfiles they are ordered lexicographically so that user can define a certain installation order. I think this list should be reversed when doing a ``helmfile delete``. The same applies when using glob patterns.

mumoshu:
Resolves #312


mumoshu:
Thanks for your review @davidovich ☺️

PR #340, opened 2018-09-14
fix: readFile and exec should run in the directory the template file resides in

### LINKED ISSUE #299: Go template "exec" function calls programs relative to current working directory
First of all: Thanks for helmfile and your fast and very user oriented development style. It is becoming more useful with every day. Your work is appreciated!

From my tests it seems that the new "exec" function calls programs relative to the current working directory which is different for example to the way value files (normal and *.gotmpl) are included or "readFile" works. I'd like to suggest two possible other options:
1. Call the program relative to the helmfile the Go template was included from
2. Call the program relative to the Go template
As "readFile" already uses the first option, "exec" should probably follow.

I'm using 0.33.0.

mumoshu:
Fixes #299

### DIFF
--- a/main.go
+++ b/main.go
@@ -607,7 +607,7 @@ type twoPassRenderer struct {
 
 func (r *twoPassRenderer) renderEnvironment(content []byte) environment.Environment {
 	firstPassEnv := environment.Environment{Name: r.env, Values: map[string]interface{}(nil)}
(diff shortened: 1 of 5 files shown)


davidovich:
LGTM, There should probably be a change to docs also to document this change.

PR #343, opened 2018-09-14
Use --namespace value within Helmfile template

### LINKED ISSUE #326: Use --namespace value within Helmfile template
Use case: I want to interpolate the namespace specified on the Helmfile command-line via `--namespace` into the release name in a `helmfile.yaml`. My motivation is that I'm trying to prefix all Helm release names with the name of the namespace that they correspond to, because I've got a global cluster-wide Helm Tiller and `helm list` is kind of useless if I can't tell at a glance which namespace a particular release belongs to. And perhaps more importantly, I want to avoid Helm release name collisions if the same thing is deployed to multiple different namespaces in a cluster.

Here's the sort of syntax I'm imagining:

```
releases:
  - name: {{ .Namespace }}-my-chart
```

Please let me know if this isn't possible for some reason, e.g. the namespace isn't known at release name templating time.

I know that I could use the environment variable feature for this, but that would be duplicative with `--namespace`.

witten:
Resolves #326.

**Comments**

This is my first Go, so let me know if there are any obvious rookie mistakes you'd like fixed. Also, let me know if you'd like to see any additional automated testing.

Note that I did *not* implement plumbing the namespace into *values* file templating, so please let me know if you think that should be part of this.

**Testing done**

Beyond running automated tests, I also put together a test `helmfile.yaml` containing `{{ .Namespace }}` and manually ran `helmfile sync` on it. I confirmed that:

1. When `--namespace` is provided, it gets plumbed through as the value for `{{ .Namespace }}`.
2. When `--namespace` is omitted, `{{ .Namespace }}` is an empty string.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -41,11 +41,11 @@ func TestReadFromYaml_DuplicateReleaseName(t *testing.T) {
 	}
 }
 
-func makeRenderer(readFile func(string) ([]byte, error), env string, namespace string) *twoPassRenderer {
+func makeRenderer(readFile func(string) ([]byte, error), env string) *twoPassRenderer {
 	return &twoPassRenderer{
 		reader:    readFile,
 		env:       env,
-		namespace: namespace,
+		namespace: "namespace",
 		filename:  "",
 		logger:    logger,
 		abs:       filepath.Abs,
@@ -80,7 +80,7 @@ releases:
 		return []byte(""), nil
 	}
 
-	r := makeRenderer(fileReader, "staging", "namespace")
+	r := makeRenderer(fileReader, "staging")
 	yamlBuf, err := r.renderTemplate(yamlContent)
 	if err != nil {
 		t.Errorf("unexpected error: %v", err)
@@ -131,7 +131,7 @@ releases:
 		return defaultValuesYaml, nil
 	}
 
-	r := makeRenderer(fileReader, "staging", "namespace")
+	r := makeRenderer(fileReader, "staging")
 	// test the double rendering
 	yamlBuf, err := r.renderTemplate(yamlContent)
 	if err != nil {
@@ -180,7 +180,7 @@ releases:
 		return defaultValuesYaml, nil
 	}
 
-	r := makeRenderer(fileReader, "staging", "namespace")
+	r := makeRenderer(fileReader, "staging")
 	// test the double rendering
 	_, err := r.renderTemplate(yamlContent)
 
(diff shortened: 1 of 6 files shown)


witten:
Also, I'm unclear on while `make pristine` is failing in CI. Any help on that would be appreciated.

witten:
I think I figured out the `make pristine` thing.. Go whitespace is weird.

main_test.go
```diff
@@ -41,13 +41,14 @@ func TestReadFromYaml_DuplicateReleaseName(t *testing.T) {
 	}
 }
 
-func makeRenderer(readFile func(string) ([]byte, error), env string) *twoPassRenderer {
+func makeRenderer(readFile func(string) ([]byte, error), env string, namespace string) *twoPassRenderer {
```
davidovich: this is a test helper, and I see that you do not use the parameter, what do you think about hard-coding it in the struct creation?

```
return &twoPassRenderer{
		reader:    readFile,
		env:       env,
		namespace: "namespace",
		filename:  "",
		logger:    logger,
		abs:       filepath.Abs,
	}

File: tmpl/file.go [hunk 2]
```diff
@@ -16,13 +16,14 @@ type templateFileRenderer struct {
 type TemplateData struct {
 	// Environment is accessible as `.Environment` from any template executed by the renderer
 	Environment environment.Environment
+	Namespace   string
```
davidovich: Adding a comment on a exported identifier is a good Go practice, what do you think about adding one here ?

tmpl/file.go
```diff
@@ -15,6 +15,7 @@ type templateFileRenderer struct {
 
 type TemplateData struct {
 	// Environment is accessible as `.Environment` from any template executed by the renderer
+	// Namespace is accessible as `.Namespace` from any non-values template executed by the renderer
```
davidovich: For go doc to operate correctly, this has to be on top of the documented field.

witten: Huh okay. Fixed!

valuesfile/valuesfile.go
```diff
@@ -15,7 +15,7 @@ type renderer struct {
 func NewRenderer(readFile func(filename string) ([]byte, error), basePath string, env environment.Environment) *renderer {
 	return &renderer{
 		readFile:         readFile,
-		tmplFileRenderer: tmpl.NewFileRenderer(readFile, basePath, env),
+		tmplFileRenderer: tmpl.NewFileRenderer(readFile, basePath, env, ""),
```
mumoshu: We should probably make `{{ .Namespace }}` in values.yaml.gotmpl files fail rather than returning an empty string. But that's an another story! Let me just note about it here as a reminder

mumoshu:
Rerunning the failed CircleCI build believing it was a flake.

valuesfile/valuesfile.go
witten: Okay. Please let me know if you'd like me to make any changes here!

mumoshu:
Looks good. Thanks a lot for your supports @witten and @davidovich!

PR #345, opened 2018-09-15
Fix small typo in error string in state.go

terrycorley:


mumoshu:
Thanks for your supports @terrycorley and @davidovich!

PR #349, opened 2018-09-19
feat: `prepare` and `cleanup` release event hooks

### LINKED ISSUE #295: feat: Hooks
I imagine that you use helmfile in combination with various tools 😃
 
I'd like to investigate the possibility to enhance `helmfile.yaml`, so that you can also declaratively manage which tools are executed how, on which timing, and so on. And I'd like to achieve it without implementing so many features into the helmfile core.

#214 is one of feature requests relevant to this feature. #214 overrides specific helmfile internal task to use the specified command. So that helmfile could run any interchangeable command instead of e.g. `helm upgrade --install` to install/upgrade a helm release.

Hooks, on the other hand, would allow you to prepend or append arbitrary logics or commands to be executed along with a specific helmfile task.

For example:

- `preHelmUpgrade` hook would allow you to add commands to be executed before `helm upgrade --install` while in `helmfile sync`
- `preChartLoad` would allow you to run arbitrary commands in order to auto-fix or auto-generate the chart that is used for installation
- `preApplyRun` would allow you to run arbitrary command before `helmfile apply`. Use it to e.g. automatically retrieve a token to authenticate against the kubernetes cluster that corresponds to the helmfile environment (#267).

Consider it as something like [draft tasks](https://github.com/Azure/draft/blob/e657813f23d23b967ec99a6363533fc3f1253312/docs/reference/dep-008.md#motivation) or [helm2 plugin hooks](https://github.com/helm/helm/blob/master/pkg/plugin/hooks.go), or [helm3 lifecycle event handlers](https://github.com/helm/community/blob/master/helm-v3/002-events.md#scripting-event-handlers).

Please share your use-cases! Thanks.

### LINKED ISSUE #324: Use values/secrets from within a remote chart
My use case: I'm using Helm charts hosted on a private package server not only as a means of packaging up and distributing Kubernetes templates, but also for distributing associated per-environment values and secrets. This means that our helm-secrets files are included in the charts themselves. The main benefit of an approach like this versus keeping the secrets separately is that we don't need a source code checkout (or more realistically, n source code checkouts, one per chart) to deploy. Instead, we can just use the remote chart on the chart server (effectively a build artifact) and deploy it directly.

Where Helmfile comes in is that I'd like to declare several Helm charts to deploy in a `helmfile.yaml`'s `releases:`, and have Helmfile able to pull the relevant values and secrets out of *remote* charts during deployment (`helmfile sync`). Under the hood, this might work by first `helm fetch`ing each remote chart so that the values and secrets files can be accessed.

Also, there's the question of how to refer to those values/secrets. As far as I know, `values:` and `secrets:` currently are interpreted as ~absolute paths or~ paths relative to the current directory. What we'd need for this feature request to work is a path relative to the root of the remote chart.

First, is something like this possible today? Perhaps with [`exec`](https://github.com/roboll/helmfile/#importing-values-from-any-source) or some other mechanism? If not, does this make sense as a feature request?

### LINKED ISSUE #329: feat: Per-value templateing in desired state files (helmfile.yaml)
Instead of templating the entire file, there seem to be use-cases that requires additional layer of templating.

One of examples is `preChartLoad` hooks:

- https://github.com/roboll/helmfile/issues/295#issuecomment-419870215
- https://github.com/roboll/helmfile/issues/324#issuecomment-420108479

Let's gather more use-cases like that to shape what this feature should provide.

### LINKED ISSUE #330: feat: preChartLoad hooks
A `preChartLoad` hook allows you to run an arbitrary command before helmfile runs a helm command that requires the existence of the chart.

```yaml
releases:
- ...
  hooks:
    preChartLoad: # register as many commands as you want to this hook
    - command: {{` ./chartify -e {{ .Environment.Name }} {{ .Chart }} `}}
```

Use-cases:

- https://github.com/roboll/helmfile/issues/295#issuecomment-419870215
- https://github.com/roboll/helmfile/issues/324#issuecomment-420108479

This depends on the per-value templating feature in helmfile.yaml as described in #329.
If we are to implement this at the first time, probably we should enable per-value templating only on chart hook commands. It will be enough for preChartLoad hooks, YAGNI, and could become a base of further discussions about how we'd evolve the per-value templating feature.

mumoshu:
Resolves #295
Resolves #330
Resolves #329 (Supports templating of only `releases[].hooks[].command` and `args` right now
Resolves #324

### DIFF
--- a/README.md
+++ b/README.md
@@ -593,6 +593,110 @@ mysetting: |
 
 The possibility is endless. Try importing values from your golang app, bash script, jsonnet, or anything!
 
+## Hooks
+
+A helmfile hook is a per-release extension point that is composed of:
+
+- `events`
+- `command`
+- `args`
+
+helmfile triggers various `events` while it is running.
+Once `events` are triggered, associated `hooks` are executed, by running the `command` with `args`.
+
+Currently supported `events` are:
+
+- `prepare`
+- `cleanup`
+
+Hooks associated to `prepare` events are triggered after each release in your helmfile is loaded from YAML, before executed.
+
+Hooks associated to `cleanup` events are triggered after each release is processed.
+
+The following is an example hook that just prints the contextual information provided to hook:
+
+```
+releases:
+- name: myapp
+  chart: mychart
+  # *snip*
+  hooks:
+  - events: ["prepare", "cleanup"]
+    command: "echo"
+    args: ["{{`{{.Environment.Name}}`}}", "{{`{{.Release.Name}}`}}", "{{`{{.HelmfileCommand}}`}}\
+"]
+```
+
+Let's say you ran `helmfile --environment prod sync`, the above hook results in executing:
+
+```
+echo {{Environment.Name}} {{.Release.Name}} {{.HelmfileCommand}}
+```
+
+Whereas the template expressions are executed thus the command becomes:
+
+```
+echo prod myapp sync
+```
+
+Now, replace `echo` with any command you like, and rewrite `args` that actually conforms to the command, so that you can integrate any command that does:
+
+- templating
+- linting
+- testing
+
+For templating, imagine that you created a hook that generates a helm chart on-the-fly by running an external tool like ksonnet, kustomize, or your own template engine.
+It will allow you to write your helm releases with any language you like, while still leveraging goodies provided by helm.
+
+### Helmfile + Kustomize
+
+Do you prefer `kustomize` to write and organize your Kubernetes apps, but still want to leverage helm's useful features
+like rollback, history, and so on? This section is for you!
+
+The combination of `hooks` and [helmify-kustomize](https://gist.github.com/mumoshu/f9d0bd98e0eb77f636f79fc2fb130690)
+enables you to integrate [kustomize](https://github.com/kubernetes-sigs/kustomize) into helmfle.
+
+That is, you can use `kustommize` to build a local helm chart from a kustomize overlay.
+
+Let's assume you have a kustomize project named `foo-kustomize` like this:
+
+```
+foo-kustomize/
+├── base
+│   ├── configMap.yaml
+│   ├── deployment.yaml
+│   ├── kustomization.yaml
+│   └── service.yaml
+└── overlays
+    ├── default
+    │   ├── kustomization.yaml
+    │   └── map.yaml
+    ├── production
+    │   ├── deployment.yaml
+    │   └── kustomization.yaml
+    └── staging
+        ├── kustomization.yaml
+        └── map.yaml
+
+5 directories, 10 files
+```
+
+Write `helmfile.yaml`:
+
+```yaml
+- name: kustomize
+  chart: ./foo
+  hooks:
+  - events: ["prepare", "cleanup"]
+    command: "../helmify"
+    args: ["{{`{{if eq .Event.Name \"prepare\"}}build{{else}}clean{{end}}`}}", "{{`{{.Release.Ch\
+art}}`}}", "{{`{{.Environment.Name}}`}}"]
+```
+
+Run `helmfile --environment staging sync` and see it results in helmfile running `kustomize build foo-kustomize/overlays/staging > foo/templates/all.yaml`.
+
+Voilà! You can mix helm releases that are backed by remote charts, local charts, and even kustomize overlays.
+
 ## Using env files
 
 helmfile itself doesn't have an ability to load env files. But you can write some bash script to achieve the goal:
--- a/helmexec/runner.go
+++ b/helmexec/runner.go
@@ -15,10 +15,13 @@ type Runner interface {
 }
 
 // ShellRunner implemention for shell commands
-type ShellRunner struct{}
+type ShellRunner struct {
+	Dir string
+}
 
 // Execute a shell command
 func (shell ShellRunner) Execute(cmd string, args []string) ([]byte, error) {
 	preparedCmd := exec.Command(cmd, args...)
+	preparedCmd.Dir = shell.Dir
 	return preparedCmd.CombinedOutput()
 }
--- a/event/bus.go
+++ b/event/bus.go
@@ -1,0 +1,101 @@
+package event
+
+import (
+	"fmt"
+	"github.com/roboll/helmfile/environment"
+	"github.com/roboll/helmfile/helmexec"
+	"github.com/roboll/helmfile/tmpl"
+	"go.uber.org/zap"
+	"os"
+)
+
+type Hook struct {
+	Name    string   `yaml:"name"`
+	Events  []string `yaml:"events"`
+	Command string   `yaml:"command"`
+	Args    []string `yaml:"args"`
+}
+
+type event struct {
+	Name string
+}
+
+type Bus struct {
+	Runner helmexec.Runner
+	Hooks  []Hook
+
+	BasePath      string
+	StateFilePath string
+	Namespace     string
+
+	Env environment.Environment
+
+	ReadFile func(string) ([]byte, error)
+	Logger   *zap.SugaredLogger
+}
+
+func (bus *Bus) Trigger(evt string, context map[string]interface{}) (bool, error) {
+	if bus.Runner == nil {
+		bus.Runner = helmexec.ShellRunner{
+			Dir: bus.BasePath,
+		}
+	}
+
+	executed := false
+
+	for _, hook := range bus.Hooks {
+		contained := false
+		for _, e := range hook.Events {
+			contained = contained || e == evt
+		}
+		if !contained {
+			continue
+		}
+
+		var err error
+
+		name := hook.Name
+		if name == "" {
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -593,6 +593,61 @@ mysetting: |
 
 The possibility is endless. Try importing values from your golang app, bash script, jsonnet, or anything!
 
+## Hooks
+
+Hook is an per-release extension point of helmfile that is composed of:
```
davidovich: What about this formulation?
```
A helmfile hook is a per-release extension point that is composed of:
```

README.md
```diff
@@ -593,6 +593,61 @@ mysetting: |
...
+
+- `events`
+- `command`
+- `args`
+
+helmfile triggers various `events` while it is running.
+One `events` are triggered, associated `hooks` are executed, by running the `command` with `args`.
```
davidovich: s/One/Once/

README.md
```diff
@@ -593,6 +593,61 @@ mysetting: |
...
+
+The following is an example hook that just prints the contextual information provided to hook:
+
+```
+releases:
+- name: myapp
+  chart: mychart
+  # *snip*
+  hooks:
+  - events: ["prepare", "cleanup"]
```
davidovich: Is there a technical reason you didn't go with this the structure proposed in #330?:

```
 hooks:
    preChartLoad: # register as many commands as you want to this hook
    - command: {{` ./chartify -e {{ .Environment.Name }} {{ .Chart }} `}}
```

I see that you would want to register the same command for a specific hook, but that seems to be over generic (I believe it is more common to have differing calls on each event, than the same call on multiple events).

YAML anchors could come handy if you have the same command on multiple events.

The `preChartLoad` schema is verifiable at load time.

davidovich:
small nit: (title contains a typo: `faet`) and one issue will not be closed: `Resolevs #324`

README.md
mumoshu: Sounds great!

README.md
```diff
@@ -593,6 +593,110 @@ mysetting: |
...
+    command: ["../helmify"]
+    args: ["{{`{{if eq .Event.Name \"prepare\"}}build{{else}}clean{{end}}`}}", "{{`{{.Release.Ch\
+art}}`}}", "{{`{{.Environment.Name}}`}}"]
+```
+
+Run `helmfile --environment staging sync` and see it results in helmfile running `kustomize build foo-kustomize/overlays/staging > foo/templates/all.yaml`.
+
+Viola! You can mix helm releases that are backed by remote charts, local charts, and even kustomize overlays.
```
davidovich: s/Viola/Voilà/

helmexec/runner.go
```diff
@@ -15,10 +15,15 @@ type Runner interface {
...
-type ShellRunner struct{}
+type ShellRunner struct {
+	Dir string
+}
 
 // Execute a shell command
 func (shell ShellRunner) Execute(cmd string, args []string) ([]byte, error) {
 	preparedCmd := exec.Command(cmd, args...)
+	if shell.Dir != "" {
+		preparedCmd.Dir = shell.Dir
```
davidovich: What do you think about setting `preparedCmd.Dir` unconditionally? It's default value is the empty string.

helmexec/runner.go
```diff
@@ -15,10 +15,15 @@ type Runner interface {
...
+	}
 	return preparedCmd.CombinedOutput()
```
davidovich: Are we sure we want to combine the stdout and stderr here? If my subcommand behaves correctly with regards to what it puts in stdout (meant to communicate info in a pipe) and stderr (meant for user info/errors), `CombinedOutput()` would break further pipeline building (as it would possibly interlace program ouptut and errors in the stdout).

README.md
mumoshu: Good catch! Indeed my intention was to make it even more generic with (hopefully) comparable complexity to the third option in [my proposal](https://github.com/roboll/helmfile/issues/295#issuecomment-419870215).

> I believe it is more common to have differing calls on each event, than the same call on multiple events

I think so too. My intention here was to provide the most DRY way to support this use-case:

> To be able to use the same script to implement multiple hooks the name of the hook should also be passed along.

See https://github.com/roboll/helmfile/issues/295#issuecomment-419051383 for more context.

> The preChartLoad schema is verifiable at load time.

Good point! I was thinking to add validation for unsupported events. That is, `events: ["foo"]` results in an error like `unsupported event to hook: foo`.

README.md
mumoshu: Good catch! Thanks. Fixed in the latest commit.

README.md
mumoshu: And I wasn't sure if "events` are the only possible hook target in helmfile. Googling around, the term `hook` seems to be used for NOT only for expressing extension points associated to events. For example, something like "action hook" can be considered:

```
hooks:
- actions: ["lint"]
   command: ...
```

An event hook reacts on specific event and doesn't have ability to return data from the hook to helmfile.

An action hook, on the other hand, could be used for passing  data from the hook to helmfile, in an output format helmfile expects for the specific action. In the above example, the `lint` action hook can return a specifically formatted json to add lint results that are integrated into the original output of `helmfile lint`.

Explicitly wording hook targets as `events` prevents ambiguity and creates future possibility like this.

state/hook.go
```diff
@@ -0,0 +1,107 @@
...
+			args[i], err = render.RenderTemplateText(raw)
+			if err != nil {
+				return false, fmt.Errorf("hook[%s]: %v", name, err)
+			}
+		}
+
+		allargs := []string{}
+		allargs = append(allargs, command[1:]...)
+		allargs = append(allargs, args...)
+		bytes, err := state.runner.Execute(command[0], allargs)
```
davidovich: command is a slice, but only the first one is executed. How does that work if the user specifies multiple commands ?

```
command: ["echo", "helm"]
args: ["hello", "world", "list"]
```

What do you think about these UIs?:

```
command: ["command", "arg1", "arg2"]
```
or
```
commands:
  - cmd: "helm"
    args: "list"
```

helmexec/runner.go
mumoshu: I believe we shouldn't, because we use it to run helm commands and it assumes Dir unspecified today.

helmexec/runner.go
mumoshu: As this func isn't used in places that are relevant to piping, it would be ok. We do differente stdout/err in other places, like impl of the `exec` template func.

state/hook.go
mumoshu: Trailing items in the command array is considerd args. I made it so to make it consistent with k8s pod spec, hoping it is easy to remember along with it.

helmexec/runner.go
davidovich: What I meant is that the condition is not necessary:

if shell.Dir is empty, assign (it won't change the default of preparedCmd which is empty)
if shell.Dir is not empty, assign (as you you in the condition)

helmexec/runner.go
davidovich: What about piping helmfile output to other commands ?

mumoshu: a possible usecase! however helmfile as of today emits too verbose logs, that only differentiating stdout/err here does't make sense. good for another issue though

helmexec/runner.go
mumoshu: Ah make sense! I'll fix it asap. Thanks for your review!

state/hook.go
davidovich: OK, but the podSpec is [meant to override CMD and ENTRYPOINTS of containers](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes). Not sure the model applies. Your call :-)

davidovich:
I believe this change could benefit from additional tests.

state/hook.go
mumoshu: Good point! Maybe blindly trying to make it consistent with pod spec doesn't make sense. I'll just make it a singular value. Thanks a lot for your comments 👍

mumoshu:
@davidovich I've added tests and some tweaks. Would you mind confirming? Thanks!

davidovich:
Will have a look tomorrow morning
Le mer. 19 sept. 2018 à 22:36, KUOKA Yusuke <notifications@github.com> a
écrit :

README.md
```diff
@@ -593,6 +593,110 @@ mysetting: |
...
+The following is an example hook that just prints the contextual information provided to hook:
+
+```
+releases:
+- name: myapp
+  chart: mychart
+  # *snip*
+  hooks:
+  - events: ["prepare", "cleanup"]
+    command: ["echo"]
```
davidovich: Upon your latest changes, I believe this is not an array anymore, what do you think?

mumoshu: Good catch! I'll fix it asap.

mumoshu:
@davidovich Thanks a lot for your review ☺️

PR #353, opened 2018-09-20
feat: `helmfile repos` updates all repos regardless of selector and releases

### LINKED ISSUE #338: when no charts are selected repos don't get synced
With the below helmfile a `helmfile repos` will result in 
`err: no releases found that matches specified selector() and environment(default), in any helmfile`

```
repositories:
  - name: stable
    url: https://kubernetes-charts.storage.googleapis.com
```

On a similar vein the following will result in the stable repo never being added as this helmfile itself has no matching charts in it.
```
repositories:
  - name: stable
    url: https://kubernetes-charts.storage.googleapis.com

environments:
  prod:
  dev:

helmfiles:
- helm/account/helmfile.yaml
{{ if (eq .Environment.Name "prod") }}
- helm/account/prod/helmfile.yaml
{{- end }}
{{ if (eq .Environment.Name "dev") }}
- helm/account/dev/helmfile.yaml
{{- end }}

```

mumoshu:
Resolves #338

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -4,13 +4,15 @@ import (
 	"fmt"
 	"github.com/roboll/helmfile/helmexec"
 	"github.com/roboll/helmfile/state"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"reflect"
 	"testing"
 )
 
 // See https://github.com/roboll/helmfile/issues/193
-func TestFindAndIterateOverDesiredStates(t *testing.T) {
+func TestVisitDesiredStatesWithReleasesFiltered(t *testing.T) {
 	absPaths := map[string]string{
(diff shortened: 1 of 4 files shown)


abdennour:
I think this is a bug ,, not fixing a bug. 
Why I need to wait all helm repos when  I am releasing a helm chart from a single repo `helmfile -l name=myrelease apply`

mumoshu:
@abdennour This is working as expected according to #338. What you're asking is a feature request.

mumoshu:
We opted to fetch all charts to fix #338 without implementing logic to correlate releases entries to repositories entries by chart names. Maybe it's easy that you can try to implement it?

PR #358, opened 2018-09-21
relax missing keys when evaluating final value files

### LINKED ISSUE #357: Go templates `default` functionality is broken after version 0.33.0
The following example (it's provided in the helmfile readme) works on version 0.33 but on version 0.34 and onwards:

`helmfile.yaml`
```
environments:
  production:
    values:
    - production.yaml

releases:
- name: myapp
  values:
  - values.yaml.gotmpl
```

`production.yaml`
```
domain: prod.example.com
releaseName: prod
values.yaml.gotmpl

domain: {{ .Environment.Values.domain | default "dev.example.com" }}
```

The above setup fails on version 0.34 and onwards with:
```
 map has no entry for key "domain"
```

when the `domain` value is missing from the environment file `production.yaml`

Probably the functionality of go templates is not the expected one when the key is missing from the values file.

davidovich:
This PR reintroduces `missingkey=zero` template rendering option for last render of values (before handing them to helm), to allow piping to default.

This PR can be rejected, depending on what is the outcome of #357 discussion.

Fixes #357

### DIFF
--- a/valuesfile/valuesfile.go
+++ b/valuesfile/valuesfile.go
@@ -19,6 +19,10 @@ func NewRenderer(readFile func(filename string) ([]byte, error), basePath string
 	}
 }
 
+func (r *renderer) SetMissingKeyZero(value bool) {
+	r.tmplFileRenderer.SetMissingKeyZero(value)
+}
+
 func (r *renderer) RenderToBytes(path string) ([]byte, error) {
 	var yamlBytes []byte
 	splits := strings.Split(path, ".")
--- a/main_test.go
+++ b/main_test.go
@@ -259,3 +259,57 @@ func TestReadFromYaml_RenderTemplateWithNamespace(t *testing.T) {
 		t.Errorf("release name should be namespace-myrelease")
 	}
 }
+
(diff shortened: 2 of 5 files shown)


sgandon:
It.would also be great to have some documentation on new features.
I am afraid there are too many things undocumented right now.

mumoshu:
@sgandon Hey! Thanks for the feedback.

What do you think should be documented in term of this PR?

sgandon:
it seems that there is a change in behaviour in the the rendering for the 'default' function so I guess people should be aware of it through an example.
I may be wrong there I am not really mastering the templating rendering phases right now.

valuesfile/valuesfile.go
```diff
@@ -19,6 +19,10 @@ func NewRenderer(readFile func(filename string) ([]byte, error), basePath string
 	}
 }
```
pppppppppp99:
```suggestion
ซ้ำซ้ำของ #
```suggestion
}
```
```suggestion
}
```

lastthyme (approved):

PR #370, opened 2018-09-26
feat: `get` and `getOrNil` template funcs to allow defaulting in templates

mumoshu:
Ref #357


mumoshu:
@sstarcher Thanks for the review! I've added tests and docs since then. This should be good to be merged now.

PR #372, opened 2018-09-27
Add instructions for installing via scoop on Windows

nightscape:


nightscape:
Hi @roboll, do you mind having a look at this?
I think having a scoop installer makes life much easier for Windows users :wink:

mumoshu:
@nightscape Hey! Thanks for sending this.

Two things I'm afraid before merging this is that I'm not aware of to whom I should ask for updating the scoop pkg when new versions of helmfile are released. Would you mind becoming the one?

nightscape:
The cool thing about scoop is that it automatically updates the package definitions by checking the Github release page:
https://github.com/lukesampson/scoop-extras/blob/master/helmfile.json#L22-L33
In case you should move from Github to Gitlab or use different file names for the executables I'll happily make the corresponding changes 😃

PR #459, opened 2019-01-30
support labels for namespace and chart

sstarcher:


osterman:
That's nice! we've been manually adding a namespace label, but this works too.

Issue #487, opened 2019-03-08
helm 2.13 and atomic support


sstarcher:
Helm rolled out 2.13 and a new flag called --atomic it would be nice to support it in helmfile.

Evesy:
I'm happy to try and tackle this one if no one minds?

sstarcher:
@Evesy please do it would be very appreciated.

Evesy:
I've only just thought, is this not covered under `--args`?

Unless it wants to be a first class option in helmfile like recreate pods etc.:

```
helmDefaults:
  tillerNamespace: tiller-namespace  #dedicated default key for tiller-namespace
  kubeContext: kube-context          #dedicated default key for kube-context (--kube-context)
  # additional and global args passed to helm
  args:
    - "--set k=v"
  # defaults for verify, wait, force, timeout and recreatePods under releases[]
  verify: true
  wait: true
  timeout: 600
  recreatePods: true
  force: true
  atomic: true
```

sstarcher:
We certainly want it to be first class.  The issue with args is it passes that to everything you call.

So if you do a helmfile diff or a helmfile template.  It will pass args which --atomic is not valid for either of those 2.  Overall  I think args should be removed from helmfile.

PR #491, opened 2019-03-14
Support --atomic as a first class directive in helmfile

Evesy:
- Adds support for `atomic` in `helmDefaults` or in Helmfile `releases` to enable `--atomic` flag on Helm upgrade/apply
- Bumps Helm to 2.13, a minimum required version for this change (I could not see any implementation of feature toggles based on Helm client versions, so this would just increase minimum Helm requirements)

I've kept the logic the same as existing options, however it doesn't make sense to me that setting an option explicitly to `false` in a release will not override if the equivalent in `helmDefault` is set to `true`. In my mind `helmDefault` should be the default value, but a provided option in the release should take precedence

Ref https://github.com/roboll/helmfile/issues/487

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ jobs:
       - run:
           name: Install helm
           environment:
-            HELM_VERSION: v2.12.0
+            HELM_VERSION: v2.13.0
           command: |
             HELM_FILENAME="helm-${HELM_VERSION}-linux-amd64.tar.gz"
             curl -Lo ${HELM_FILENAME} "https://kubernetes-helm.storage.googleapis.com/${HELM_FILENAME}"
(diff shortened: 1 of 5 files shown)


mumoshu:
> however it doesn't make sense to me that setting an option explicitly to false in a release will not override if the equivalent in helmDefault is set to true. 

Ah! Good point. I haven't used `helmDefaults` much myself so didn't notice that.

You are correct, we should make all these like the below. Let's create a dedicated issue for it.

```
if release.Atomic != nil && *release.Atomic || release.Atomic == nil && st.HelmDefaults.Atomic {
	flags = append(flags, "--atomic")
}
```

mumoshu:
Opened #492 for fixing defaulting

Evesy:
I can fix #492 too after this has been merged in

Issue #476, opened 2019-02-15
Integer environment variable gets converted to scientific notation even when quoted


daaain:
I have an API key for a service which happens to be a 12 digit integer (ie like 123456789012), which unfortunately gets converted to scientific notation even when quoted 🙀 

So with this Helmfile bit:

```yaml
set:
  - name: integer_var
    value: {{ requiredEnv "INTEGER_VAR" | quote }}
```

Using it like this in a Helm template:

```yaml
env:
- name: INTEGER_VAR
  value: {{ .Values.integer_var | quote }}
```

I get back:

```yaml
- name: INTEGER_VAR
  value: "1.23456789012+11"
```

Only this workaround fixes it:

```yaml
- name: INTEGER_VAR
  value: {{ int64 .Values.integer_var | quote }}
```

But it's annoying having to remember to explicitly convert it back to an integer.

Am I doing something wrong? The environment variable comes unquoted, but I can't change that.

Evesy:
I don't believe you're doing anything wrong. Helm introduced `--set-string` to tackle this exact issue (https://github.com/helm/helm/pull/3599).

I just tested a potential quick fix which is to support `--set-string` in Helmfile via something like the below (Specifying `string` instead of `value`):

```
set:
  - name: integer_var
    string: {{ requiredEnv "INTEGER_VAR" | quote }}
```

As with your original example the value is templated to:

```yaml
# data:
#   key: {{ .Values.integer_var | quote }}
data:
  key: "1.23456789012e+11"
```

Changing to use `string` in set it now templates to:

```yaml
# data:
#   key: {{ .Values.integer_var | quote }}
data:
  key: "123456789012"
```

Would this be an easier syntax to remember than your current workaround?

cc/ @sstarcher @mumoshu - Do you think something like this is worthwhile?

daaain:
Thanks a lot for looking into this @Evesy!

Yeah, that sounds like a much nicer way to deal with it, being guaranteed string can also prevent other potential duck typing surprises.

mumoshu:
@daaain @Evesy Hey! Sorry for the delayed response - but my thought was `values:` works in this case.

Instead of:

```yaml
set:
  - name: integer_var
    value: {{ requiredEnv "INTEGER_VAR" | quote }}
```

Write:

```
values:
- integer_var: {{ requiredEnv "INTEGER_VAR" | quote }}
```

This basically generates a temporary `values.yaml` to be consumed by helm, which should not suffer from the conversion issue.

[...]

PR #495, opened 2019-03-15
Add option to set to force helm to template value as string

Evesy:
Adds additional capabilities to `set` to force a value to be evaluated as a string (Translates to Helms `--set-string`)

Activated by specifying `string` instead of `value`:

```yaml
set:
- name: bigint
  string: 123456789012
```

Fixes https://github.com/roboll/helmfile/issues/476

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,9 @@ releases:
     # set a templated value
     - name: namespace
       value: {{ .Namespace }}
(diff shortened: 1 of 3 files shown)


mumoshu:
@Evesy Thanks for your contribution! Would you mind taking a look into my comment at #476?

Evesy:
Closing as alternative solution proposed in https://github.com/roboll/helmfile/issues/476

PR #517, opened 2019-03-29
Support multiple -f arguments

blissd:


blissd:
Hmm... actually, this is probably a big premature. Let me close it and do some better testing.

blissd:
Also, not sure how to this should be tested as part of the build process.

PR #522, opened 2019-03-31
feat: easier way to share configuration across helmfiles

### LINKED ISSUE #445: helmfiles include and charts in same repo
Is it intended to ignore `charts` if you have a helmfiles include

```
helmfiles:
- helmfiles_blah.yaml

charts:
- blah
```

With the above helmfiles_blah.yaml will run and anything under charts will not be processed.

mumoshu:
This basically allows to define both `releases` and `helmfiles` within a helmfile.yaml, so that you can start using sub-helmfiles easily, by extracting only reused releases.

Resolves #445

### DIFF
--- a/docs/shared-configuration-across-teams.md
+++ b/docs/shared-configuration-across-teams.md
@@ -1,0 +1,160 @@
+# Shared Configuration Across Teams
+
+Assume you have a two or more teams, each works for a different internal or external service, like:
+
+- Product 1
+- Product 2
+- Observability
+
+The simplest `helmfile.yaml` that declares the whole cluster that is composed of the three services would look like the below:
+
+```yaml
+releases:
+- name: product1-api
+  chart: product1-charts/api
+  # snip
+- name: product1-web
+  chart: product1-charts/web
+  # snip
+- name: product2-api
+  chart: saas-charts/api
+  # snip
+- name: product2-web
+  chart: product2-charts/web
+  # snip
+- name: observability-prometheus-operator
+  chart: stable/prometheus-operator
+  # snip
+- name: observability-process-exporter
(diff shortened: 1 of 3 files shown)


docs/shared-configuration-across-teams.md
```diff
@@ -0,0 +1,160 @@
...
+
+`product1/helmfile.yaml`:
+
+```yaml
+helmfiles:
+- ../observability/helmfile.yaml
+- ../product/helmfile.yaml
+```
+
+`product2/helmfile.yaml`:
```
sgandon: Why would you need a second file here since it is perfectly the same as the product1 ?

docs/shared-configuration-across-teams.md
```diff
@@ -0,0 +1,160 @@
...
+helmfiles:
+- ../observability/helmfile.yaml
+- ../product/helmfile.yaml
+```
+
+Now that we use the environment variable `PRODUCT_ID` to as the parameters of release names, you need to set it before running `helmfile`, so that it produces the differently named releases per product:
+
+```console
+$ PRODUCT_ID=1 helmfile -f product1/helmfile.yaml apply
+$ PRODUCT_ID=2 helmfile -f product2/helmfile.yaml apply
```
sgandon: same here, why not use product1/helmfile as they are the same and get rid off the product2

docs/shared-configuration-across-teams.md
mumoshu: Thanks for reviewing!

That was to get the first version of the doc done anyway 😄 

Please feel free to add the final section to this doc, something that removes `product1/helmfile.yaml` and `"product2/helmfile.yaml` in favor of `product/helmfile.yaml`.

Issue #449, opened 2019-01-23
feat: helm tiller integration


astorath:
# Proposal: helm tiller plugin integration

## Motivation
Provide a clean solution to run tillerless kubernetes using helmfile.
See #381 

### Problems
- Using helm tiller with helmfile requires separate setup of execution scripts. This separates helmfile setup and helmfile execution setup which is somewhat error-prone.

### Proposal
- Add an option to execute helm commands via helm-tiller addon to helmfile spec (`helmDefaults`), e.g.:
```
helm tiller run tiller-namespace -- helm diff ...
```
- Provide settings for: HELM_TILLER_SILENT, HELM_TILLER_PORT, HELM_TILLER_STORAGE, tiller_namespace, HELM_TILLER_HISTORY_MAX
- Provide an option to workaround `--kube-context` problem in helm-tiller addon (running helm with helm-tiller addon does not respect --kube-context option)
- Optionally provide settings for HELM_TILLER_LOGS_DIR
- Optionally support `helm tiller start` and `helm tiller start-ci`, but I don't think this is necessary

mumoshu:
Hey! Thanks for writing this. I believe this would be a great addition to helmfile.

helm-tiller seems like a practical way that works today to achieve Tilerless Helm without waiting until Helm v3.

Let me add a few notes on your proposal:

> Provide settings for: HELM_TILLER_SILENT, HELM_TILLER_PORT, HELM_TILLER_STORAGE, tiller_namespace, HELM_TILLER_HISTORY_MAX

Making `tiller_namespace` per per each `relelase` would be nice.

I don't have strong opinions on fine-tuning HELM_TILLER_SILENT, HELM_TILLER_PORT, HELM_TILLER_STORAGE, HELM_TILLER_HISTORY_MAX, so I'd just use envvars to globally set em.

HELM_TILLER_PORT can be automatically computed per a group of releases sharing the same `tillerNamespace` (and `HELM_TILLER_STORAGE` if we allow fine-tuning it per release).

I'd set `HELM_TILLER_STORAGE` to the secret backend by default when the helm-tiller integration is enabled, as long as there's no strong opinion to use configmaps.

> Provide an option to workaround --kube-context problem in helm-tiller addon (running helm with helm-tiller addon does not respect --kube-context option)

[...]

PR #531, opened 2019-04-02
Simple implementation of the tillerless mode

pvalsecc:
Partial implementation of #449

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,7 @@ clean:
 .PHONY: clean
 
 pristine: generate fmt
+	git diff | cat
 	git ls-files --exclude-standard --modified --deleted --others | diff /dev/null -
 .PHONY: pristine
 
--- a/helmexec/context.go
+++ b/helmexec/context.go
@@ -1,11 +1,17 @@
 package helmexec
 
+import (
+	"os"
+	"path/filepath"
+)
+
 type HelmContext struct {
 	Tillerless      bool
 	TillerNamespace string
+	Worker          int
 }
 
-func (context *HelmContext) GetPrefixArgs(helmBinary string) []string {
+func (context *HelmContext) GetTillerlessArgs(helmBinary string) []string {
 	if context.Tillerless {
 		if context.TillerNamespace != "" {
 			return []string{"tiller", "run", context.TillerNamespace, "--", helmBinary}
@@ -14,5 +20,25 @@ func (context *HelmContext) GetPrefixArgs(helmBinary string) []string {
 		}
 	} else {
 		return []string{}
+	}
+}
+
+func (context *HelmContext) getTillerlessEnv() map[string]string {
+	if context.Tillerless {
+		result := map[string]string{
+			"HELM_TILLER_SILENT": "true",
+			// Changing the TILLER port doesn't really work: https://github.com/helm/helm/issues/3159
+			// So this is not used for the moment.
+			// "HELM_TILLER_PORT":   strconv.Itoa(44134 + context.Worker),
+		}
+		if config := os.Getenv("KUBECONFIG"); config != "" {
+			absConfig, err := filepath.Abs(config)
+			if err == nil {
+				result["KUBECONFIG"] = absConfig
+			}
+		}
+		return result
+	} else {
+		return map[string]string{}
 	}
 }
--- a/helmexec/context.go
+++ b/helmexec/context.go
@@ -1,0 +1,18 @@
+package helmexec
+
+type HelmContext struct {
+	Tillerless      bool
+	TillerNamespace string
+}
+
+func (context *HelmContext) GetPrefixArgs(helmBinary string) []string {
+	if context.Tillerless {
+		if context.TillerNamespace != "" {
+			return []string{"tiller", "run", context.TillerNamespace, "--", helmBinary}
+		} else {
+			return []string{"tiller", "run", "--", helmBinary}
+		}
+	} else {
+		return []string{}
+	}
+}
--- a/helmexec/context.go
+++ b/helmexec/context.go
@@ -8,7 +8,7 @@ import (
 type HelmContext struct {
 	Tillerless      bool
 	TillerNamespace string
-	Worker          int
+	WorkerIndex     int
 }
 
 func (context *HelmContext) GetTillerlessArgs(helmBinary string) []string {
@@ -29,7 +29,7 @@ func (context *HelmContext) getTillerlessEnv() map[string]string {
 			"HELM_TILLER_SILENT": "true",
 			// Changing the TILLER port doesn't really work: https://github.com/helm/helm/issues/3159
 			// So this is not used for the moment.
-			// "HELM_TILLER_PORT":   strconv.Itoa(44134 + context.Worker),
+			// "HELM_TILLER_PORT":   strconv.Itoa(44134 + context.WorkerIndex),
 		}
 		if config := os.Getenv("KUBECONFIG"); config != "" {
 			absConfig, err := filepath.Abs(config)
--- a/helmexec/exec.go
+++ b/helmexec/exec.go
@@ -119,13 +119,14 @@ func (helm *execer) List(context HelmContext, filter string, flags ...string) (s
 	return string(out), err
 }
 
-func (helm *execer) DecryptSecret(name string) (string, error) {
+func (helm *execer) DecryptSecret(context HelmContext, name string, flags ...string) (string, error) {
 	// Prevents https://github.com/roboll/helmfile/issues/258
 	helm.decryptionMutex.Lock()
 	defer helm.decryptionMutex.Unlock()
 
 	helm.logger.Infof("Decrypting secret %v", name)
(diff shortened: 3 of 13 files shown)


helmexec/runner.go
```diff
@@ -23,5 +24,7 @@ type ShellRunner struct {
 func (shell ShellRunner) Execute(cmd string, args []string) ([]byte, error) {
 	preparedCmd := exec.Command(cmd, args...)
 	preparedCmd.Dir = shell.Dir
+	preparedCmd.Env = os.Environ()
+	preparedCmd.Env = append(preparedCmd.Env, "HELM_TILLER_SILENT=true")
 	return preparedCmd.CombinedOutput()
```
mumoshu: Very nit but you should also alter `KUBECONFIG` to an absolute path if it is already set, otherwise `helm-tiller` is unable to connect the cluster.

```
# This doesn't work AFAIK
KUBECONFIG=kubeconfig helm tiller run -- helm list

# This works
KUBECONFIG=$(pwd)/kubeconfig helm tiller run -- helm list
```

Perhaps something like this is needed:

```
kubeconfig := os.Getenv("KUBECONFIG")
if kubeconfig != "" {
    abs, err := filepath.Abs(kubeconfig)
    // err handling...
    if !abs {
        wd, err := os.Getwd()
        if err != nil {
            return ....
        }
        preparedCmd.Env = append(preparedCmd.Env, "KUBECONFIG=" + filepath.Join(wd, kubeconfig))
    }
    // ....
}
```

helmexec/exec.go
```diff
@@ -110,20 +111,22 @@ func (helm *execer) ReleaseStatus(name string) error {
...
 
-func (helm *execer) DecryptSecret(name string) (string, error) {
+func (helm *execer) DecryptSecret(context HelmContext, name string, flags ...string) (string, error) {
 	// Prevents https://github.com/roboll/helmfile/issues/258
 	helm.decryptionMutex.Lock()
 	defer helm.decryptionMutex.Unlock()
 
 	helm.logger.Infof("Decrypting secret %v", name)
-	out, err := helm.exec(append([]string{"secrets", "dec", name})...)
+	preArgs := context.GetPrefixArgs(helm.helmBinary)
```
mumoshu: I think `GetPrefixArgs` can be more explicit for now, like `GetTillerlessArgs`.
We can generalize it to `GetPrefixArgs` once when we add other prefixes.

helmexec/context.go
```diff
@@ -0,0 +1,18 @@
+package helmexec
+
+type HelmContext struct {
```
mumoshu: I wonder if we can safely merge this into `execer`. I have been long thinking that `execer` is the "context" and adding an another context "HelmContext" isn't a must? 🤔

pvalsecc: No can do. `execer`s lifecycle is not tied to the release, but is allocated once per helmfile. And it is used in // per multiple threads. That is why I ended up passing a context to most methods.

helmexec/runner.go
pvalsecc: OK, I ended up with a very different code, but it's done.

pvalsecc:
@mumoshu, I just have a pain point, but I don't know if I can fix it. The tillerless mode works only with `--concurrency=1` if you have more that one release per `tillerNamespace`. I've documented this limitation in the readme, but, I don't know if you guys can live with it...

mumoshu:
@pvalsecc Why is it tied to concurrency=1? Tiller ports' collision?

pvalsecc:
> Why is it tied to concurrency=1? Tiller ports' collision?

Yes, I guess... it's not like there was any error message.

helmexec/context.go
mumoshu: Ah that's good point. I dreamed of changing it to use a dedicated execer per release, but turned out it affects a lot of code including tests. Let's use HelmContext for now.

Thanks!

mumoshu:
@pvalsecc Perhaps setting a dedicated `HELM_TILLER_PORT` value per release works?  accordinvg to `helm tiller --help` it starts from 44134 so... do we need to create one HelmContext per concurrency = goroutine?

pvalsecc:
Yes, I just came to the same conclusion. I'll try tomorrow.

mumoshu:
Um.. I'm not sure if multiple tillers can safely operate on the same tiller namespace. Does tiller have a kind of exclusive lock on tiller namespace?

mumoshu:
Maybe helmfile should cap concurrency per tiller-namespace to 1. I believe it's fine to run concurrent helm-tillers across different tiller namespaces, though.

pvalsecc:
Tiller runs on your local machine, in this mode. All it uses are secrets and you have one set of secrets per release, named after the release. So it should be quite safe to deal with the different releases in parallel.

mumoshu:
> All it uses are secrets and you have one set of secrets per release, named after the release

Makes sense. Thanks for your support ☺️

pvalsecc:
hummm, looks like HELM_TILLER_PORT doesn't work and nobody really cares:
https://github.com/rimusz/helm-tiller/issues/30
https://github.com/helm/helm/issues/3159

pvalsecc:
@mumoshu, IMHO, this PR is ready for your review.

Makefile
```diff
@@ -42,6 +42,7 @@ clean:
 .PHONY: clean
 
 pristine: generate fmt
+	git diff | cat
```
mumoshu: What do we need this for?(Just curious

pvalsecc: At work, I have an old version of go that doesn't agree with the autoformatting of the one in circleci. So I needed the actual output of the diff instead of just the list of files.

helmexec/exec_test.go
```diff
@@ -258,7 +286,7 @@ func Test_TestRelease_Flags(t *testing.T) {
 	var buffer bytes.Buffer
 	logger := NewLogger(&buffer, "debug")
 	helm := MockExecer(logger, "dev")
-	helm.TestRelease("release", "--cleanup", "--timeout", "60")
+	helm.TestRelease(HelmContext{Tillerless: false}, "release", "--cleanup", "--timeout", "60")
```
mumoshu: Nit but can we safely omit Tillerless:false as its the default(zero) value?

helmexec/exec.go
```diff
@@ -70,60 +70,68 @@ func (helm *execer) AddRepo(name, repository, certfile, keyfile, username, passw
...
+func (helm *execer) DecryptSecret(context HelmContext, name string, flags ...string) (string, error) {
 	// Prevents https://github.com/roboll/helmfile/issues/258
 	helm.decryptionMutex.Lock()
 	defer helm.decryptionMutex.Unlock()
 
 	helm.logger.Infof("Decrypting secret %v", name)
-	out, err := helm.exec(append([]string{"secrets", "dec", name})...)
+	preArgs := context.GetTillerlessArgs(helm.helmBinary)
+	env := context.getTillerlessEnv()
+	out, err := helm.exec(append(append(preArgs, "secrets", "dec", name), flags...), env)
```
mumoshu: I thought so but please let me confirm - You run helm-tiller because helm-plugin is tied to tiller, even when the actual plugin code(helm-secrets in this case, which doesnt access tiller itself) isnt?

helmexec/context.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"os"
+	"path/filepath"
+)
+
+type HelmContext struct {
+	Tillerless      bool
+	TillerNamespace string
+	Worker          int
```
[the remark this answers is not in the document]
pvalsecc: OK, renaming

helmexec/exec_test.go
pvalsecc: What do you mean by "nit"?
OK, thanks. I never did write any go until previous PR... so I'm learning...  ;-)

helmexec/exec.go
pvalsecc: Exactly. Looks like helm, before calling any plugin, tries to talk to its server. Sadly inefficient.

helmexec/exec_test.go
mumoshu: Thanks for taking your time :)

"Nit" stands for "nitpick" so I mean "I'm hesitant to point out such a small thing even though your PR is awesome overall, but…" :relaxed:

Makefile
mumoshu: Got it. Thanks for clarifying 👍

pvalsecc:
Thanks for the quick release, @mumoshu !

mumoshu:
@pvalsecc Thanks for the great work!

Also, I've submitted a tiny PR to helm that makes the probe port configurable, which should allow us to support concurrency > 1

I'd appreciate it if you could take a look 😃

mumoshu:
The PR has been merged 🎉

pvalsecc:
Thanks for the merge and release.

> I'd appreciate it if you could take a look

It's already merged :-)

But then there would be more work on the helm-tiller plugin to use that feature. So many layers...

PR #560, opened 2019-04-18
feat(#502): display summary of upgraded, deleted and error releases

### LINKED ISSUE #502: chart version not displayed when doing sync
When running `helmfile sync`  we don't know which chart version was used especially if some semver constraint was used .
That would be useful to display the version installed.

sgandon:
Fixes #502 
This fix display the actual version of the deployed release.
But I have added more to it. At the end of the helmfile execution there is a complete summary of all the releases that where upgraded, deleted and in errors.
this show like this
```
List of updated releases :
RELEASE          CHART                  VERSION
tdc-caralog      ../../tdc-caralog        0.1.3
tdc-gateway      ../../tdc-gateway        0.1.2
common-gagaset   ../../common-gagaset    0.1.15
common-rating    ../../common-rating     0.1.10
common-sharing   ../../common-sharing    0.1.13
tpsvc-pirtal     ../../tpsvc-pirtal       0.1.3

List of deleted releases :
RELEASE
tpsvc-concafig
tpsvc-crouto

List of releases in error :
RELEASE
tpsvc-foox
```

I am brand new at Golang so please do not hesitate on giving me all the advices I could use to improve the code.
I have added a new lib for displaying nicely the updated releases table, I hope it is fine.


sgandon:
The build failed cause the makefile does not update the dependencies before build the project. 
I'll look at that tomorrow.

sgandon:
I have removed the `vendor` folder as it is a bad practice to store the depenencies in the CSV system.
the pkg.toml is here for that purpose.

Makefile
```diff
@@ -1,8 +1,12 @@
 ORG     ?= $(shell basename $(realpath ..))
 PKGS    := $(shell go list ./... | grep -v /vendor/)
 
-build:
-	go build ${TARGETS}
+dep: ## Get build dependencies
+	go get -v -u github.com/golang/dep/cmd/dep
```
mumoshu: Great!
Excuse me but I would tweak this a bit when I enable build cache in CircleCI so that you don't need to re-download the whole depencies on each build.

Perhaps I'll make build depends on `dep`, dep runs `go mod download`.

main.go
```diff
@@ -288,20 +291,23 @@ func main() {
...
-				return findAndIterateOverDesiredStatesUsingFlags(c, func(state *state.HelmState, helm helmexec.Interface, ctx app.Context) []error {
+				affectedReleases := state.AffectedReleases{}
+				errs := findAndIterateOverDesiredStatesUsingFlags(c, func(st *state.HelmState, helm helmexec.Interface, ctx app.Context) []error {
 					if !c.Bool("skip-deps") {
-						if errs := ctx.SyncReposOnce(state, helm); errs != nil && len(errs) > 0 {
+						if errs := ctx.SyncReposOnce(st, helm); errs != nil && len(errs) > 0 {
 							return errs
 						}
-						if errs := state.BuildDeps(helm); errs != nil && len(errs) > 0 {
+						if errs := st.BuildDeps(helm); errs != nil && len(errs) > 0 {
```
mumoshu: Thanks for making `st`s consistent 👍

main.go
```diff
@@ -507,10 +520,12 @@ Do you really want to delete?
 `, strings.Join(names, "\n"))
 					interactive := c.GlobalBool("interactive")
 					if !interactive || interactive && askForConfirmation(msg) {
-						return state.DeleteReleases(helm, true)
+						return state.DeleteReleases(&affectedReleases, helm, true)
 					}
 					return nil
 				})
+				affectedReleases.DisplayAffectedReleases(c.App.Metadata["logger"].(*zap.SugaredLogger))
```
mumoshu: Just a note: I think I'll be changing all the verbose log messages from helmfile to be sent via the zap logger connected to STDERR,
and useful messages like ones printed by `DisplayAffectedReleases` to be redirected to STDOUT.

state/state.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/roboll/helmfile/environment"
 	"github.com/roboll/helmfile/event"
 	"github.com/roboll/helmfile/tmpl"
+	"github.com/tatsushid/go-prettytable"
```
mumoshu: There's a small chance I'd change this to use another table printer, but I do believe printing tables for better summarization is a great idea! Thanks about that.

state/state.go
```diff
@@ -150,6 +153,13 @@ type SetValue struct {
 	Values []string `yaml:"values"`
 }
 
+// AffectedReleases hold the list of released that where updated, deleted, or in error
+type AffectedReleases struct {
+	Upgraded []*ReleaseSpec
+	Deleted  []*ReleaseSpec
+	Error    []*ReleaseSpec
```
mumoshu: Better use Errored, or Failed for consistency?

mumoshu:
@sgandon Actually, commiting `vendor` has been a common sense in the golang community for long time. That was to avoid build-time dependency to external, possibly unreliable repositories, and golang pkg managers had no proper shared dependency cache.

The world has changed and we now have "go modules" or "vgo", a.k.a the `go mod` command. It has a relatively proper(i think) shared dep cache. So I was about to migrate helmfile to use go modules.

Removing `vendor/` before switching to go modules isn't good, but assuming we'd switch soon, I'm happy with merging this as-is :)

Thanks again for your effort!

state/state.go
sgandon: Yes, I think that is an old library that was never released so I guess we can use another one.
this one `github.com/olekukonko/tablewriter v0.0.1` is pretty similar and has a fixed release.

state/state.go
sgandon: yes good idea, I'll try to make another PR with your comments

Issue #548, opened 2019-04-09
Environment values files should allow missing files


sgandon:
The use case is the following.
We are storing chart version in a values file which is stored allong with the helmfile but could be automatically overridden by a local values file created from versions found in a given production cluster for instance.
But this local file may be missing and should not prevent the helmfile to run.

Just like the values files for _releases_ the ones from the `environment` element should allow missing files.
I would suggest to use the same mechanism for explicitly allow that using `missingFileHandler`

So something like that would be great.
```
environments:
  missingFileHandler: warn
  default:
    values:
      - ../values/chart-versions.yaml
```

lruslan:
I have similar issue only with release values.  It would be super useful just show warning when values file is missing (or at least have this behaviour configurable)
In our case we want to be able to allow customisation of the configuration for some GKE clusters. 
In majority of cases using common settings is fine but few clusters need customisation. 
Our users need to be able to provision clusters when they want and currently to make this helmfile workflow work we need them also publish empty values file even if they intended to use default common values, which is not really convenient. 

Our helmfile:
```
 - name: nginx-ingress-open
    chart: stable/nginx-ingress
    namespace: ingress-nginx-open
    version: "1.4.0"
    values:
    - ../values/prometheus/common.yaml
    - ../values/promehteus/{{ requiredEnv "ENVIRONMENT" }}/{{ requiredEnv "GKE_CLUSTER_IDENTIFIER" }}.yaml

  - name: prometheus
    chart: stable/prometheus
    namespace: monitoring
    version: "8.9.2"
    values:
    - ../values/prometheus/common.yaml
    - ../values/promehteus/{{ requiredEnv "ENVIRONMENT" }}/{{ requiredEnv "GKE_CLUSTER_IDENTIFIER" }}.yaml
```

sgandon:
there is a solution for those releases values, use `missingFileHandler: warn`
like this
```
- name: nginx-ingress-open
    chart: stable/nginx-ingress
    namespace: ingress-nginx-open
    version: "1.4.0"
    missingFileHandler: warn # prevent the install to stop if some files are not found
    values:
    - ../values/prometheus/common.yaml
    - ../values/promehteus/{{ requiredEnv "ENVIRONMENT" }}/{{ requiredEnv "GKE_CLUSTER_IDENTIFIER" }}.yaml

  - name: prometheus
    chart: stable/prometheus
    namespace: monitoring
    version: "8.9.2"
[...]

PR #561, opened 2019-04-19
readme example for failing when values file is missing

aegershman:
Believe this would assist with #548

Don't mean to steal anyone's thunder in updating the doc, if anyone wants to take charge on this and replace it please do. All credit to @sgandon


aegershman:
Agreed, updated to include the comment yanked from [the source code](https://github.com/roboll/helmfile/search?q=missingFileHandler&unscoped_q=missingFileHandler)

PR #564, opened 2019-04-25
Bump go to 1.12.4 / Switch to go modules

mumoshu:
Follow-up for https://github.com/roboll/helmfile/pull/560#issuecomment-486516109


Makefile
```diff
@@ -1,12 +1,8 @@
...
 PKGS    := $(shell go list ./... | grep -v /vendor/)
 
-dep: ## Get build dependencies
-	go get -v -u github.com/golang/dep/cmd/dep
-.PHONY: dep		
-
-build: ## Build the app after getting all dependencies (make sure to call make dep before)
-	dep ensure && go build ${TARGETS}
+build:
+	go build ${TARGETS}
```
mumoshu: Note that after vgo/go modules, `go build` automatically fetch all the dependencies, as long as you're outside of $GOPATH. No need to run a command like `dep ensure`.

So If you've been checking out helmfile's tree into e.g. `~/go/src/github.com/roboll/helmfile`, move the helmfile dir out of `~/go`, like `~/helmfile`.

mumoshu: cc/ @sgandon FYI

sgandon: cool!, that was just all so wierd to have to clone the repo into a specific folder.

.circleci/config.yml
```diff
@@ -4,24 +4,37 @@ jobs:
 
   build:
     docker:
-      - image: circleci/golang:1.12.1
-    working_directory: /go/src/github.com/roboll/helmfile
+      - image: circleci/golang:1.12.4
+    environment:
+      GOPATH: /go
```
davidovich: @mumoshu you might want to add GO111MODULE=on here. (https://circleci.com/blog/go-v1.11-modules-and-circleci/)

With go modules, we are not bound to GOPATH anymore (yay!).

Makefile
```diff
@@ -1,12 +1,8 @@
 ORG     ?= $(shell basename $(realpath ..))
 PKGS    := $(shell go list ./... | grep -v /vendor/)
```
davidovich: If you add 
```
export GO111MODULE := on
```
This will enable go modules independently of it's checked-out path. See the [relevant entry in the go modules wiki](https://github.com/golang/go/wiki/Modules#when-do-i-get-old-behavior-vs-new-module-based-behavior).

.circleci/config.yml
mumoshu: Thanks for the info!

I was thinking that we should rely on the `GO111MODULE=auto` which is the default, so I tweaked `working_directory` rather than `GO111MODULE` like this.

Also `GOPATH=/go` is the default value AFAIK and here to tell readers that the GOPATH is there, so that they can see why `save_cache` is pointing to the paths under `/go`.

Please correct me or add anything if you feel like to do so. Thanks a lot for your review ☺️

Makefile
mumoshu: Ah, that sounds good, so that I don't need the instruction like https://github.com/roboll/helmfile/pull/564#discussion_r278399438 for everyone?

PR #566, opened 2019-04-25
fix(struct): renamed Error struct field into Failed

sgandon:
as @mumoshu suggested I have renamed the AffectedReleases.Error field to AffectedReleases.Failed to be more homogeneous with the other fields and to avoir conflits with other Error references.


mumoshu:
@sgandon LGTM. Thanks a lot for fixing it 👍

PR #567, opened 2019-04-27
feat(#344): add sub helmfiles explicit selectors

### LINKED ISSUE #344: Possible to set selector in parent helmfile?
I have a helmfile that pulls in another helmfile
_cluster1-helmfile.yaml_
```
---
context: cluster1
helmfiles:
 - apps-helmfile.yaml
```
Inside the second helmfile there are selectors
_apps-helmfile.yaml_
```
releases:
- name: app1
  chart: repo/app1
  version: '*'
  labels:
    enabled: app1
- name: app2
  chart: repo/app2
  version: '*'
  labels:
    enabled: app2
```
Currently I deploy it using
```
helmfile --selector enabled=app1 -f cluster-helmfile.yaml sync
```
Instead is it possible to set the enabled label in the parent helmfile? Similar to https://github.com/roboll/helmfile/issues/168#issuecomment-419620834

sgandon:
Fixes #344 by allowing explicit selectors to be specified for composed helmfiles using the following structure
```
helmfiles:
- helmfile.d/a*.yaml:
    selectors:
    - name=prometheus      
    - name!=zipkin      
- helmfile.d/b*.yaml
- helmfile.d/c*.yaml:
    selectors: {}
```
2 modes here : 
* legacy mode when no the env var EXPERIMENTAL is not set to true
  * no selector : inherit from the command line.
  * selector:  is specified then it is used (an emty means no inheritance from command line and take everything).
* experimental when the env var EXPERIMENTAL=true
  * no selector : nothing is inherited from the command line so use all releases.
  * selector:  is specified then it is used (an emty means no inheritance from command line and take everything).

### DIFF
--- a/README.md
+++ b/README.md
@@ -660,7 +660,7 @@ helmfiles:
 * When a selector is specified, only this selector applies and the parents or CLI selectors are ignored.
 * When not selector is specified there are 2 modes for the selector inheritance because we would like to change the current inheritance behavior (see [issue #344](https://github.com/roboll/helmfile/issues/344)  ).
   * Legacy mode, sub-helmfiles without selectors inherit selectors from their parent helmfile. The initial helmfiles inherit from the command line selectors. 
-  * explicit mode, sub-helmfile without selectors do not inherit from their parent or the CLI selector. If you want them to inherit from their parent selector then use `selectors: inherits`. To enable this explicit mode you need to set the following environment variable `EXPERIMENTAL=explicit-selector-inheritance`.
+  * explicit mode, sub-helmfile without selectors do not inherit from their parent or the CLI selector. If you want them to inherit from their parent selector then use `selectors: inherits`. To enable this explicit mode you need to set the following environment variable `HELMFILE_EXPERIMENTAL=explicit-selector-inheritance` (see [experimental](#experimental-features)).
 * Using `selector: {}` will for all releases to be used regardless of the parent selector or cli for the initial helmfile
 * using `selector: inherits` make the sub-helmfile selects release with the parent selector or the cli for the initial helmfile
 
@@ -828,10 +828,10 @@ See #155 for more information on this topic.
 
 ## Experimental features
 Some experimental features may be available for testing in perspective of being (or not) included in a future release.
-Those features are set using the environment variable `EXPERIMENTAL`. Here is the current experimental feature :
+Those features are set using the environment variable `HELMFILE_EXPERIMENTAL`. Here is the current experimental feature :
 * `explicit-selector-inheritance` : remove today implicit cli selectors inheritance for composed helmfiles, see [composition selector](#selectors)
 
-If you want to enable all experimental features set the env var to `EXPERIMENTAL=true`
+If you want to enable all experimental features set the env var to `HELMFILE_EXPERIMENTAL=true`
 
 ## Examples
 
--- a/README.md
+++ b/README.md
@@ -332,7 +332,7 @@ The `selector` parameter can be specified multiple times. Each parameter is reso
 
 `--selector tier=frontend --selector tier=backend` will select all the charts
 
-In addition to user supplied labels the name, namespace, and chart are available to be used as selectors.  The chart will just be the chart name excluding the repository (Example `stable/filebeat` would be selected using `--selector chart=filebeat`).
+In addition to user supplied labels, the name, the namespace, and the chart are available to be used as selectors.  The chart will just be the chart name excluding the repository (Example `stable/filebeat` would be selected using `--selector chart=filebeat`).
 
 ## Templates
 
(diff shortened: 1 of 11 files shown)


sgandon:
I forgot to update the readme, I'll do so soon

davidovich:
This seems good. I also like the EXPERIMENTAL flag, but having only one Boolean value feels like it has only one purpose. What about using an identifier like EXPERIMENTAL=selectors ? That way the env var can be used to introduce simultaneous experimental features, controlled by one flag (probably allowing coma separated identifiers).

sgandon:
Year that's a good idea. 
I'll change it to a feature enablement flag.

sgandon:
in fact I like the idea of enabling all experimental feature at once, so I guess I'll provide both identifiers and a "true" value to enable them all.

sgandon:
all done, waiting for review.

README.md
```diff
@@ -805,6 +826,12 @@ Once you download all required charts into your machine, you can run `helmfile c
 It basically run only `helm upgrade --install` with your already-downloaded charts, hence no Internet connection is required.
 See #155 for more information on this topic.
 
+## Experimental features
+Some experimental features may be available for testing in perspective of being (or not) included in a future release.
+Those features are set using the environment variable `EXPERIMENTAL`. Here is the current experimental feature :
```
davidovich: Nice. As I read this though, I think that EXPERIMENTAL alone cries for name clashes. Can we prefix with HELMFILE_ to remove this potential clash?

sgandon: absolutly !

sgandon:
@mumoshu but if I do thid (which would have been much easier for pasring) we will loose the backward compatibily that I did maintain with the current PR, or am I missing something ?

sgandon:
or do think this is not an issue to maintain backward compatiblity ?

mumoshu:
@sgandon I do want to keep backward-compatibility, and it looked possible as your implementation is handling it in https://github.com/roboll/helmfile/pull/567/files#diff-f6be34387a141f652301c16cf7a5e36cR1411

That being said, the type switch there wouldn't change.

In my example, the first and the third sub-helmfile should be handled with the `map` case, and the second case with the `string` case(backward compatibility):

```
helmfiles:
- path: helmfile.d/a*.yaml
  selectors:
  - name=prometheus      
  - name!=zipkin      
- helmfile.d/b*.yaml
- path: helmfile.d/c*.yaml
  selectors: {}
```

sgandon:
ha sorry I did not see this example, so yes we can have both :)
I'll do the change and fix the conflictings files.

sgandon:
@mumoshu could you help me out with the test failure on circle-ci please ? the command `make test` works fine on my machine.

sgandon:
I believe I found the issue, map range does not guaranty the order. I'll fix that.

state/state.go
```diff
@@ -1387,3 +1398,101 @@ func escape(value string) string {
...
+		return fmt.Errorf("Expecting a \"selectors\" mapping for string [-%v] but got: %v of type %T", hf.Path, value, value)
+	}
+	return nil
+}
+
+func extractSelectorContent(hf *SubHelmfileSpec, value interface{}) error {
+	switch selectors := value.(type) {
+	case string: //check the string is `inherits` else error
+		if selectors == InheritsYamlValue {
+			hf.Inherits = true
```
mumoshu: Can be a nit but I'd suggest using:

```
helmfiles:
- path: path/to/subhelmfile.yaml
  selectorsInherited: true
```

So that the implementation becomes a bit simpler without the overloading of `selectors`.

state/state.go
```diff
@@ -54,6 +56,15 @@ type HelmState struct {
...
 }
 
+// SubHelmfileSpec defines the subhelmfile path and options
+type SubHelmfileSpec struct {
+	Path      string   //path or glob pattern for the sub helmfiles
+	Selectors []string //chosen selectors for the sub helmfiles
+	Inherits  bool     //do the sub helmfiles inherits from parent selectors
+}
+
+const InheritsYamlValue = "inherits"
```
mumoshu: Nit: Can be `SelectorsInherits` so that it looks a bit more consistent with golang community's convention that uses `Where + What` for const names

mumoshu: This can be removed completely if we go this way https://github.com/roboll/helmfile/pull/567/files#r281003831

state/state.go
```diff
@@ -1387,3 +1398,101 @@ func escape(value string) string {
...
+			switch key := k.(type) {
+			case string:
+				//get the path
+				if key == "path" {
+					hf.Path = v.(string)
+				} else if key == "selectors" {
+					if err := extractSelectorContent(hf, v); err != nil {
+						return err
+					}
+				} else {
```
mumoshu: Can you by any chance remove this implementation of "dict key as the path to a subhelmfile" in favor of `path`, so that the implementation becomes simpler, and there's fewer ways to achieve the same goal?

state/state.go
sgandon: yes, I'll do that

state/state.go
sgandon: yes, I'll do

Issue #392, opened 2018-11-02
Feat: Allow simple Vault integration


AndresPineros:
Currently there are many users that want to integrate Vault with Kubernetes, but there are no high level tools for this. The current Kubernetes AuthMethod for Vault is too complex and coupled to the definition of the charts. This proposal could allow for a Vault integration with Kubernetes that is simple and decoupled.

This could be the structure of a helmfile integrated with Vault:
```
releases:
  - chart: .
    name: release-test
    version: 0.1.0
    values:
      - application:
           mysql_url: jdbc.....
           mysql_user: localtest
           mysql_password: localtest
    vault:
      active: true
      secrets:
        - vaultPath: secret-engine/my-application
          key: mysql.username
          targetVar: application.mysql_user
        - vaultPath: secret-engine/my-application
          key: mysql.password
          targetVar: application.mysql_password
```

Basically the project could use the VAULT_ADDR and VAULT_TOKEN environment variables to query all the secrets from Vault using the `vaultPath` and the `key` to inject them to the helm release using `--set`.

So, for that example the `--set` flags of the `helm upgrade -i ...` command would look like:
```
--set application.mysql_user=<queried-secret> --set application.mysql_password=<queried-secret>
```

If the `vault.active` field is set to false, or the `vault` field is not used, the release will just use the "localtest" values.

The main advantage I see for this proposal is that it allows centralization of secrets. Usually storing encrypted secrets in repositories result in secret-management-madness as there is no unique source of truth in case multiple repositories require the same secret. Vault solves this. Of course, this approach only addresses static Vault secrets (secrets in git repos are static too). For dynamic Vault secrets, I think there's no other alternative than the Kubernetes Auth Method.

There is one security issue with this proposal: Tiller has to use the --storage (Secret) backend instead of the ConfigMap backend to avoid fetching injected secrets with `helm get values <release-name>`. Although, I think this is also a problem for the `helm-secrets` pluign.

AndresPineros:
I would really like to PR this, but I don't know Golang. I just started learning it to be able to contribute, but I don't know how long it can take for me to learn and do a decent PR.

mumoshu:
@AndresPineros Hey!

> there is no unique source of truth in case multiple repositories require the same secret

Yeah, I believe this is definitely the thing we should consider.

[...]

Issue #388, opened 2018-10-25
Using helmfiles, but unable to pass Environment down to helmfile.yaml


nikolajbrinch:
When using
```
{{ readFile "environments.yaml }}}
--- 
helmfiles:
- ./*/helmfile.yaml
```

Environment.Values are not passed to the globbed helmfiles. I would expect that to happen/work, in order to keep helmfiles DRY.

Tarick:
The same issue, using helmfile version v0.40.1

```
{{ readFile "common.yaml" }}
---
{{ readFile "environments.yaml" }}
```
With debug:
> Processing helmfile.yaml
could not deduce `environment:` block, configuring only .Environment.Name. error: failed to read helmfile.yaml: environment "dev" is not defined
error in first-pass rendering: result of "helmfile.yaml":

Once I put environments block in helmfile.yaml, readFile of environments.yaml succedes as well and I have 2 environments blocks in the output.

mumoshu:
@nikolajbrinch Hey! Your case isn't supposed to work as of today. That's because each helmfile is meant to be self-contained. Injecting environments like prod/staging/testing from the "parent" helmfile into "child" helmfiles is considered harmful because it works like an implicitly dependency from children to the parent.

I'll shortly reconsider though, about how we could support "inline environment values" and "injecting environment values from the command-line". Once we support either or both of them, I think it will justify passing environment values from the parent to the child.

mumoshu:
@Tarick Hey! Just to clarify, are you talking about the fact that helmfile doesn't propagate the environment and the environment values from the parent to children? Or you're maybe talking about a possible bug(?) that multiple yaml docs declared within a single helmfile.yaml doesn't get merged?

Tarick:
> @Tarick Hey! Just to clarify, are you talking about the fact that helmfile doesn't propagate the environment and the environment values from the parent to children? Or you're maybe talking about a possible bug(?) that multiple yaml docs declared within a single helmfile.yaml doesn't get merged?

The latter case. Basically, I tried to do what https://github.com/roboll/helmfile/blob/master/docs/writing-helmfile.md#layering recommends and this fails.

nikolajbrinch:
I think one way could be conventions. i.e. environment.yaml declares environment. Just like values.yaml in Helm Charts declares values.
I really like this project and uses it for our infrastructure, but I have a couple of questions:
[...]

PR #569, opened 2019-04-29
Added SSM integration

rms1000watt:
There's been some interest for `helmfile` integration with SSM. Here is an example of what it can look like.

For our current workflows, we have Bash scripts that export Env Vars via `aws-env` then run `helmfile`. It would be way cleaner to have this integrated

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,16 +2,18 @@ package main
 
 import (
 	"fmt"
+	"os"
+	"strings"
+
 	"github.com/roboll/helmfile/args"
 	"github.com/roboll/helmfile/cmd"
 	"github.com/roboll/helmfile/helmexec"
 	"github.com/roboll/helmfile/pkg/app"
+	"github.com/roboll/helmfile/ssm"
 	"github.com/roboll/helmfile/state"
 	"github.com/urfave/cli"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
-	"os"
-	"strings"
 )
 
 var Version string
@@ -569,6 +571,7 @@ Do you really want to delete?
 		},
 	}
 
+	ssm.Run()
 	err := cliApp.Run(os.Args)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "%v\n", err)
--- a/ssm/ssm.go
+++ b/ssm/ssm.go
@@ -1,0 +1,117 @@
+package ssm
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ssm"
+	"github.com/roboll/helmfile/helmexec"
+	"go.uber.org/zap"
+)
+
+var (
+	svcSSM  *ssm.SSM
+	ssmPath string
+	logger  *zap.SugaredLogger
+)
+
+func Run() {
+	logger = helmexec.NewLogger(os.Stdout, "debug")
+	ssmPath, exists := os.LookupEnv("SSM_PATH")
+	if !exists {
+		return
+	}
+
+	if strings.TrimSpace(ssmPath) == "" {
+		ssmPath = "/"
+	}
+
+	logger.Debugf("Attempting to populate environment with SSM values at path: %s", ssmPath)
+	if err := configureAWS(); err != nil {
+		logger.Error("Failed to configure AWS")
+		return
+	}
+
+	fmt.Println("SSM path:", ssmPath)
+	getSet(ssmPath, "")
+}
+
+func configureAWS() (err error) {
+	awsRegion := os.Getenv("AWS_REGION")
+	awsDefaultRegion := os.Getenv("AWS_DEFAULT_REGION")
+
+	if strings.TrimSpace(awsRegion) == "" && strings.TrimSpace(awsDefaultRegion) == "" {
+		logger.Debug("ERROR: $AWS_REGION && $AWS_DEFAULT_REGION are empty (need 1 exported). Unable to set SSM parameters")
+		err = errors.New("Bad region env vars")
+		return
+	}
+
+	region := awsDefaultRegion
+	if strings.TrimSpace(awsRegion) != "" {
+		region = awsRegion
+	}
+	cfg := aws.NewConfig().WithRegion(region)
+
+	sess := session.New(cfg)
+	svcSSM = ssm.New(sess)
+
+	return
+}
+
+func getSet(ssmPath, nextToken string) {
+	in := &ssm.GetParametersByPathInput{
+		Path:           &ssmPath,
+		WithDecryption: aws.Bool(true),
+	}
+
+	if nextToken != "" {
+		in.SetNextToken(nextToken)
+	}
+
+	out, err := svcSSM.GetParametersByPath(in)
+	if err != nil {
+		logger.Error("Failed getting parameter by path:", err)
+		return
+	}
+
+	for _, parameter := range out.Parameters {
+		setParameter(ssmPath, parameter)
+	}
+
+	if out.NextToken != nil {
+		getSet(ssmPath, *out.NextToken)
+	}
+
+}
+
+func setParameter(ssmPath string, parameter *ssm.Parameter) {
+	if parameter == nil {
+		logger.Error("SSM parameter is nil")
+		return
+	}
+
+	name := ""
+	if parameter.Name != nil {
+		name = *parameter.Name
+	}
+
+	value := ""
+	if parameter.Value != nil {
+		value = *parameter.Value
+	}
+
+	length := len(ssmPath)
+	if len(ssmPath) == 1 {
+		length = 0
+	}
+
+	key := strings.Replace(strings.Trim(name[length:], "/"), "/", "_", -1)
+	value = strings.Replace(value, "\n", "\\n", -1)
+
+	os.Setenv(key, value)
+	logger.Debugf("Setenv: key=%s", key)
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 dist/
 .idea/
 helmfile
+vendor
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -26,6 +26,47 @@
   version = "v1.0.1"
 
 [[projects]]
+  digest = "1:9cddbbc9d2c109832cb83e92c05d939b56dac13653219d7a6ed8f5b1a645cd5a"
+  name = "github.com/aws/aws-sdk-go"
+  packages = [
+    "aws",
+    "aws/awserr",
+    "aws/awsutil",
+    "aws/client",
+    "aws/client/metadata",
+    "aws/corehandlers",
+    "aws/credentials",
+    "aws/credentials/ec2rolecreds",
+    "aws/credentials/endpointcreds",
+    "aws/credentials/processcreds",
+    "aws/credentials/stscreds",
+    "aws/csm",
+    "aws/defaults",
+    "aws/ec2metadata",
+    "aws/endpoints",
+    "aws/request",
+    "aws/session",
+    "aws/signer/v4",
+    "internal/ini",
+    "internal/sdkio",
+    "internal/sdkrand",
+    "internal/sdkuri",
+    "internal/shareddefaults",
+    "private/protocol",
+    "private/protocol/json/jsonutil",
+    "private/protocol/jsonrpc",
+    "private/protocol/query",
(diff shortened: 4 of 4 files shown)


ssm/ssm.go
```diff
@@ -0,0 +1,117 @@
...
+
+var (
+	svcSSM  *ssm.SSM
+	ssmPath string
+	logger  *zap.SugaredLogger
+)
+
+func Run() {
+	logger = helmexec.NewLogger(os.Stdout, "debug")
+	ssmPath, exists := os.LookupEnv("SSM_PATH")
```
osterman: Ideally we could support multiple paths. Perhaps use `;` delimiter like other `PATH` vars

ssm/ssm.go
```diff
@@ -0,0 +1,117 @@
...
+	if parameter.Value != nil {
+		value = *parameter.Value
+	}
+
+	length := len(ssmPath)
+	if len(ssmPath) == 1 {
+		length = 0
+	}
+
+	key := strings.Replace(strings.Trim(name[length:], "/"), "/", "_", -1)
```
osterman: For greater interoperability with `chamber` which lowercases all parameters, it would be nice if there was an option to uppercase them for environment variables.

ssm/ssm.go
```diff
@@ -0,0 +1,117 @@
...
+
+	if out.NextToken != nil {
+		getSet(ssmPath, *out.NextToken)
+	}
+
+}
+
+func setParameter(ssmPath string, parameter *ssm.Parameter) {
+	if parameter == nil {
+		logger.Error("SSM parameter is nil")
```
osterman: I think this is more of a warning or debug.

ssm/ssm.go
```diff
@@ -0,0 +1,117 @@
...
+
+	length := len(ssmPath)
+	if len(ssmPath) == 1 {
+		length = 0
+	}
+
+	key := strings.Replace(strings.Trim(name[length:], "/"), "/", "_", -1)
+	value = strings.Replace(value, "\n", "\\n", -1)
+
+	os.Setenv(key, value)
```
osterman: Maybe warn if overwriting an existing variable

sgandon:
Hi guys, could you explain what is SSM, and what does this have in common with helmfile and this PR brings.
I am not sure to understand the need for this ?

rms1000watt:
@sgandon Ah, yes, will do!

**What is AWS SSM**

AWS SSM: https://docs.aws.amazon.com/systems-manager/latest/userguide/what-is-systems-manager.html

Specifically, we use Parameter Store within SSM: https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html

At a high level, AWS SSM Parameter Store is a free KV store provided by AWS. I've used it across various projects & companies. @osterman has https://cloudposse.com/ where his teams use it across multiple projects & companies as well. So it's really popular.

**What's the intersection between SSM & Helmfile?**

Helmfile is very powerful for many reasons! One of the awesome features is that we can `{{requireEnv "postgres_password"}}` inside the Helmfile YAML. But then the question is, how do you get that (secret) value into your environment so Helmfile can use it?

What we do currently is use tools like https://github.com/segmentio/chamber and https://github.com/Droplr/aws-env to extract SSM variables and add to our Env before running `helmfile` commands. This results in extra logic/dependencies in our CI/CD workflows.

This PR was aimed at solving that issue of the extra logic/dependencies.

**Next steps**

There are a few different approaches to take:

- Address the suggestions in this PR proposed by @osterman and merge
- Consider a larger approach at handling secret values & adding to the environment (outside of AWS SSM). (AWS SSM would be one implementation of a Golang interface for this `SecretGetter` idea.)
- Consider a larger approach of retrieving secret values directly from source as defined in the helmfile (for example serverless.com does SSM retrieval natively, amongst others: https://serverless.com/framework/docs/providers/aws/guide/variables#reference-variables-using-the-ssm-parameter-store)

aegershman:
Interesting. Just piggybacking to say I have a use-case for something similar, except using [`credhub` as the backing store](https://github.com/cloudfoundry-incubator/credhub) for secret retrieval. Current solution is like what you described in original post: export secrets as an environment variable with `.envrc` and `direnv` or something, but it's not ideal. Just commenting with this use-case to point out that there could be value in other secret management backends in the future 👍

sgandon:
@rms1000watt thanks a lot for the very clear exaplanation.
I think this is a very interresting idea and I also believe we need some more pluggable and general mechanism for handling values comming from any kind of persistant storage.
You seem to be focused on secret values, but is there also a need for handling any kind of values ?
I would imagin some additions to the helmfile yaml definition in order to configure specific external key/values stores.

rms1000watt:
@sgandon I could mock up something similar to how serverless does it:
https://github.com/rms1000watt/hello-world-serverless-golang/blob/master/serverless.yml#L3-L9
https://serverless.com/framework/docs/providers/aws/guide/serverless.yml/

But I think it would be more specific to implementations like:

```yaml
ssm:
  region: us-west-2
  prefix: /path/within/ssm
  # credentials for connecting to AWS would be omitted here and follow standard Golang SDK for connecting to AWS

repositories:
    - name: "my-repo"
      url: "s3://my-private-helm-chart-repo"

releases:
  - name: hello-world
    chart: my-repo/hello-world
    values:
    - hello-world:
        url: http://www.fake-url.com
        postgres_password: {{ ssm "postgres_password" }}
```

If and only if `ssm` is defined at the top level of the helmfile.yaml, then the `ssm` key for funcMap would work

(Add `ssm` to funcMap here: https://github.com/roboll/helmfile/blob/master/tmpl/context_funcs.go#L17-L40)

And to @aegershman 's point, the same structure could happen with credhub (although i've never used it and am not sure if this is makes sense with credhub).

But for other KV stores (consul, etcd) this might make sense. It's sort of like dependency injection at the YAML level.

What do you guys think?

sgandon:
That is a great proposal !.
I was more thining of a generic way of geting access to the values with a kind of build-in object like .Values but for external KV stores. 
But I think your approach with a specific function is even more flexible cause then you can have very fancy parameters if required, like in you case the path within ssm could also be a parameter of the ssm function : `{{ ssm "posgres_password" "/path/within/ssm"}}`.
And also mix multiple key stores if required.
This is really nice.

main.go
```diff
@@ -569,6 +571,7 @@ Do you really want to delete?
 		},
 	}
 
+	ssm.Run()
```
mumoshu: You've brought me a very good idea making environment variables as the interface between helmfile and kv-backend like ssm. Good job.

mumoshu: I think we can merge this as an experimental feature and start implementing more like this once you wrap behind a feature flag. Perhaps something like `if _, exists := os.LookupEnv("SSM_PATH"); exists { ssm.Run() }`?

sgandon: I agree this would be nice to have the `ssm.Run` not triggerd everytime helmfile is launched and you don't care about ssm at all.

rms1000watt: @sgandon @mumoshu look at the second line of the ssm.Run() function. It does that check already. I just didn’t want to make the main function messy with a bunch of checks.

rms1000watt: But also, I don’t mind just working on that more flexible solution proposed above. Like, we can keep this PR open (but on hold) just in case I hit a snag with that code implementation. 

I really just wanted to throw something together here to start the conversation so we could structure something awesome together. ☺️

ssm/ssm.go
lanmalkieri: Definitely think we should have the ability to support multiple paths. 

Would you rather define each path in the environment? Or just directly in the helmfile? I think it makes more sense to just put right in the helmfile.

rms1000watt: I'm looking at updating the helmfile.yml spec to include `ssm:` at the root level of the YAML.

```yaml
ssm:
  region: us-west-2
  prefix: /path/in/ssm
```

The idea is then the prefix is optional when you start using ssm directly

```
{{ ssm "postgres_pass" }}
```

instead of 

```
{{ ssm "/path/in/ssm/postgres_pass" }}
```

But of course you can do something like..

```
{{ ssm "/path/one/postgres_pass" }}
{{ ssm "/path/two/redis_pass" }}
```

rms1000watt:
@sgandon @mumoshu @osterman Here is a newer proposal I'm going to start working on, after some discussions with @lanmalkieri. This incorporates some of their requests to handle multiple paths. (This assumes they're all from the same AWS account. Future iterations could specify an assume role or something? But that's down the road for SSM.)

```yaml
ssm:
  - region: us-west-2
    prefix: /global/path/to/ssm
    name: global
  - region: us-east-1
    prefix: /us-east-1/path/to/ssm
    name: east1

repositories:
  - name: "my-repo"
    url: "s3://my-private-helm-chart-repo"

releases:
  - name: hello-world
    chart: my-repo/hello-world
    values:
    - hello-world:
        url: http://www.fake-url.com
        postgres_password: {{ ssm "global:postgres_password" }}
        redis_password: {{ ssm "east1:redis_password" }}
```

osterman:
Also, just throwing this out there... Gomplate supports lots of data sources. For an interface, it could be helpful to reference. 

https://github.com/hairyhenderson/gomplate/blob/master/docs/content/datasources.md

## related
- https://github.com/roboll/helmfile/issues/388#issuecomment-480270012
- https://github.com/roboll/helmfile/issues/392#issuecomment-455061273

rms1000watt:
@osterman the only reason I'm not crazy about this solution: https://github.com/roboll/helmfile/issues/392#issuecomment-455061273

is because the `datasources` is buried within a release. When you have multiple releases in one helmfile, you'd have to redeclare the `datasources`.

But yeah, I like that `gomplate` handles a lot of datasources already. That would give a ton of functionality out of the box.. hmm..

```yaml
datasources:
  ssmGlobal: aws+smp:///global/path/to/ssm
  ssmEast1: aws+smp:///us-east-1/path/to/ssm
  myvault: vault://vault.example.com:8200///secret-engine/my-application

repositories:
  - name: "my-repo"
    url: "s3://my-private-helm-chart-repo"

releases:
  - name: hello-world
    chart: my-repo/hello-world
    values:
    - hello-world:
        url: http://www.fake-url.com
        postgres_password: {{ (ds "ssmGlobal" "postgres_password").Value }}
        redis_password: {{ (ds "ssmEast1" "redis_password").Value }}
        es_password: {{ (ds "myvault" "es_password").Value }}
```

Somethingggg like that.. would have to play around with it

rms1000watt:
@osterman Gomplate might not be flexible enough in the long run for SSM. I might run with it for now though. I can imagine cases where we're pulling SSM values from different accounts and/or regions, and I'm not seeing anything in Gomplate to specify that: https://github.com/hairyhenderson/gomplate/blob/master/docs/content/datasources.md#using-awssmp-datasources

rms1000watt:
@osterman Doing this with native gomplate syntax in the helmfile.yml feels weird:

```yaml
es_password: {{ gomplate `ds=aws+smp:///dev/us-west-2/ssm` `{{ (ds "ds" "es_password").Value }}` }}
```

with code:

```go
	cfg := &gomplate.Config{
		DataSources: []string{ds},
		Input:       key,
		OutputFiles: []string{tmpfile.Name()},
	}

	if err = gomplate.RunTemplates(cfg); err != nil {
		return
	}
```

I have it working with the code above.

But to make it more convenient for helmfile users.. I'd want to to create a translation layer so it's really simple. But at that point, then there's no real benefit to the end user if I use gomplate internally or not. So I'm actually going to look back at this syntax to make it easier for end users:

```ssm:
  - region: us-west-2
    prefix: /global/path/to/ssm
    name: global
  - region: us-east-1
    prefix: /us-east-1/path/to/ssm
    name: east1

releases:
  - name: hello-world
    chart: stable/hello-world
    values:
    - hello-world:
        url: http://www.fake-url.com
        postgres_password: {{ ssm "global:postgres_password" }}
        redis_password: {{ ssm "east1:redis_password" }}```

rms1000watt:
Also, gomplate is weird to try and integrate into your code base (unless I'm totally just looking at it the wrong way). The repo doesn't expose that many public functions from their packages: https://godoc.org/github.com/hairyhenderson/gomplate

So I had to do things like write to a temporary file for each variable, instead of keeping them in memory. (gomplate seems like it was designed for just CLI use only? I'm open to learning more if anyone else has a different opinion.)

rms1000watt:
Closing in favor of: https://github.com/roboll/helmfile/pull/573

sgandon:
I don't know about gomplate but I think this is really promizing. Thanks a lot for having a detailed look at it. You are pointing out many limitation of the library but do you think we could contribute to it to make more easily "integratable" ? cause it is nice promise to avoid duplicating some code already working and test that provide the feature we want.

PR #573, opened 2019-05-03
Added AWS SSM Parameter Store Integration

rms1000watt:
See discussion here: https://github.com/roboll/helmfile/pull/569

@osterman @mumoshu @sgandon @lanmalkieri Here is working code for the YAMLs below.

This allows `ssm` to be defined and used:

```yaml
ssm:
  - name: east1
    prefix: /dev/us-east-1/redis
    region: us-east-1
  - name: west2
    prefix: /dev/us-west-2/redis
    region: us-west-2

releases:
  - name: redis-test
    chart: stable/redis
    values:
    - fake_key:
        redis_pass_east: {{ ssm "east1:redis_password" }}
        redis_pass_west: {{ ssm "west2:redis_password" }}
        redis_pass_seast: {{ ssm "/dev/ap-southeast-1/redis/redis_password" }}
    - cluster:
        enabled: false
    - persistence:
        enabled: false
    - metrics:
        enabled: false
```

You can reference an SSM declaration at the root of the yaml, or pass in the entire ssm_path as the argument. (The downside with passing in the full path as an argument is you rely on your $AWS_REGION or $AWS_DEFAULT_REGION to pull from SSM, rather than explicitly defining region at the top.)

Also, for the sake of convenience, you can omit the `ssm` declaration entirely and do this:

```yaml
releases:
  - name: redis-test
    chart: stable/redis
    values:
    - fake_key:
        redis_tag: {{ ssm "/dev/us-west-2/redis/redis_tag" }}
        redis_host: {{ ssm "/dev/us-west-2/redis/redis_host" }}
        redis_password: {{ ssm "/dev/us-west-2/redis/redis_password" }}
    - cluster:
        enabled: false
    - persistence:
        enabled: false
    - metrics:
        enabled: false
```


main.go
```diff
@@ -2,6 +2,9 @@ package main
 
 import (
 	"fmt"
+	"os"
```
sgandon: looking at these import refactor I am wondering if you are using the right formatter. Are you using Visual Studio code ?
I had the same issue until I changed the default format tool.
in VSCode go to preferences and type `go format tool` then pick *gofmt* instead of the default `goretruns`

sgandon:
Nice proposal, 
I believe you should also add some tests to your PR if at all possible.
I think we should also look at the gomplate solution, I'll try to contact the owner of the repo through slack to see what he thinks about it.

main.go
rms1000watt: hehe, yeah, VSCode, but using `goimports`. I don't have a strong opinion either way (I probably should though). But I don't mind reverting any import diffs in this PR.

rms1000watt:
@sgandon Ah, yeah, I can add some tests tonight.

As far as `gomplate` I think it could be a good solution down the road. However, right now I'm willfully against it.

There are 2 independent things I think about when creating cli tools:
- The user experience
- The underlying technology

**As far as user experience**

If helmfile makes claims to use gomplate syntax, then it should follow it to a T. Doing so, would look something like:

```yaml
redis_password: {{ gomplate `west2=aws+smp:///dev/us-west-2/ssm` `{{ (ds "west2" "redis_password").Value }}` }}
```

To simplify it could look like:

```yaml
datasources:
  west2: aws+smp:///dev/us-west-2/ssm

releases:
  - name: redis-test
    chart: stable/redis
    values:
    - fake_key:
        redis_password: {{`{{ (ds "west2" "redis_password").Value }}`}}
```

To simplify further, we could drop the `.Value` portion off, as suggested by @mumoshu  here https://github.com/roboll/helmfile/issues/392#issuecomment-455061273

But at this point, there are 2 unfavorable things:
- We deviated from `gomplate` syntax
- We're creating our own helmfile-gomplate syntax that users now have to learn

If helmfile users have to learn a new syntax, it ought to be dirt simple (not nested `{{` with strategically positioned backticks).

**As far as underlying technology**

I have a few issues with `gomplate` currently:
- It was designed as a cli tool (which is fine), but doesn't leave many public functions from the `gomplate` packages (as seen here: https://godoc.org/github.com/hairyhenderson/gomplate). To deal with these issues, you have to write the output of every `gomplate` call to a temporary file on the user's OS, opposed to keeping everything in memory (unless I'm totally missing something 😅).
- `gomplate` doesn't let you specify region when pulling from SSM. However, I imagine it's an easy fix to `gomplate` to add that functionality. (This PR adds this region functionality.)
- `gomplate` doesn't maintain the AWS session between subsequent SSM calls. (This PR maintains the AWS session between calls.)

**Future development concerns**

I understand the concern from the helmfile maintainers that `gomplate` could reduce code repeatability. However, this PR doesn't actually prevent a `gomplate` integration in the future. The folder structure is:

```
helmfile
├── datasource
│   ├── prepare.go
│   └── ssm.go
```

Which allows for other datasources: `vault`, `consul`, `boltdb`, etc. to be added as individual files with individual implementations. All the functions in `ssm.go` are prepended with either `SSM` or `ssm`. In the future (maybe even a week from now), we can always move `ssm.go` to a `gomplate` implementation. 

**However**, please keep in mind the user experience is independent to the underlying technologies. So pleazzzze don't sacrifice user convenience because of a technology decision 😅 ❤️

main.go
sgandon: I don't mind either way, it is just sometimes make it hard to focus on the real changes.

sgandon:
@rms1000watt both your comments on user XP and underlying technologies make perfect sense and I also agree that the chosen technology should not impose anything on the user experience. 
To my knowledge the user experience is being more and more important to get people to not throw your software away within minutes.

Let's wait and see if gomplate gets more easy to integrate ( https://github.com/hairyhenderson/gomplate/issues/548 ) in the future and think again about some nice user experience around this integration.

lanmalkieri:
@sgandon Hey there!

Just want to clarify, are you suggesting to wait for gomplate to get more easy to integrate, and to not continue work on this PR until that occurs?

hairyhenderson:
I'll try to answer a few of the issues raised above with gomplate:

> To simplify further, we could drop the `.Value` portion off

IMO having the `.Value` part in the returned data was a mistake and I'd like to change that behaviour in gomplate. The reason is that it deviates from how other datasources work (all the rest _just_ return the data, with no metadata attached). However, that would also mean a breaking change (which I'm not totally opposed to - it just means bumping the major version).

> * It was designed as a cli tool (which is fine), but doesn't leave many public functions from the `gomplate` packages (as seen here: https://godoc.org/github.com/hairyhenderson/gomplate). To deal with these issues, you have to write the output of every `gomplate` call to a temporary file on the user's OS, opposed to keeping everything in memory (unless I'm totally missing something 😅).

Right now that's true, and I'm motivated to fix that (which is why I filed hairyhenderson/gomplate#548).

> * `gomplate` doesn't let you specify region when pulling from SSM. However, I imagine it's an easy fix to `gomplate` to add that functionality. (This PR adds this region functionality.)

Right, the only way to specify a region is by setting the `AWS_REGION` environment variable, or if running in EC2 it'll default to using the EC2 instance's region (if the EC2 metadata Service is accessible). But different per-datasource regions aren't supported right now. Nothing stopping this from being implemented - there's just never been a user asking for it 😉.

> * `gomplate` doesn't maintain the AWS session between subsequent SSM calls. (This PR maintains the AWS session between calls.)

Each separate SSM datasource maintains its own session, so in that sense, no, the session isn't maintained. But if using sub-paths on the same datasource (like with `ds "west2" "redis_password"` or `ds "west2" "some_other_password"`), the same session is re-used. If that's not the behaviour observed in gomplate, it's a bug 😉


As for syntax, obviously it's up to the helmfile maintainers and my opinion should bear no weight here, _but_ I really don't like the look of either of these forms:

```yaml
redis_password: {{ gomplate `west2=aws+smp:///dev/us-west-2/ssm` `{{ (ds "west2" "redis_password").Value }}` }}
```

```yaml
redis_password: {{`{{ (ds "west2" "redis_password").Value }}`}}
```

If you're going to use gomplate's datasource support, it should be way more user-friendly than that.

> **However**, please keep in mind the user experience is independent to the underlying technologies. So pleazzzze don't sacrifice user convenience because of a technology decision 😅 ❤️

💯 agree - if gomplate too awkward to use for helmfile's purposes, then it's probably best not to! (though I would _totally_ appreciate issues filed so I can fix that awkwardness 😉)

sgandon:
@lanmalkieri no, I suggest to proceed with this PR.
We will see when gomplate reaches a level where we can integrate it with a nice user experience and hopfully make the work to add it to helmfile to "inherits" from the datasources it provides. I'd be happy to help on that.

mumoshu:
@rms1000watt You've made a great proposal! I love this. Your implementation and expectation on good U/X sounds very valid to me. Let's go ahead with your apparoach!

---

Note that, if you forgive me becoming nit-picky, my concern on this approach is that this:

```yaml
releases:
  - name: redis-test
    chart: stable/redis
    values:
    - fake_key:
        redis_pass_east: {{ ssm "east1:redis_password" }}
        redis_pass_west: {{ ssm "west2:redis_password" }}
        redis_pass_seast: {{ ssm "/dev/ap-southeast-1/redis/redis_password" }}
```

ends up querying `ssm` even when you've disabled the release by setting `installed: false`:

```yaml
releases:
  - name: redis-test
    chart: stable/redis
    installed: false
    values:
    - fake_key:
        redis_pass_east: {{ ssm "east1:redis_password" }}
```

This happens because `helmfile` currently have basically two template phases, one for the **load time** and one for **sync time**. The former **load time** templating happens when loading `helmfile.yaml`, so that you can dynamically change the structure of your `helmfile.yaml` with go template exrepssions. The latter **sync time** templating happens when syncing each release - so it doesn't happen on releases disabled with `installed: false`.

To corp with `installed: false`, I think `ssm` should be eventually moved to the latter phase, by preventing it from evaluated in the former phase using go template escapes:

```
releases:
  - name: redis-test
    chart: stable/redis
    values:
    - fake_key:
        redis_password: {{`{{ ssm "east1:redis_password" }}`}}
```

Actually that's why I've used the verbose syntax with the escapes in my proposal https://github.com/roboll/helmfile/issues/392#issuecomment-455061273 you also mentioned.

rms1000watt:
@hairyhenderson First I want to say, I suuuuuper appreciate your response. It's rare being able to talk to authors about their open source projects (they usually get too busy) Again, super awesome!! 😄 

1 thing I want to point out about this:

> Each separate SSM datasource maintains its own session, so in that sense, no, the session isn't maintained. But if using sub-paths on the same datasource (like with ds "west2" "redis_password" or ds "west2" "some_other_password"), the same session is re-used. If that's not the behaviour observed in gomplate, it's a bug 😉

I 100% understand what you're saying and it makes sense. The only catch is _how often_ does `gomplate` get called? If `gomplate` was being used as designed and running once against 1 template for 1 render.. then I'm with ya. But with helmfile, the `ds` or `gomplate` or `ssm` is a function that runs once for each occurrence of that function-name. So, if `ds` is written 3 times, it would be 3 separate calls of `gomplate` for rendering 1 string at a time. Since there isn't a way to hook `gomplate` state into memory, I imagine it's re-instantiating everything, every time. 🤷‍♂In the grand scheme of things, it's not a _big_ deal. Honestly, I was just trying to think architectural differences to make this PR more substantial. 🤷‍♂ 

And again man, I appreciate all your openness talking about the project! I apologize if I'm too blunt or anything. Much appreciate everything! @hairyhenderson 

---
@mumoshu pleaaaaase nit to your heart's content. I'm the same way and encourage it!!

Ohh, that explanation helps a lot. I noticed SSM was running twice for each and every variable call. What is your concern though? (I understand your observation of what's happening and explanation of how it's happening.. but _what is the problem?_

When I saw this behavior, I thought it would be super lame to hit AWS twice for every variable (for many reasons), so I added this to cache variables I've already received: https://github.com/roboll/helmfile/pull/573/files#diff-94e9c44ce590540ca463d78859c012ebR65

🤞Hopefully this ^^^ addresses your concerns. I just really want to do all I can to prevent people from having to use backticks and nested `{{` in their templates (they're not obvious concepts for newcomers).

mumoshu:
> but what is the problem?

What I had in my mind were (1)unnecessary calls to SSM, (2)duplicate calls to SSM, (3)inconsistent template functions U/X. Maybe (1) and (3) are the remaining problems.

Regarding (1), `{{ ssm ... }}` results in a SSM API call even though the release is disabled by setting `installed: false`. Suppose an admin restricted my IAM policy to forbid access to certain set of SSM params and we have a common `helmfile.yaml` containing all the apps my team maintains. I would try to disable only the release requires the set of SSM params by setting `installed: false`, but it would result in 403s due to that helmfile stills attempts to read the SSM params for the disabled release.

Regarding (2), the second templating phase is executed only once per release so using backticks to defer templating to the second solves the duplicate calls issue by its nature. But you've addressed this by another way of caching SSM calls, which would work as well!

Regarding (3), I think we need backticks for [Release Template](https://github.com/roboll/helmfile/blob/master/docs/writing-helmfile.md#release-template--conventional-directory-structure), especially when you use `Release.Name` in a template expression. `Release` is available only in the second templating phase, so evaluating an expression containing that, like `values: ["{{ .Release.Name }}/values.yaml"]`, ends up with an error due to missing `Release` in the templating context.

Deferring it like:

```
values: ["{{` {{ .Release.Name }} `}}/values.yaml"]
```

would indeed work.

But what if we suggest NOT using backticks for SSM params? The helmfile.yaml would end up with the mix of expressions with and without backticks. I feel like it isn't very intuitive, and maybe it's better to always add backticks for release templates?

```
values:
- "{{` {{ .Release.Name }} `}}/values.yaml"
- fake_key:
    redis_password: {{ ssm "east1:redis_password" }}
```

hairyhenderson:
@rms1000watt 

> @hairyhenderson First I want to say, I suuuuuper appreciate your response. It's rare being able to talk to authors about their open source projects (they usually get too busy) Again, super awesome!! 😄

🙇 You're welcome!

> The only catch is how often does gomplate get called? If gomplate was being used as designed and running once against 1 template for 1 render.. then I'm with ya. But with helmfile, the ds or gomplate or ssm is a function that runs once for each occurrence of that function-name. So, if ds is written 3 times, it would be 3 separate calls of gomplate for rendering 1 string at a time. Since there isn't a way to hook gomplate state into memory, I imagine it's re-instantiating everything, every time.

Ok, I think I understand. If you call gomplate's [`data.Datasource`](https://godoc.org/github.com/hairyhenderson/gomplate/data#Data.Datasource) function, the state should be persisted. I've thrown a quick example up at https://gist.github.com/hairyhenderson/4555ad1571a1a33597dafb8d429b33e1 - feel free to test it out. 

The `data.Datasource` function is what's bound to the `ds`/`datasource` template function in gomplate, so it should work exactly the same as if you were using the CLI.

rms1000watt:
@mumoshu beautiful perspectives!! Thank you for the clarity!!

> Regarding (1), {{ ssm ... }} results in a SSM API call even though the release is disabled by setting installed: false. Suppose an admin restricted my IAM policy to forbid access to certain set of SSM params and we have a common helmfile.yaml containing all the apps my team maintains. I would try to disable only the release requires the set of SSM params by setting installed: false, but it would result in 403s due to that helmfile stills attempts to read the SSM params for the disabled release.

Hmm, do you know of a way to get this state into the context so I can disable external calls if `context.State.Releases[0].installed == false` inside the `context_funcs.go`?

Or should I just skip all SSM on `c.preRender`, like this?

```go
func (c *Context) SSM(key string) (val string, err error) {
        if c.preRender {
                return
        }
	return datasource.SSMGet(key)
}
```

Regarding (3):

> I feel like it isn't very intuitive, and maybe it's better to always add backticks for release templates?

I really don't want to influence the way things are done now (I'm not aware of all the ramifications). I was hoping to just integrate with the way things are done already. 🤷‍♂ 

For the template in template thing.. (thinking about context in `context_funcs.go`) is there a way to:

```go
// pseudocode
func (c *Context) SSM(key string) (val string, err error) {
        c.RenderTemplateToBuffer(key, c.state)
        // Add error handling above and format result for consumption below
	return datasource.SSMGet(key)
}
```

I'm just trying to think how to address (1) and (3).

rms1000watt:
@hairyhenderson Your gist: https://gist.github.com/hairyhenderson/4555ad1571a1a33597dafb8d429b33e1 is freaking awesome. Damn.. this is making me rethink some of the implementation on the code side.

Going to noodle on this one.

mumoshu:
@rms1000watt Thanks for your effort!

> Hmm, do you know of a way to get this state into the context so I can disable external calls if context.State.Releases[0].installed == false inside the context_funcs.go?

Unfortunately, this is impossible without backticks. Without backticks, the "context" of your function only contains Environment.Name and Environment.Values, because it happens before parsing `releases:`.

> For the template in template thing.. (thinking about context in context_funcs.go) is there a way to

I wasn't good enough to understand your intention in the pseudocode. But after reading it I came up with a syntax like this, as an alternative way than backticks to "defer" template evaluations:

```
redis_pass_seast: {{ ssm "/dev/ap-southeast-1/redis/redis_password" }}
```

mumoshu:
@rms1000watt The foundational problem helmfile has is that `helmfile.yaml` is actually a go template, not a regular YAML file as its file extension states.

I should have required it to have explicit `.gotmpl` extensions like `helmfile.yaml.gotmpl` to treat it a YAML or go template. So that we can selectively enable the whole-file template.

If it was a regular YAML file, we don't need to use backticks in `values:` arrays. We instead need single quotes like `redis_pass_seast: '{{ ssm "/dev/ap-southeast-1/redis/redis_password" }}'`, not `redis_pass_seast: {{ ssm "/dev/ap-southeast-1/redis/redis_password" }}`,as the latter is invalid as YAML 😓 

Do you think we can change helmfile to require the explicit file extension `.gotmpl` to enable the whole file templating?

rms1000watt:
@mumoshu Soooo, there's a little bit of an easter egg in my PR. I purposefully added `os.ExpandEnv()` in a few places so I can do things like:

```yaml
ssm:
  - name: ssm1
    prefix: /${TIER}/${REGION}/redis
    region: ${REGION}

releases:
  - name: redis-test
    chart: stable/redis
    values:
    - fake_key:
        redis_pass_east: {{ ssm "ssm1:redis_password" }}
        redis_pass_west: {{ ssm "ssm1:redis_password" }}
        redis_pass_seast: {{ ssm "/${TIER}/ap-southeast-1/redis/redis_password" }}
    - cluster:
        enabled: false
```

Where `TIER` and `REGION` have to be in the env.

As seen here:
- https://github.com/roboll/helmfile/pull/573/files#diff-94e9c44ce590540ca463d78859c012ebR48
- https://github.com/roboll/helmfile/pull/573/files#diff-94e9c44ce590540ca463d78859c012ebR62

This kind of remedies the problems above about nested templating.. but this functionality is a deviation from native `helmfile` and wasn't documented in the PR.

---

The long term solution would be to change up the logic in `helmfile` where a lexer/parser can run templating from an arbitrarily-deep nested template: 

```
{{ ssm "/{{ requiredEnv "TIER"}}/{{ requiredEnv "REGION"}}/{{exec "./{{requiredEnv "REGION"}}.sh"}}" }}
```

However, I think short term solutions could be something like in https://github.com/roboll/helmfile/issues/584

tmpl/context_funcs.go
```diff
@@ -206,3 +210,7 @@ func RequiredEnv(name string) (string, error) {
 
 	return "", fmt.Errorf("required env var `%s` is not set", name)
 }
+
+func SSM(key string) (val string, err error) {
+	return datasource.SSMGet(key)
```
mumoshu: Not to say it is required in order to merge this, but just to note - I think we might want `datasource` to be in the `context` and separate per a context.

Alternatively it can be achieved by moving `datasource.PrepareAll(c)` to somewhere so that it runs once per each `helmfile.yaml` and its sub-helmfiles, assuming `helmfile` runs them one by one. `datasource` seems like a global state so moving it into the template context would be good to keep helmfile maintainable.

mumoshu:
> The long term solution would be to change up the logic in helmfile where a lexer/parser can run templating from an arbitrarily-deep nested template

Seems like we end up implementing a big beast that looks similar to but different than golang templates. But that would definitely be one way towards the solution!

Also, thanks for the easter egg 👍 

Just wanted to confirm to better understand it - Does it mean that it is an alternative syntax to `{{ requiredEnv "TIER" }}` that evaluates BEFORE go templates, and it can't be enhanced to cover other use-cases like nesting go template expressions?

rms1000watt:
> Seems like we end up implementing a big beast that looks similar to but different than golang templates. But that would definitely be one way towards the solution!

Golang templating would still be the template/render engine, but there would be a preprocessing step to determine a dependency graph. Then the rendering happens at each level of the graph in the appropriate order. I've also seen brute force attempts where recursion was used to render until 0 errors with a recursion depth of 100 (arbitrary number). But yes, it would be beast.. maybe not worth the effort--but definitely a more robust solution.

>  Does it mean that it is an alternative syntax to {{ requiredEnv "TIER" }} that evaluates BEFORE go templates, and it can't be enhanced to cover other use-cases like nesting go template expressions?

- It's *almost* alternative, except `os.ExpandEnv()` gives no error if the env-var is undefined
- Ahhh, I don't recall if it runs *before* or *after*
- It wouldn't be hard to lex/parse nested template expression, but I don't know if I have a mechanism to render them. This goes back to my comment above about using `c.RenderTemplateToBuffer(key, c.state)`. If we can run that function, then we could start testing templates without affecting the UX by adding extra backticks or single quotes.

tmpl/context_funcs.go
rms1000watt: @mumoshu this should probably be a blocker for the PR honestly. I only ever tested with 1 helmfile.

https://github.com/roboll/helmfile/pull/573/files#diff-94e9c44ce590540ca463d78859c012ebR56

The `ssmGetSpecs()` runs once to parse the helmfile to get the `ssm` definition from the root of the yaml

```yaml
ssm:
  - name: ssm1
    prefix: /dev/us-east-1/redis
    region: us-east-1
``` 

So this logic would have to be enhanced to parse multiple helmfiles.

Honestly, I'm not aware of what this looks like. I'll have to do my homework and learn `helmfile` a bit more.

datasource/ssm.go
```diff
@@ -0,0 +1,224 @@
...
+	if strings.TrimSpace(ssmFile) == "" {
+		logger.Warn("SSM spec missing in helmfile; running with defaults")
+		return
+	}
+
+	file, err := os.Open(ssmFile)
+	if err != nil {
+		return
+	}
+
```
rms1000watt: Crap.. forgot:

```go
defer file.Close()
```

mumoshu:
@rms1000watt Thanks for the response!

> Golang templating would still be the template/render engine, but there would be a preprocessing step to determine a dependency graph. Then the rendering happens at each level of the graph in the appropriate order. I've also seen brute force attempts where recursion was used to render until 0 errors with a recursion depth of 100 (arbitrary number)

Interested in how that works!

rms1000watt:
@mumoshu This was the man, the legend: @jdtobe 

I wonder if we could persuade him into something 😉

tmpl/context_funcs.go
rms1000watt: Let me look at this again. So this can run per `helmfile.yaml`

rms1000watt:
@mumoshu I added `ssm_test.go`

still looking at datasource.Prepare(c) within the `converge` function

rms1000watt:
Closing in favor of: https://github.com/roboll/helmfile/pull/662

Issue #565, opened 2019-04-25
Feature Request: make helmfile to understand .yml


afoninsky:
There are bunch of tools which generates `yml` extension instead of `yaml`. For example I'm able to create `helmfile.d/*.yml` files using `kapitan` tool, but unable to create `helmfile.d/*.yaml` as extension is hardcoded. In this case it's hard to do `helmfile diff` out of the box ... this makes me sad.

Is it possible to support both? Thx.
https://stackoverflow.com/questions/21059124/is-it-yaml-or-yml/21059164

mumoshu:
@afoninsky Hey! Thanks for the feature request. That makes sense, and implementation would be simple as adding a `yml` glob after the one for `yaml` in https://github.com/roboll/helmfile/blob/fa95e0dd920fa953cb1c7eea245b2dbf6a735c15/pkg/app/app.go#L287-L290.

I'd greatly appreciate it if you could submit a PR!

afoninsky:
Sure, thank you fo pointing it. I've not found tests according to extensions, feel free to tell me if you need some.

afoninsky:
Sure, thanks for pointing to it. I've not noticed tests which verificate extensions, I guess they are not needed? Indicate if I missed them or eed to create them, thx again :)

mumoshu:
@afoninsky We currently don't have one for the file extension. I'd appreciate it if you could add test cases for that ☺️

PR #575, opened 2019-05-05
feat: add .yml extension support

afoninsky:
Adds .yml support according to https://github.com/roboll/helmfile/issues/565 discussion

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -2,9 +2,6 @@ package app
 
 import (
 	"fmt"
-	"github.com/roboll/helmfile/helmexec"
-	"github.com/roboll/helmfile/state"
-	"go.uber.org/zap"
 	"io/ioutil"
 	"log"
 	"os"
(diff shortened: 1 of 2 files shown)


pkg/app/app.go
```diff
@@ -284,11 +285,11 @@ func (a *App) findDesiredStateFiles(specifiedPath string) ([]string, error) {
 		} else if defaultFile != "" {
 			return []string{defaultFile}, nil
 		} else {
-			return []string{}, fmt.Errorf("no state file found. It must be named %s/*.yaml, %s, or %s, or otherwise specified with the --file flag", DefaultHelmfileDirectory, DefaultHelmfile, DeprecatedHelmfile)
+			return []string{}, fmt.Errorf("no state file found. It must be named %s/*.{yaml,yml}, %s, or %s, or otherwise specified with the --file flag", DefaultHelmfileDirectory, DefaultHelmfile, DeprecatedHelmfile)
 		}
 	}
 
-	files, err := a.glob(filepath.Join(helmfileDir, "*.yaml"))
+	files, err := a.glob(filepath.Join(helmfileDir, "*.{yaml,yml}"))
```
mumoshu: This seems not work in golang as `filepath.Glob` doesn't support braces expansion, but I think we can happily use `"*.y*ml"` instead. https://play.golang.org/p/77in7sAcorP

PR #576, opened 2019-05-05
removes dictionary key for subhelm and uses selectorsInherited

sgandon:
following @mumoshu  comments here https://github.com/roboll/helmfile/pull/567, I removed the usage of subhelmfile path as map key.
I also introduced the *selectorsInherited* key for explicit parent selector inheritance.

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -1422,14 +1422,6 @@ func (hf *SubHelmfileSpec) UnmarshalYAML(unmarshal func(interface{}) error) erro
 					if err := extractSelectorContent(hf, v); err != nil {
 						return err
 					}
-				} else {
-					hf.Path = key
-					//get the selectors if something is specified
-					if v != nil { //we have a path, now compute the selector
-						if err := extractSelector(hf, v); err != nil {
-							return err
-						}
-					} //else it is a path with ending semi-colon without anything else below and it is fine
 				}
 			default:
 				return fmt.Errorf("Expecting a \"string\" scalar for the helmfile collection but got: %v", key)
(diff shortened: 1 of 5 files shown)


state/state.go
```diff
@@ -1409,90 +1407,29 @@ func (hf *SubHelmfileSpec) UnmarshalYAML(unmarshal func(interface{}) error) erro
...
-	case map[interface{}]interface{}:
-		if len(selectors) == 0 {
-			hf.Selectors = make([]string, 0) //allocate and non nil empty array
-		} else { //unexpected unempty map so error
-			return fmt.Errorf("unexpected unempty map in selector [-%v] but got: %v", hf.Path, selectors)
-		}
-	default:
-		return fmt.Errorf("Expecting list of strings or and empty {} mapping [-%v] but got: [%v] of type [%T] ", hf.Path, selectors, selectors)
+	//also exclude SelectorsInherited to true and explicit selectors
+	if hf.SelectorsInherited && len(hf.Selectors) > 0 {
```
mumoshu: I like this guard!

sgandon: @mumoshu about this. I was wondering if we could not accept inheritance and selectors like this
```
helmfiles:
- path: /infrastructure-helmfile.yaml
  selectors:
  - name=grafana
  - name=prometheus
  SelectorsInherited : true
```
where if nothing comes from the parent helmfile or the command line then only grafana and prometheus are selected and if there is parent selectors then they would be combined together, meaning inherited selectors added to the default selectors.
What do you think ?

PR #583, opened 2019-05-09
fix: Send log messages to STDERR

### LINKED ISSUE #551: Template subcommand is "invalid"
Hi all,

The `template` command is printing more than templates. The command print information message:
```
helmfile template | grep Decrypting
Decrypting secret [REDACTED]/prod/secrets.yaml
Decrypting [REDACTED]/prod/secrets.yaml
```

This result in invalid kubernetes templates:
```
helmfile template | kubectl diff -f -
error: unable to decode "STDIN": json: cannot unmarshal string into Go value of type unstructured.detector
```

This can be easily fix with `grep -v -e '^Decrypting .*'` workaround, but I would suggest to adapt this.

EDIT: the `--log-level=warn` make both "Decrypting" message and rendered template disappearing. It's like the template is part of the info-log. IMO it should be a standard print.

Thank you.

mumoshu:
And only useful outputs from helm commands like ones from `helm template` to STDOUT.

Fixes #551

### DIFF
--- a/helmexec/exec.go
+++ b/helmexec/exec.go
@@ -1,6 +1,7 @@
 package helmexec
 
 import (
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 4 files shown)


faustos:
May I ask what was the reasoning behind this change? My octopus pipeline displays now useful status information as errors and marks the execution with warnings

PR #586, opened 2019-05-10
Various U/X improvements for `helmfile apply`

### LINKED ISSUE #540: remove "helm diff" errors when doing apply
when doing a `helmfile apply` there are errors coming from the `helm diff` plugin and this is really confusing.
I understand this is the `--detailed-exitcode` option of `helm diff` that does this but that would be nice to not have those errors displayed.
I would suggest either file the output in helmfile or may be contribute in to `helm diff` repos to have a `quiet` version.

### LINKED ISSUE #543: helmfile exits with an error when user chooses to not install a helmfile
launch an `helmfile apply` and say no when you are asked for applying the changes.
then the command exits with many errors and I should not expect any.
```
Do you really want to apply?
  Helmfile will apply all your changes, as shown above.

 [y/n]: n
err: exit status 2
err: exit status 2
err: exit status 2
err: exit status 2
err: exit status 2
changing working directory back to "/Users/sgandon/git-repos/helm-charts/platform-services/helmfile"
exit status 2
```
we are calling helmfile from another tool and we would not want this choice exits as an error.

mumoshu:
This improves the U/X of `helmfile apply`, by allowing you to selectively apply sub-helmfiles.
When you have two or more sub-helmfiles processed, typing `n` to cancel the first doesn't automatically stop the whole helmfile execution.
Instead, it proceeds by diffing the next sub-helmfile, and asks you to apply it, which should be what the user would expect.

To support this workflow, I have suppressed useless exec logs, correct exit status when diff exists in sub-helmfiles but not in the parent helmfile, and made the final error message emitted by helmfile better.

More concretely, this moves more output from `helm` to STDERR and the `debug` log-level.

The overall output from `helmfile` should be a bit more cleaner especially for `apply`, `sync`, `diff` and perhaps other `helmfile` sub-commands, too.

For example, when one of release failed, `helmfile`'s final error message now includes the error message from the failed `helm` execution, like seen in the last line:

```
List of updated releases :
RELEASE   CHART          VERSION
envoy     stable/envoy     1.5.0

List of releases in error :
RELEASE
envoy2
in ./helmfile.yaml: in .helmfiles[0]: in /Users/c-ykuoka/helmfile/helmfile.1.yaml: failed processing release envoy2: helm exited with status 1:
  Error: UPGRADE FAILED: "envoy2" has no deployed releases
```

This way you can better understand what caused helmfile to finally fail.

`helmfile` has been streaminig a lot of stdout and stderr contents from the `helm` commands regardless of the helmfile's log-level. It has been suppressed by default and moved to the `debug` log-level.
You will see that it helps you focus on what was the cause of a failure.

While working on the above, I found an another bug that made `--detailed-exitcode` useless in some case.
That is, `helmfile diff --detailed-exitcode`, when any diff existed only in sub-helmfiles, has been returning an exit code of `1`. It should return `2` when any release had diff and no release had an error, regardless of the target is a sub-helmfile or a parent helmfile. Why? Because that's what `--detailed-exitcode` meant for!

After this PR gets merged, `helmfile diff --detailed-exitcode`  propery return exit code `2` in such cases.

Fixes #543
Resolves #540

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -15,9 +15,6 @@ import (
 
 	"regexp"
 
-	"os/exec"
-	"syscall"
-
 	"net/url"
 
 	"github.com/roboll/helmfile/environment"
@@ -204,11 +201,12 @@ func (st *HelmState) SyncRepos(helm RepoUpdater) []error {
 
 type ReleaseError struct {
 	*ReleaseSpec
-	underlying error
+	err  error
+	Code int
 }
 
 func (e *ReleaseError) Error() string {
-	return e.underlying.Error()
+	return fmt.Sprintf("failed processing release %s: %v", e.Name, e.err.Error())
 }
 
 type syncResult struct {
@@ -250,7 +248,7 @@ func (st *HelmState) prepareSyncReleases(helm helmexec.Interface, additionalValu
 
 				flags, flagsErr := st.flagsForUpgrade(helm, release, workerIndex)
 				if flagsErr != nil {
-					results <- syncPrepareResult{errors: []*ReleaseError{&ReleaseError{release, flagsErr}}}
+					results <- syncPrepareResult{errors: []*ReleaseError{&ReleaseError{release, flagsErr, 1}}}
 					continue
 				}
 
@@ -258,14 +256,14 @@ func (st *HelmState) prepareSyncReleases(helm helmexec.Interface, additionalValu
 				for _, value := range additionalValues {
 					valfile, err := filepath.Abs(value)
 					if err != nil {
-						errs = append(errs, &ReleaseError{release, err})
+						errs = append(errs, &ReleaseError{release, err, 1})
 					}
 
 					ok, err := st.fileExists(valfile)
 					if err != nil {
-						errs = append(errs, &ReleaseError{release, err})
+						errs = append(errs, &ReleaseError{release, err, 1})
 					} else if !ok {
-						errs = append(errs, &ReleaseError{release, fmt.Errorf("file does not exist: %s", valfile)})
+						errs = append(errs, &ReleaseError{release, fmt.Errorf("file does not exist: %s", valfile), 1})
 					}
 					flags = append(flags, "--values", valfile)
 				}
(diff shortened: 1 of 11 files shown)


sgandon:
fanstastic proposal, I love the error message detail !

state/state.go
```diff
@@ -250,22 +248,22 @@ func (st *HelmState) prepareSyncReleases(helm helmexec.Interface, additionalValu
...
+					results <- syncPrepareResult{errors: []*ReleaseError{&ReleaseError{release, flagsErr, 1}}}
 					continue
 				}
 
 				errs := []*ReleaseError{}
 				for _, value := range additionalValues {
 					valfile, err := filepath.Abs(value)
 					if err != nil {
-						errs = append(errs, &ReleaseError{release, err})
+						errs = append(errs, &ReleaseError{release, err, 1})
```
sgandon: that would be great if 1 would be a constant with a name so that we know what is this parameter, it would make the code easier to read.

sgandon:
I am sorry I don't have much time to do a full review (I am on vacation tonight and need to finish some things) but that looks ok to me.

aegershman:
Apologies for bloating this PR with my comment, but:

Assuming you have a list of charts to apply in a `helmfile.yml`, would the workflow simply be `helmfile apply [name-of-some-chart name-of-some-other-chart]`?

state/state.go
mumoshu: Good point! I'd introduce three constants, `ExitStatusSucess`, `ExitStatusFailure`, `ExitStatusDiff` for `0`, `1`, `2` respectively.

mumoshu:
@aegershman Hey!

`helmfile` remains to be unable to accept two or more helmfile.yaml as separate command-line args. But I think you can use`helmfile -l name=name-of-some-chart -l name=name-of-some-other-chart apply`.

state/state.go
mumoshu: I ended up just avoid using magic numbers whereas possible. Hope it makes sense now!

PR #587, opened 2019-05-13
feat: helmfile.yaml layering enhancements

### LINKED ISSUE #388: Using helmfiles, but unable to pass Environment down to helmfile.yaml

### LINKED ISSUE #568: helmDefaults not interpreted when defined in separate file
I use helmfile.d for my releases and a helm-default.yaml file to define global values.
After upgrading to v 0.54.2, the defaults are no longer interpreted in the releases definitions.

I have to workaround this with a dirty inclusion of the helm default file in my release definitions like this:

```
{{ readFile "000-helm-defaults.yaml" }}
---
releases:
- name: myrelease1
....

```

The helmDefaults in separated file was working with version 0.45.3

### LINKED ISSUE #584: breaking change? `.gotmpl` should be required to enable templating of `helmfile.yaml`
Extracted from https://github.com/roboll/helmfile/pull/573#issuecomment-491142465

I think this, and the upcoming `bases` feature #388, would allow the user to choose either to prefer, (1) readability of release template expressions #439, or (2)less number of files involved.

Let's say we have a legacy helmfile.yaml below:

`helmfile.legacy.yaml`:

```
environments:
  {{ .Environment.Name }}:
    values:
    - environments/{{ .Environment.Name }}/values.yaml

releases:
- name: {{ .Environment.Values.releaseName }}
  values:
  - releases/{{` {{.Release.Name }} `}}/values.yaml
```

This can be reimplemented with `bases` like this:

`base.yaml`:

```
environments:
  {{ .Environment.Name }}:
    values:
    - environments/{{ .Environment.Name }}/values.yaml
```

`helmfile.legacy2.yaml`:

```
bases:
- base.yaml

releases:
- name: {{ .Environment.Values.releaseName }}
  values:
  - releases/{{` {{.Release.Name }} `}}/values.yaml
```

`bases` and the selective whole-file templating, we can reimplement these with:

`base.yaml.gotpl`:

```
environments:
  {{ .Environment.Name }}:
    values:
    - environments/{{ .Environment.Name }}/values.yaml
```

`helmfile.new.yaml`:

```yaml
bases:
- base.yaml.gotpl

releases:
- name: '{{ .Environment.Values.releaseName }}'
  values:
  - releases/{{.Release.Name}}/values.yaml
```

Notice `helmfile.new.yaml` is a valid YAML, but you can still use Release Template #439 in `name` and the `values` item to dynamically generate their values.

This may eventually be a breaking-change, as your `helmfile.yaml` won't work after this change.

But we can defer the breakness until helmfile v1. In the meantime, we can just required `.v1` ID in file names, like `helmfile.v1.yaml` to enable this mode.

`helmfile.v1.yaml` is indeed verbose so deprecating it in helmfile v1 and making the `helmfile.yaml` defaults to a regular YAML file would be necessary in the end.

### LINKED ISSUE #585: feat: Disable the whole-file templating of helmfile.yaml
Instead make it require `.gotmpl` extension to enable templating so that it looks like `helmfile.yaml.gotmpl` or `helmfile.gotmpl`.

As originally described in #584, the benefit of `helmfile.yaml` being not a go template is that you don't need "ugly"(some people think so) backticks to escape release template expressions #439.

And it would make helmfile slightly easier to integrate with tools like [kapitan](https://github.com/deepmind/kapitan) and `jsonnet`, that allows you to leverage richer language features to generate helmfile configs.

Note that #439 are enabled regardless of the file extensions. They're already vanilla YAML string literals containing go template expressions, that aren't handled by the whole-file templating.

Extracted from https://github.com/roboll/helmfile/issues/584#issuecomment-491177956

mumoshu:
The current  [Layering](https://github.com/roboll/helmfile/blob/master/docs/writing-helmfile.md#layering) system didn't work as documented, as it relies on helmfile to template each "part" of your helmfile.yaml THEN merge them one by one.

The reality was that helmfile template all the parts of your helmfile.yaml at once, and then merge those YAML documents. In https://github.com/roboll/helmfile/issues/388#issuecomment-436186278, @sruon was making a GREAT point that we may need to change helmfile to render templates earlier - that is to evaluate a template per each helmfile.yaml part separated by `---`. Sorry I missed my expertise to follow your great idea last year @sruon  😭 

Anyways, this, in combination with the wrong documentation, has made so many people confused. To finally overcome this situation, here's a fairly large PR that introduces the 2 enhancements:

- `bases:` for easier layering without go template expressions, especially `{{ readFunc "path/to/file" }}`s. This is the first commit of this PR.
- `helmfile.yaml` is splited by the separator `---` at first. Each part is then rendered as a go template(double-render applies as before). Finally, All the results are merged in the order of occurence. I assume this as an enhanced version of @sruon's work. This is the second commit of this PR.

Resolves #388
Resolve #584
Resolves #585 (`HELMFILE_EXPERIMENTA=true -f helmfile.yaml helmfile` disables the whole-file templating, treating the helmfile.yaml as a regular YAML file as the file ext. denotes. Use `helmfile.yaml.gotmpl` or `helmfile.gotmpl` to enable)
Fixes #568 (Use `bases` or `readFile` rather than not importing implicitly with `helmfile.d`

### DIFF
--- a/docs/writing-helmfile.md
+++ b/docs/writing-helmfile.md
@@ -100,10 +100,10 @@ Use Layering to extract the common parts into a dedicated *library helmfile*s, s
 Let's assume that your `helmfile.yaml` looks like:
 
 ```
-{ readFile "commons.yaml" }}
----
-{{ readFile "environments.yaml" }}
----
+bases:
+- commons.yaml
+- environments.yaml
+
 releases:
 - name: myapp
   chart: mychart
--- a/docs/writing-helmfile.md
+++ b/docs/writing-helmfile.md
@@ -125,23 +125,26 @@ environments:
   production:
 ```
 
-At run time, template expressions in your `helmfile.yaml` are executed:
+At run time, `bases` in your `helmfile.yaml` are evaluated to produce:
 
 ```yaml
+# commons.yaml
 releases:
 - name: metricbaet
   chart: stable/metricbeat
 ---
+# environments.yaml
 environments:
   development:
   production:
 ---
+# helmfile.yaml
 releases:
 - name: myapp
   chart: mychart
 ```
 
-Resulting YAML documents are merged in the order of occurrence,
+Finally the resulting YAML documents are merged in the order of occurrence,
 so that your `helmfile.yaml` becomes:
 
 ```yaml
@@ -159,3 +162,5 @@ releases:
 Great!
 
 Now, repeat the above steps for each your `helmfile.yaml`, so that all your helmfiles becomes DRY.
+
+Please also see [the discussion in the issue 388](https://github.com/roboll/helmfile/issues/388#issuecomment-491710348) for more advanced layering examples.
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/signal"
 	"strings"
+	"syscall"
 
 	"github.com/roboll/helmfile/helmexec"
 	"github.com/roboll/helmfile/state"
@@ -14,7 +15,6 @@ import (
 
 	"path/filepath"
 	"sort"
-	"syscall"
 )
 
 type App struct {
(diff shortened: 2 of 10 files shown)


orbiran88:
Awesome!

mumoshu:
Thanks for reviewing!

bitsofinfo:
trying to use `bases:` as documented, in 0.63, and completely does not work. Is this PR above to fix it? is there a document update to come as well?

mumoshu:
@bitsofinfo This PR includes the documentation update as well. Do you see any error messages?

PR #593, opened 2019-05-14
feat: Dependency locking

### LINKED ISSUE #483: Lock file
### Summary

In order to maintain predictable deployments, as developer I want to generate and use lock file for all chart versions retrieved from a helmfile.

### Rationale

We have an environment git repo for deployment with Helmfile. Helmfiles typically specify **version range**, e.g. "^1.0.0". Charts are developed separately. We're deploying from environment repo to "dev" automatically whenever some chart repo is updated, then to "staging" when chart repo is bumped to a stable version and then to "production" manually after "staging" is verified.

Since there's a chance that a new chart version arrives between deploying to staging and production, it is possible to unexpectedly deploy to production a version that is different from staging (i.e. untested). It is possible because of using version ranges.

Before we were using a huge helm umbrella chart for that purpose.

Therefore, it would be great if we could use helmfile to generate a "lock file" that can be pushed to git repo and then used for deploying _exact_ versions idempotently on "helmfile apply".

### Workaround

As a possible workaround, we could maintain a fake umbrella chart in a local dir with all requirements. Instead of a repo in helmfiles, reference local dir of a subchart from umbrella chart. Requires "helm dep build" and "tar -x *.tar.gz" on subcharts dir every time before deployment.

Is this use-case valid for Helmfile usage ...or can you recommend an alternative way to organize an automated CD?

Thank you in advance!

### LINKED ISSUE #563: Feature Request: Workflow to simplify upgrades
Background

We have a setup which involves several charts from the stable repo as well as self hosted repos. We would like to periodically update the charts in a convenient manner. 

Proposed Solution:

1. New flag --force-latest for apply and diff commands which would apply/diff the charts with the latest version (should ideally not be run with --skip-repo-update)
2. New subcommand freeze. Would populate the version number in helmfile.yaml/helmfile.d with what is installed in the cluster. Similar to a `pip freeze` functionality of python.

mumoshu:
In order to maintain predictable deployments, as developer I want to generate and use "lock files" for all chart versions retrieved from a helmfile.

This change solves it by (1)enhancing `helmfile deps` to generate a lock file containing all the direct chart dependencies of each helmfile state file and
(2)making other helmfile sub-commands reads the lock file and merge the locked version numbers to the helmfile state file being processed.

The lock file is named after the helmfile state file being locked, so that you can have multiple set of the helmfile state file and the lock file pairs in a directory.

When `helmfile deps` are not explicitly run before commands like `sync`, all the helmfile behavior should remain as before.

Let's say you have `helmfile.1.yaml`:

```
repositories:
- name: stable
  url: https://kubernetes-charts.storage.googleapis.com

releases:
- name: envoy
  chart: stable/envoy
- name: envoy2
  chart: stable/envoy
```

`helmfile deps` generates `helmfile.1.lock` that looks like:

```
dependencies:
- name: envoy
  repository: https://kubernetes-charts.storage.googleapis.com
  version: 1.5.0
digest: sha256:e43b05c8528ea8ef1560f4980a519719ad2a634658abde0a98daefdb83a104e9
generated: 2019-05-14T16:45:37.78205+09:00
```

Under the hood, `helmfile deps` creates a temporary local helm chart with a dummy `Chart.yaml` and `requirements.yaml` deduced from the `helmfile.yaml` content, then runs `helm dependency update` to produce and update the corresponding `requirements.lock` file.

`helmfile` then renames it to match the name of the targeted helmfile state file and moves it,  so that it becomes adjacent to each `helmfile.yaml`.

Other `helmfile` commands like `sync`, `diiff`, `apply`, `lint` read chart version numbers from the lock file.

Resolves #483
Resolves #563

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,20 @@ dependencies of any referenced local charts.
 
 For Helm 2.9+ you can use a username and password to authenticate to a remote repository.
 
+### deps
+
+The `helmfile deps` sub-command locks your helmfile state and local charts dependencies.
+
+It basically runs `helm dependency update` on your helmfile state file and all the referenced local charts, so that you get a "lock" file per each helmfile state or local chart.
+
+All the other `helmfile` sub-commands like `sync` use chart versions recorded in the lock files, so that e.g. untested chart versions won't suddenly get deployed to the production environment.
+
+For example, the lock file for a helmfile state file named `helmfile.1.yaml` will be `helmfile.1.lock`. The lock file for a local chart would be `requirements.lock`, which is the same as `helm`.
+
+It is recommended to version-control all the lock files, so that they can be used in the production deployment pipeline for extra reproducibility.
+
+To bring in chart updates systematically, it would also be a good idea to run `helmfile deps` regularly, test it, and then update the lock files in the version-control system.
+
 ### diff
 
 The `helmfile diff` sub-command executes the [helm-diff](https://github.com/databus23/helm-diff) plugin across all of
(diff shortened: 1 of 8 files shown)


mumoshu:
Let's see how this works in the wild.

PR #594, opened 2019-05-14
fix: Fetch devel

sstarcher:


mumoshu:
LGTM. Thanks for fixing!

PR #620, opened 2019-05-28
feat: optionally allow missing environment values/secrets files

### LINKED ISSUE #548: Environment values files should allow missing files

### LINKED ISSUE #559: Inconsistant config on missing file and no fail
Hi,

I just realised an ingress resources was not generated, since some values was not interpreted.
The reason was because a secret file did not exist. This happen only when using env secrets.

# How to reproduce
1. `helm create nginx`
2. create a Sops `env-secrets.yaml` file
3. Write this helmfile.yaml
```
environments:
  default:
    secrets:
      - env-secrets.yaml

releases:
  - name: myapp
    chart: nginx
    namespace: default
    secrets: [secrets.yaml]    # Notice this file does not exist
    values:
      - ingress:
          enabled: true
```
4. `helmfile diff`

# What is happening
Rendering provides one deployment and one service.

# What is expected
Helmfile should crash because the `secret.yaml` does not exist

# Workaround
Remove the `secrets: [secrets.yaml]` to fix the issue, of course :)
All deployment, service and ingress objects are rendered.

Context: helmfile version v0.54.2

mumoshu:
```yaml
environments:
  default:
    missingFileHandler: Warn
    values:
    - path/to/values.yaml
    secrets:
    - path/to/secrets.yaml
```

`missingFileHandler` set to `Warn`, `Info`, or `Debug` results in helmfile NOT stop when `path/to/values.yaml` or `path/to/secrets.yaml` is missing.

Resolves #548

While implementing the above feature, I also found a bug that has been causing #559. This also fixes that.

To verify it is actually fixed, create an example helmfile.yaml that looks like the below, and run `helmfile diff`:

```
$ cat helmfile.yaml
environments:
  default:
    secrets:
      - env-secrets.yaml

releases:
  - name: myapp
    chart: nginx
    namespace: default
    secrets: [secrets.yaml]    # Notice this file does not exist
    values:
      - ingress:
          enabled: true

$ helmfile diff
could not deduce `environment:` block, configuring only .Environment.Name. error: failed to read helmfile.yaml.part.0: environment values file matching "env-secrets.yaml" does not exist
in ./helmfile.yaml: failed to read helmfile.yaml: environment values file matching "env-secrets.yaml" does not exist
```

Fixes #559

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,10 @@ To avoid upgrades for each iteration of `helm`, the `helmfile` executable delega
 The default helmfile is `helmfile.yaml`:
 
 ```yaml
+# Chart repositories used from within this state file
+#
+# Use `helm-s3` and `helm-git` and whatever Helm Downloader plugins
+# to use repositories other than the official repository or one backend by chartmuseum.
 repositories:
   - name: roboll
     url: http://roboll.io/charts
@@ -63,6 +67,9 @@ helmDefaults:
   # path to TLS key file (default "$HELM_HOME/key.pem")
   tlsKey: "path/to/key.pem"
 
+# The desired states of Helm releases.
+#
+# Helmfile runs various helm commands to converge the current state in the live cluster to the desired state defined here.
 releases:
   # Published chart example
   - name: vault                            # name of this release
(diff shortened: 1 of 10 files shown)


sgandon:
Nice ! thanks a lot.

Issue #630, opened 2019-05-31
0.68.1: is Sprig 'values' dict function supported?  function "values" not defined


bitsofinfo:
Will provide more details in a bit:

Getting 
```
err: error during deployments.helmfile.yaml.part.2 parsing: template: stringTemplate:8: function "values" not defined
in ./deployments.helmfile.yaml: error during deployments.helmfile.yaml.part.2 parsing: template: stringTemplate:8: function "values" not defined
```

Is the sprig `values` function available? http://masterminds.github.io/sprig/dicts.html

mumoshu:
@bitsofinfo Good catch!

Seems like helmfile is currently using a version of sprig (2.15.0) that is one-minor-ver-earlier than the version added `values` (2.16.0+, https://github.com/Masterminds/sprig/commit/df393122472a0abccc1046e2cf05173b14a00535).

Bump `sprig` in `go.mod` and `values` should be available.

PR #631, opened 2019-05-31
feat: upgrade sprig to 2.16.0 for #630

bitsofinfo:
Upgrades sprig from 2.15.0 to 2.16.0 to get `values` function support for Dicts. Addresses #630

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.12
 
 require (
 	github.com/Masterminds/semver v1.4.1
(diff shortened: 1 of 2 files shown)


mumoshu:
@bitsofinfo Thanks! Would you mind running `go build . ` locally, so that it will update `go.sum` accordingly, and includes the update to `go.sum` into this PR?

mumoshu:
JFYI, try including something like `Resolves #630` in the dedicted line in your commit, on next ocassion, so that merging this PR automatically closes the issue!

PR #645, opened 2019-06-04
fix: "cannot unmarshal !!str `<no value>` into bool" errors in state templates

### LINKED ISSUE #553: v0.53.3: environment yaml files: "cannot unmarshal !!str `<no value>` into bool" typing issues?
helmfile version v0.53.3

I have an environment YAML like this

x.yaml
```
kubeContext: "mycontext"
wait: false
```

My helmfile is like this
```
environments:
  x:
    values:
      - x.yaml

helmDefaults:
  tillerNamespace: kube-system 
  kubeContext: {{ .Environment.Name }} 
  verify: false
  wait: {{ .Environment.Values.wait }} 
  timeout: 600
  recreatePods: true
  force: true
....
```

When I run `helmfile -e x -f app.yaml apply` I get 

```
could not deduce `environment:` block, configuring only .Environment.Name. error: failed to read app.yaml: reading document at index 1: yaml: unmarshal errors:
line XX: cannot unmarshal !!str `<no value>` into bool
error during app.yaml parsing: template: stringTemplate:11:30: executing "stringTemplate" at <.Environment.Values.kubeContext>: map
 has no entry for key "kubeContext"
```

mumoshu:
Seems like we are affected by https://github.com/golang/go/issues/24963. That is, even though we internally use the template option `missingkey=zero`, in some cases it still prints `<no value>` instead of zero values, which has been confusing the state yaml parsing.

This fixes the issue by naively replacing all the remaining occurrences of `<no value>` in the rendered text, while printing debug logs to ease debugging in the future when there is unexpected side-effects introduced by this native method.

Fixes #553

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/Masterminds/semver v1.4.1
 	github.com/Masterminds/sprig v2.16.0+incompatible
 	github.com/aokoli/goutils v1.0.1 // indirect
-	github.com/google/go-cmp v0.3.0 // indirect
+	github.com/google/go-cmp v0.3.0
 	github.com/google/uuid v0.0.0-20161128191214-064e2069ce9c // indirect
 	github.com/huandu/xstrings v1.0.0 // indirect
 	github.com/imdario/mergo v0.3.6
--- a/pkg/app/app_test.go
+++ b/pkg/app/app_test.go
@@ -838,7 +838,7 @@ releases:
 	}
 }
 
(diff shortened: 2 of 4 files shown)


sgandon:
have you considered moving to the yaml parser v3 that was released recently ?
https://blog.ubuntu.com/2019/04/05/api-v3-of-the-yaml-package-for-go-is-available

mumoshu:
@sgandon Yeah! But the problem behind this change has nothing to do with the underlying YAML parser used(If you've introduced yaml v3 in regard to it

PR #648, opened 2019-06-04
feat: remote state files

### LINKED ISSUE #347: Feature Request/Discussion: Remote Helmfiles?
## what

```
helmfiles:
  - https://raw.githubusercontent.com/cloudposse/helmfiles/0.6.1/helmfile.d/0020.kiam.yaml
  - https://raw.githubusercontent.com/cloudposse/helmfiles/0.7.0/helmfile.d/0100.external-dns.yaml
```

## why
* Create definitions that stitch together different versions of helmfile definitions
* Do not force upgrades of all helmfiles just because one file was updated

## use-case

We write a lot of helmfiles used by different organizations. We version pin to a release of our helmfiles distribution, however, sometimes users need to update only one specific service and are not ready to upgrade all services in a given release of helmfiles.

Using the approach above works more like `terraform` modules empowering the user to surgically import resources into their environment, while keeping it DRY and CODEOWNERS friendly by using a single source of truth. 

## references
- https://github.com/roboll/helmfile/pull/266

mumoshu:
This change enhances helmfile to accept terraform-module-like URLs in nested state files a.k.a sub-helmfiles.

```yaml
helmfiles:
- # Terraform-module-like URL for importing a remote directory and use a file in it as a nested-state file
  # The nested-state file is locally checked-out along with the remote directory containing it.
  # Therefore all the local paths in the file are resolved relative to the file
  path: git::https://github.com/cloudposse/helmfiles.git@releases/kiam.yaml?ref=0.40.0
```

The URL isn't equivalent to terraform module sources. The difference is that we use `@` to distinguish between (1) the path to the repository and directory containing the state file and (2) the path to the state file being loaded. This distinction provides us enough fleibiity to instruct helmfile to check-out necessary and sufficient directory to make the state file works.

Under the hood, it uses [hashicorp/go-getter](https://github.com/hashicorp/go-getter), that is used for [terraform module sources](https://www.terraform.io/docs/modules/sources.html) as well.

Only the git provider without authentication like git-credentials helper is tested. But theoretically any go-getter providers should work. Please feel free to test the provider of your choice and contribute documentation or instruction to use it :)

Resolves #347


osterman:
Thank you soooooo much @mumoshu ! We will give this a shot right away.

PR #655, opened 2019-06-05
fix: use the same logic as helm-secrets when decoding secrets

vbehar:
helm-secrets uses the `HELM_SECRETS_DEC_SUFFIX` env var to define the name of the output file
we should have the same logic in helmfile, to come up with the same filename

It only affects people using the `HELM_SECRETS_DEC_SUFFIX` env var

Use-case: if you want to run multiple `helmfile` commands in parallel, without conflicts. in this case, you need to decrypt secrets with different suffixes.

### DIFF
--- a/pkg/helmexec/exec.go
+++ b/pkg/helmexec/exec.go
@@ -145,10 +145,16 @@ func (helm *execer) DecryptSecret(context HelmContext, name string, flags ...str
 	}
 	defer tmpFile.Close()
 
+	// HELM_SECRETS_DEC_SUFFIX is used by the helm-secrets plugin to define the output file
+	decSuffix := os.Getenv("HELM_SECRETS_DEC_SUFFIX")
+	if len(decSuffix) == 0 {
+		decSuffix = ".yaml.dec"
+	}
+	decFilename := strings.Replace(name, ".yaml", decSuffix, 1)
+
 	// os.Rename seems to results in "cross-device link` errors in some cases
 	// Instead of moving, copy it to the destination temp file as a work-around
 	// See https://github.com/roboll/helmfile/issues/251#issuecomment-417166296f
-	decFilename := name + ".dec"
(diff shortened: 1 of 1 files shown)


pkg/helmexec/exec.go
```diff
@@ -145,10 +145,16 @@ func (helm *execer) DecryptSecret(context HelmContext, name string, flags ...str
 	}
 	defer tmpFile.Close()
 
+	// HELM_SECRETS_DEC_SUFFIX is used by the helm-secrets plugin to define the output file
+	decSuffix := os.Getenv("HELM_SECRETS_DEC_SUFFIX")
+	if len(decSuffix) == 0 {
+		decSuffix = ".yaml.dec"
+	}
+	decFilename := strings.Replace(name, ".yaml", decSuffix, 1)
```
mumoshu: This seems to break when the file ext. is something other than `.yaml`.
For example `foo.secrets.yml` would remain `foo.secrets.yml` regardless of `HELM_SECRETS_DEC_SUFFIX` which isn't what you would want:

How about trimming the trailing `.yaml`, and then adding the decSuffix?

Perhaps:

```
decFilename := strings.TrimSuffix(name, ".yaml") + decSuffix
```

vbehar: Well, in fact the helm-secrets plugin doesn't support any other extension than `.yaml` - see https://github.com/futuresimple/helm-secrets/blob/master/secrets.sh#L306

```
__ymldec=$(sed -e "s/\\.yaml$/${DEC_SUFFIX}/" <<<"$yml")
```

so anyway a `.yml` extension will never be decrypted.
and if we come up with a different naming for the decrypted file, it will just not exists.

mumoshu: Good to know! Thanks for confirming ☺️

vbehar:
thanks for the quick review & merge!

PR #660, opened 2019-06-06
Sync repos before update deps

nauxliu:


pkg/app/run.go
```diff
@@ -30,6 +30,10 @@ func (r *Run) askForConfirmation(msg string) bool {
 func (r *Run) Deps(c DepsConfigProvider) []error {
 	r.helm.SetExtraArgs(argparser.GetArgs(c.Args(), r.state)...)
 
+	if errs := r.ctx.SyncReposOnce(r.state, r.helm); errs != nil && len(errs) > 0 {
```
mumoshu: Would you mind implementing `--skip-repos` for `deps` as well, so that you can skip time-consuming `helm repo up` in advanced use-cases that the user runs `helm repos` independently and beforehand?

Perhaps it would look like in `main.go`:

```
cli.BoolFlag{
	Name:  "skip-repos",
	Usage: "skip running `helm repo update` before running `helm dependency build`",
},
```

And you would also need to add `SkipRepos() bool` to `DepsConfigProvider`.

PR #670, opened 2019-06-10
feat: `helmfile -f <go-getter url>`

mumoshu:
Extends the remote-helmfile feature to also work when loading the first state file.
This should be useful for people who wants to give helmfile a try without ever opening `$EDITOR`.


osterman:
Great idea!

PR #671, opened 2019-06-10
feat: make release version templatized

### LINKED ISSUE #669: release version is not "templatizable"
in the release definition many element can be set as templates like *chart*, *namespace*, *name*... but not the chart *version*.
see https://github.com/roboll/helmfile/blob/3710f6233e71220e35b48ebd38ff196c822f991a/pkg/state/release.go#L9
This would be very usefull if the version is to maintained in some values file.

sgandon:
Allows the release version to be a Go template.
Fixes #669

### DIFF
--- a/pkg/state/release.go
+++ b/pkg/state/release.go
@@ -2,6 +2,7 @@ package state
 
 import (
 	"fmt"
+
 	"github.com/roboll/helmfile/pkg/tmpl"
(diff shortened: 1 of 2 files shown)


sgandon:
Please do not merge, I need to make sure it answers the requirements.

sgandon:
ok I have verified it on my helmfiles, this should be fine.

PR #672, opened 2019-06-11
fix: persist original file path when using bases

yurrriq:
Prior to this change, the resulting lock file was called `<bases[0]>.lock`, instead of `<filename>.lock`.

This change ensures the final, merged state has the correct `.FilePath`.

Please let me know how I can improve/augment this one-liner. It'd be nice, for example, to have a test that prevents regressions.


mumoshu:
@yurrriq Good catch!

>  It'd be nice, for example, to have a test that prevents regressions.

We have a bunch of tests for `ParseAndLoad` in `pkg/state/create_test.go`. Probably it would make sense to add assertions on the `FilePath` of the returned `HelmState` there?

yurrriq:
I found a good place to add an assertion that fails prior to, and succeeds after the change. See CircleCI build history for proof. 😄

yurrriq:
Thank you for merging, and for your work this project!

PR #693, opened 2019-06-14
feat: add --output-dir on template command

### LINKED ISSUE #629: Add  --output-dir to the helmfile template command
For cases when helmfile is used only for templating, will be great to support `--output-dir` as original helm does.

```
$ helm template --help
...
      --output-dir string        writes the executed templates to files in output-dir instead of stdout
...
```
Some benefits:
* it renders yamls to corresponding files, preserving structure of the original chart
* it ignores empty files [helm/helm/issues/3979]

olivierboudet:
Resolves #629

### DIFF
--- a/pkg/helmexec/exec.go
+++ b/pkg/helmexec/exec.go
@@ -44,14 +44,12 @@ func NewLogger(writer io.Writer, logLevel string) *zap.SugaredLogger {
 }
 
 // New for running helm commands
-func New(logger *zap.SugaredLogger, kubeContext string) *execer {
+func New(logger *zap.SugaredLogger, kubeContext string, runner Runner) *execer {
 	return &execer{
 		helmBinary:  command,
 		logger:      logger,
 		kubeContext: kubeContext,
-		runner: &ShellRunner{
-			logger: logger,
-		},
+		runner:      runner,
 	}
 }
 
--- a/main.go
+++ b/main.go
@@ -200,6 +200,10 @@ func main() {
 					Name:  "values",
 					Usage: "additional value files to be merged into the command",
 				},
+				cli.StringFlag{
+					Name:  "output-dir",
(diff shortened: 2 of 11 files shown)


olivierboudet:
Hello @mumoshu ,

I have added a unit test. Can you check if it looks OK ?

pkg/helmexec/exec.go
```diff
@@ -17,7 +17,7 @@ const (
 	command = "helm"
 )
 
-type execer struct {
+type Execer struct {
```
mumoshu: I think you can safely keep this private within the helmexec pkg, as your code is well decoupled by using [`helmexec.Interface`](https://github.com/roboll/helmfile/pull/693/files#diff-637a60a871f8b69622bdfd3d3b2fcad1R49) and there's no direct calls to `Execer{}` outside of `helmexec`.

So, would you mind if I changed the struct name back to `execer` later?

// I'd appreciate it if you could correct me if I'm missing something and this needs to be exposed to other pkgs :)

olivierboudet: I passed it public to allow create a MockExecer in app_test.go (but I may have missed someting to do better because it my first golang code :) ) : 
```
func MockExecer(logger *zap.SugaredLogger, kubeContext string) *helmexec.Execer {
	execer := helmexec.New(logger, kubeContext, &mockRunner{})
	return execer
}
```
but I am agree that it does not need to be exposed to other package if we can do another way in tests.

mumoshu: Would it work if you made the return type `helmexec.Interface`, like `func MockExecer(logger *zap.SugaredLogger, kubeContext string) helmexec.Interface`?

olivierboudet: you are right, it's fixed

mumoshu: Perfect! Thanks again for your support ☺️

PR #694, opened 2019-06-14
feat: Disable concurrency if any release is using tillerless

nauxliu:


nullck:
@nauxliu   great!!
Thank you!
I'm using tillerless flag, and I always need to remember to set ```--concurrency=1```

nauxliu:
> JFYI you can write it as `if (*r.Tillerless != nil ** r.Tillerless) || st.HelmDefaults.Tillerless {` without fearing panic'ing, but
> 
> LGTM! Feeling awesome to see a small amount of code can improve the U/X that much 🎉

Do you mean `if (r.Tillerless != nil && *r.Tillerless) || st.HelmDefaults.Tillerless` ?
This won't work as expected if `*r.Tillerless` if set to false,

mumoshu:
@nauxliu Yeah that's what I meant! Still curious - how it won't work as expected when it's set to false?

My suggestion was based on my assumption that `r.Tillerless` is considered `false` when `nil`(cuz we treat `tillerless: false` by default i.e. omitted). So it's safe to check for `r.Tillerless` to be true only when `r.Tillerless != nil`. But I may be missing something.

nauxliu:
Because explicit set release's tillerless to false will fallback to the default value.

For example:

```yaml
helmDefaults:
  tillerless: true

releases:
  - name: foo
    tillerless: false
```

```go
if (r.Tillerless != nil && *r.Tillerless) || st.HelmDefaults.Tillerless){ 
    concurrency = 1
} 
```

mumoshu:
@nauxliu Ahh, you're absolutely correct! Thanks for pointing it out ☺️

PR #722, opened 2019-06-25
fix: enabled '=' in a value for --state-values-set

kalioz:
This adds the possibility to use a "=" in the parameters passed to --state-values-set

previous comportment:
`helmfile --state-values-set 'test="abcde==fg",test2="abcde"'` => outputs 'test="abcde",test2="abcde"'

new comportment:
`helmfile --state-values-set 'test="abcde==fg",test2="abcde"'` => outputs 'test="abcde==fg",test2="abcde"'


kalioz:
[issue 727](https://github.com/roboll/helmfile/issues/727)

Issue #704, opened 2019-06-19
Unable to matches specified selector() and environment(master), in any helmfile when using Helmfiles


jjmartres:
I'm trying to make my helmfiles DRY

helmfile.yaml
```YAML
helmfiles:
- apps/99_monitor.yaml
```
apps/99_monitor.yaml
```YAML
bases:
- ../common/environments.yaml
- ../common/repositories/kubernetes.yaml
- ../common/helmdefaults.yaml

{{ readFile "../common/templates.yaml" }}

releases:
  - name: gitlab-{{ .Environment.Name }}-branch-fluent-bit
    chart: stable/fluent-bit
    version: ~1.8.0
    values:
      - "../environments/{{ `{{ .Environment.Name }}` }}/gitlab-{{ `{{ .Environment.Name }}` }}-branch-fluent-bit.yaml"
    <<: *default
```
If I lint charts, using `apps/99_monitor.yaml` I've no errors:

```bash
$ helmfile -e master --log-level debug -f apps/99_monitor.yaml lint                                                      
processing file "99_monitor.yaml" in directory "apps"
changing working directory to "/home/jjmartres/Desktop/deploy/apps"
first-pass rendering starting for "99_monitor.yaml.part.0": inherited=&{master map[] map[]}, overrode=<nil>
first-pass uses: &{master map[] map[]}
first-pass produced: &{master map[] map[]}
first-pass rendering result of "99_monitor.yaml.part.0": {master map[] map[]}
vals:
map[]
defaultVals:[]
second-pass rendering result of "99_monitor.yaml.part.0":
 0: environments:
 1:   master:
 2:   review:
 3: 

merged environment: &{master map[] map[]}
first-pass rendering starting for "99_monitor.yaml.part.1": inherited=&{master map[] map[]}, overrode=<nil>
first-pass uses: &{master map[] map[]}
first-pass produced: &{master map[] map[]}
first-pass rendering result of "99_monitor.yaml.part.1": {master map[] map[]}
vals:
map[]
defaultVals:[]
second-pass rendering result of "99_monitor.yaml.part.1":
 0: repositories:
 1:   - name: kubernetes-charts
 2:     url: https://kubernetes-charts.storage.googleapis.com/
 3: 

merged environment: &{master map[] map[]}
first-pass rendering starting for "99_monitor.yaml.part.2": inherited=&{master map[] map[]}, overrode=<nil>
first-pass uses: &{master map[] map[]}
[...]

PR #741, opened 2019-07-04
fix: incorrect error message when sub-helmfiles-only state missing env

mumoshu:
This is a revised version of #706, which doesn't make CI angry and a more straight-forward(yet smelling) implementation

Ref https://github.com/roboll/helmfile/issues/704#issuecomment-503647741


mumoshu:
Opps, wrong title and branch

Issue #782, opened 2019-07-31
feat: parallel / de-duplicated secrets decryption 


travisgroth:
Hi,

On larger installs leveraging the helm secrets plugin, a full run of lint, diff and apply can take a long time (I just watched a 3+ minute run).  After digging into the code it seems like there's two reasons:

1. A mutex to avoid two goroutines erasing each other's .dec.yaml files when moving to temp storage.  This mutex forces all decrypts to be serial.

2. No coordination of which secrets have already been decrypted and moved to temp storage.  A secrets file is decrypted N times, where N is the number of charts referencing it.

Some implementation ideas:

1.  Allow decrypt to run free unless two routines are operating on the same file.  

    a.  Use filesystem flock() to shift the locking out of process.  This is likely simpler, though may have platform corner cases.

    b.  Shift the mutex into some sort of shared map indexed on some representation of a file path, so that locking can occur against file level granularity.

2.  Be able to detect if a secret has already been decrypted and skip it (this also helps item [1]).  Like [1], there is a way both on-filesystem and in-memory.  

    a.  Hash the secret file path/name into a unique and secure value.  Use this value as the file name inside the temporary location.  Before calling decrypt, check if a file is already present.  If it is, the decrypt can be skipped.  If we try our new locking before checking for the tempfile, we can avoid any duplicate decrypts.

    b.  Using a structure much like [1b], track the files that are already decrypted into temp storage and check if they are already decrypted.

Thoughts?  This may have implications in #745

mumoshu:
@travisgroth Your proposal sounds great! Not sure I can tackle this shortly but it would definitely welcome more ideas and pull requests related to this.

Also, I was thinking that caching decrypted secrets #444 will also help.

travisgroth:
Question re #444 - I think [2] addresses that without putting them in-memory, which sounds objectively better, but more complicated.  

I can probably take a pass at this in the next week or so.

mumoshu:
> Question re #444 - I think [2] addresses that without putting them in-memory, which sounds objectively better, but more complicated.

I thought #444 can be better in terms of that it doesn't need to keep all the cached credentials on tmpfs until helmfile exits.

Anyway, I believe your #444 is an extension of your [2] as either way we need to decrypt the secret onto the disk(because that's how helm-secrets works) before saving the result in memory(#444).

travisgroth:
After looking a bit, I think you're right.  My conclusion was that the cleanup gets complicated if we cache on disk with the current design.

[...]

Issue #444, opened 2019-01-22
[feature] cache environment secret decryption


sstarcher:
After the recent addition of templates I went through a refactor and broke apart some of my massive helmfiles.  This has been a great advantage with usability and readability, but now if targeting a specific item we have a ton of verbose and time intensive spam of secret decryption.  

Can we do what we did with the repos and cache the secrets in memory as they are accessed so if you have 20 helmfiles we don't re-decrypt the environment for each of them.

```
NOTES:
The Prometheus Operator has been installed. Check its status by running:
  kubectl --namespace ops get pods -l "release=prometheus"

Visit https://github.com/coreos/prometheus-operator for instructions on how
to create & configure Alertmanager and Prometheus instances using the Operator.

Decrypting secret /Users/sstarcher/kubernetes/helm/account/dev/_environment/secrets.yaml
Decrypting /Users/sstarcher/kubernetes/helm/account/dev/_environment/secrets.yaml

Decrypting secret /Users/sstarcher/kubernetes/helm/account/dev/_environment/secrets.yaml
Decrypting /Users/sstarcher/kubernetes/helm/account/dev/_environment/secrets.yaml

Decrypting secret /Users/sstarcher/kubernetes/helm/account/dev/_environment/secrets.yaml
Decrypting /Users/sstarcher/kubernetes/helm/account/dev/_environment/secrets.yaml

Decrypting secret /Users/sstarcher/kubernetes/helm/account/dev/_environment/secrets.yaml
Decrypting /Users/sstarcher/kubernetes/helm/account/dev/_environment/secrets.yaml

Decrypting secret /Users/sstarcher/kubernetes/helm/account/dev/_environment/secrets.yaml
Decrypting /Users/sstarcher/kubernetes/helm/account/dev/_environment/secrets.yaml

Decrypting secret /Users/sstarcher/kubernetes/helm/account/dev/_environment/secrets.yaml
Decrypting /Users/sstarcher/kubernetes/helm/account/dev/_environment/secrets.yaml

Decrypting secret /Users/sstarcher/kubernetes/helm/account/dev/_environment/secrets.yaml
Decrypting /Users/sstarcher/kubernetes/helm/account/dev/_environment/secrets.yaml

Decrypting secret /Users/sstarcher/kubernetes/helm/account/dev/_environment/secrets.yaml
Decrypting /Users/sstarcher/kubernetes/helm/account/dev/_environment/secrets.yaml

[...]

PR #790, opened 2019-08-06
Cache secrets and concurrent decryption

travisgroth:
Related to #782 and #444 

- Allows concurrent decryption of different secrets files
- Caches decrypted secrets by original file path and returns decrypted results from memory
- Secrets being run through an instance of helmexec will be cached and run as fast as possible concurrently

NB: This particular PR doesn't make _all_ calls to secrets cached and concurrent.  Environment Secrets in particular seem to not be evaluated with a ScatterGather(), and doesn't use the same helmexec instance as other parts of the code, so it doesn't take advantage of these changes.  Some reworking of the plumbing there would be needed.

### DIFF
--- a/pkg/helmexec/exec.go
+++ b/pkg/helmexec/exec.go
@@ -17,13 +17,19 @@ const (
 	command = "helm"
 )
 
+type decryptedSecret struct {
+	mutex sync.RWMutex
+	bytes []byte
+}
+
 type execer struct {
-	helmBinary      string
-	runner          Runner
-	logger          *zap.SugaredLogger
-	kubeContext     string
-	extra           []string
-	decryptionMutex sync.Mutex
+	helmBinary           string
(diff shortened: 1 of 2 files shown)


mumoshu:
I think we had a hacky workaround to use helm-secrets for decrypting environment secrets. Perhaps that's one of gotchas? Anyways, I'm fine with just sharing helmexec within and outside of scatterGather, if that works.

travisgroth:
Thanks!  Also, your read on the methodology is accurate.  Should be least possible lock contention.  

>I think we had a hacky workaround to use helm-secrets for decrypting environment secrets. Perhaps that's one of gotchas? Anyways, I'm fine with just sharing helmexec within and outside of scatterGather, if that works.

👍 i'm still poking around.  I plan on doing some follow up PRs.

Issue #792, opened 2019-08-07
feat: Add support for `kubectl` and `jq` inside docker image


naphta:
I'm not really sure how to frame this (and I know I can just source from the current Dockerfile and build my own) but I was thinking it would be very useful for me at least and I expect others if there was a little bit more features added into the docker image. 

If adding it into the base Dockerfile isn't okay; what about adding a slim / full variant of the images which have these extra features?

I'm happy to open a PR with the changes I just thought I'd run it through you guys before just doing it.

Issue #789, opened 2019-08-05
Cannot install istio 1.2.2 from helm chart with helmfile


mojochao:
I am trying to install the istio-1.2.2 `istio` helm chart in `istio-1.2.2/install/kubernetes/helm/istio`.  I have copied both it and the `istio-init` chart to the root directory of my helmfile repo.

The relevant portion of my `helmfile.yaml` is:
```
  - chart: ./istio-init
    name: istio-init
    namespace: istio-system
    values:
      - certmanager:
          enabled: false

  - chart: ./istio
    name: istio
    namespace: istio-system
    values:
      - certmanager:
          enabled: false
        grafana:
          enabled: false
        prometheus:
          enabled: false
        tracing:
          enabled: false
        kiali:
          enabled: false
```
When I run `helmfile apply` with just the `istio-init` chart in my `helmfile.yaml` file, that chart installs succesfully.  When I run `helmfile apply` with the `istio` chart added back to my `helmfile.yaml` file, I see the following error:
```
$ helmfile apply
Building dependency istio-init
No requirements found in istio-init/charts.

Building dependency istio
in ./helmfile.yaml: helm exited with status 1:
  Error: no 'repository' field specified for dependency: "sidecarInjectorWebhook"
```
I can succesfully install it with `helm install`, however.
```
$ helm install ./istio --name istio --namespace istio-system --set certmanager.enabled=false --set grafana.enabled=false --set prometheus.enabled=false --set tracing.enabled=false --set kiali.enabled=false
<lots of output elided>
NOTES:
Thank you for installing istio.

Your release is named istio.

To get started running application with Istio, execute the following steps:
1. Label namespace that application object will be deployed to by the following command (take default namespace as an example)

$ kubectl label namespace default istio-injection=enabled
$ kubectl get namespace -L istio-injection

2. Deploy your applications

$ kubectl apply -f <your-application>.yaml

For more information on running Istio, visit:
https://istio.io/
```
What is the correct way to install the `istio` chart with `helmfile`?  Is this a bug or misconfiguration on my part?

I am using the following versions running on macOS Mojave (10.14.6) against an AWS EKS 1.13 cluster:
```
$ helm version
Client: &version.Version{SemVer:"v2.14.3", GitCommit:"0e7f3b6637f7af8fcfddb3d2941fcc7cbebb0085", GitTreeState:"clean"}
[...]

PR #799, opened 2019-08-08
Add kubectl and jq to Dockerfile

naphta:
As referenced in issue: https://github.com/roboll/helmfile/issues/789 and https://github.com/roboll/helmfile/issues/792.

For now I've made it contact kubernetes for the latest version every time, looking at the implementation of the helm installation would you prefer for it to be a hardcoded and checksum confirmed version added?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,10 @@ RUN wget ${HELM_LOCATION}/${HELM_FILENAME} && \
 # using the install documentation found at https://kubernetes.io/docs/tasks/tools/install-kubectl/
 # for now but in a future version of alpine (in the testing version at the time of writing)
 # we should be able to install using apk add.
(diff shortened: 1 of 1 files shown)


mumoshu:
Thanks! Checksum verification would be appreciated if possible.

Also, I thought version skew supported by kubectl was plus and minus 1 K8s minor vers. That is, to correctly support K8s 1.15(latest) and 1.13(oldest, maintained minor ver), shouldn't we use kubectl v1.14.x?

naphta:
Yep 1 forward/back seems to be the version skew. I'll move it to be hardcoded at the latest 1.14.x minor version.

https://github.com/kubernetes/kubernetes/issues/57748

naphta:
@mumoshu Just wondering if there's plans to run a release soon (looking to see the docker image built!)

PR #800, opened 2019-08-09
Handle environment secrets concurrently

travisgroth:
Related to #782 - Process the environment secrets in parallel.

### DIFF
--- a/pkg/state/create.go
+++ b/pkg/state/create.go
@@ -4,14 +4,15 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
+	"os"
+
 	"github.com/imdario/mergo"
 	"github.com/roboll/helmfile/pkg/environment"
 	"github.com/roboll/helmfile/pkg/helmexec"
 	"github.com/roboll/helmfile/pkg/maputil"
 	"go.uber.org/zap"
 	"gopkg.in/yaml.v2"
-	"io"
-	"os"
 )
 
 type StateLoadError struct {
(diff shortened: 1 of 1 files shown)


travisgroth:
@mumoshu I'm a bit lost on the best way to get pass workerLimit/Concurrency all the way down the stack to where we load the secrets.  I'd welcome a suggestion on how to handle that; right now, it's hard coded to 0.  I don't think this is problematic in practice, but obviously we should respect the limits everywhere.

Otherwise, I think this is ready.

mumoshu:
@travisgroth Thanks! Regarding the hard-coded concurrency - That sounds ok.

The purpose of --concurrency/worker-limit for me is to limit how much cluster resources are eventually consumed by running Helmfile. Decrypting env secrets doesn't affect the cluster so I have no specific reason to make concurrency configurable here.

travisgroth:
👍

pkg/state/create.go
```diff
@@ -201,59 +202,100 @@ func (st *HelmState) loadEnvValues(name string, ctxEnv *environment.Environment,
...
+			close(secrets)
+		},
+		func(id int) {
+			for path := range secrets {
 				release := &ReleaseSpec{}
 				flags := st.appendConnectionFlags([]string{}, release)
 				decFile, err := helm.DecryptSecret(st.createHelmContext(release, 0), path, flags...)
 				if err != nil {
-					return nil, err
+					results <- secretResult{nil, err, path}
```
mumoshu: I think we need `return` after L256, L260, L264 and L272 to not submit redundant results in L274.

travisgroth: Good catch.  With the way this was structured, it should be a continue to catch all the errors in a given run.  I think that also keeps it from blocking when gathering the results, so we have some sort of result from every task.  It should also reduce test/error cycles when using helmfile.

Pushed an update for that, and fixing an extra ref for the envVals map.

mumoshu: Ah that makes sense! `so we have some sort of result from every task` was the critical part of it I had forgot at the time of writing the comment. LGTM.

travisgroth:
👍

PR #819, opened 2019-08-23
Helm args are missing on diff when helmfile apply is called

hyunysmile:
This adds helm args to helm-diff when helmfile apply is called by:
- Changing position of helm.SetExtraArgs to before Diff

### DIFF
--- a/pkg/app/run.go
+++ b/pkg/app/run.go
@@ -2,10 +2,11 @@ package app
 
 import (
 	"fmt"
+	"strings"
+
 	"github.com/roboll/helmfile/pkg/argparser"
 	"github.com/roboll/helmfile/pkg/helmexec"
 	"github.com/roboll/helmfile/pkg/state"
-	"strings"
 )
 
 type Run struct {
@@ -136,6 +137,8 @@ func (r *Run) Apply(c ApplyConfigProvider) []error {
 	// helm must be 2.11+ and helm-diff should be provided `--detailed-exitcode` in order for `helmfile apply` to work properly
 	detailedExitCode := true
 
+	r.helm.SetExtraArgs(argparser.GetArgs(c.Args(), r.state)...)
+
 	releases, errs := st.DiffReleases(helm, c.Values(), c.Concurrency(), detailedExitCode, c.SuppressSecrets(), false)
 
 	releasesToBeDeleted, err := st.DetectReleasesToBeDeleted(helm)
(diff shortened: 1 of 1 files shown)


hyunysmile:
> Understood. Then you can do this:
> 
> ```
> helmDefaults:
>   tillerNamespace: kube-system
>   wait: true
>   timeout: 300
>   force: true
>   kubeContext: test
> 
> releases:
>   - name: kci-priorityclass
>     chart: ../charts/kci-priorityclass
>     devel: true
>     values:
>       - ../general/kci-priorityclass/values.yaml
>       
>   - name: nginx-ingress-production
>     namespace: production-general
>     chart: gstable/nginx-ingress
>     version: 1.8.1
>     values:
>       - ../general/nginx-ingress/values.yaml
>       - ../general/nginx-ingress/production.yaml
> ```
> 
> Or
> 
> ```
> helmDefaults:
>   tillerNamespace: kube-system
>   wait: true
>   timeout: 300
>   force: true
>   kubeContext: test
>   devel: true
> 
> releases:
>   - name: kci-priorityclass
>     chart: ../charts/kci-priorityclass
>     values:
>       - ../general/kci-priorityclass/values.yaml
>       
>   - name: nginx-ingress-production
>     namespace: production-general
>     chart: gstable/nginx-ingress
>     version: 1.8.1
>     values:
>       - ../general/nginx-ingress/values.yaml
>       - ../general/nginx-ingress/production.yaml
> ```

Ok!
The second solution would work for us :)
So the current problem when `helmfile apply` is executed, 
it drops --devel arg will be solved?

mumoshu:
@hyunysmile It depends. Do you still need it?

hyunysmile:
> @hyunysmile It depends. Do you still need it? And if so, how will you use it? At least in your use-case `--devel` seems not useful at all 🤔

You mean, setting `devel: true` will solve my problem with current helmfile version? is it different with using args?

mumoshu:
@hyunysmile 

> setting `devel: true` will solve my problem with current helmfile version?

Yes, that's what I mean!

Do still need a way to set `devel` from a helmfile command-line flag?

hyunysmile:
> @hyunysmile
> 
> > setting `devel: true` will solve my problem with current helmfile version?
> 
> Yes, that's what I mean!
> 
> Do still need a way to set `devel` from a helmfile command-line flag?

Sorry for the late reply. With the way you suggest, it works. :)
I will close this pull request. Thank you.

Issue #818, opened 2019-08-21
Using `installed` inside templates


astorath:
Hi!

Just can't figure out how to use `installed` key inside `template` context:

1. This way I get invalid yaml:
```yaml
templates:
  service: &service
    installed: {{`{{ true | ternary "yes" "no" }}`}}

releases:
  - name: test
    chart: stable/nginx-ingress
    namespace: default
    <<: *service
```

output:
```
processing file "test.yaml" in directory "."
first-pass rendering starting for "test.yaml.part.0": inherited=&{default map[] map[]}, overrode=<nil>
first-pass uses: &{default map[] map[]}
could not deduce `environment:` block, configuring only .Environment.Name. error: failed to read test.yaml.part.0: reading document at index 1: yaml: unmarshal errors:
  line 3: cannot unmarshal !!map into bool
  line 3: cannot unmarshal !!map into bool
error in first-pass rendering: result of "test.yaml.part.0":
 0: templates:
 1:   service: &service
 2:     installed: {{ true | ternary "yes" "no" }}
 3:
 4: releases:
 5:   - name: test
 6:     chart: stable/nginx-ingress
 7:     namespace: default
 8:     <<: *service
 9:

first-pass produced: &{default map[] map[]}
first-pass rendering result of "test.yaml.part.0": {default map[] map[]}
second-pass rendering result of "test.yaml.part.0":
 0: templates:
 1:   service: &service
 2:     installed: {{ true | ternary "yes" "no" }}
 3:
 4: releases:
 5:   - name: test
 6:     chart: stable/nginx-ingress
 7:     namespace: default
 8:     <<: *service
 9:

err: failed to read test.yaml: reading document at index 1: yaml: unmarshal errors:
  line 3: cannot unmarshal !!map into bool
  line 3: cannot unmarshal !!map into bool
in ./test.yaml: failed to read test.yaml: reading document at index 1: yaml: unmarshal errors:
  line 3: cannot unmarshal !!map into bool
  line 3: cannot unmarshal !!map into bool
```

2. this way, something happens inside helmfile serialization:
```yaml
templates:
  service: &service
    installed: "{{`{{ true | ternary "yes" "no" }}`}}"

releases:
  - name: test
    chart: stable/nginx-ingress
    namespace: default
    <<: *service
```

output:
```
processing file "test.yaml" in directory "."
first-pass rendering starting for "test.yaml.part.0": inherited=&{default map[] map[]}, overrode=<nil>
[...]

Issue #780, opened 2019-07-30
feat: `helmfile build` to produce rendered and flattened state file for GitOps


mumoshu:
## Problem

Helmfile's advanced templating features consisting of "release templates", "sub-helmfiles", "state values", "environments" are all GREAT for making your desired states DRY by avoiding repeated configs across environments.

But advanced templating has downside - less predictable results. This makes it very hard to debug and audit deployments made by Helmfile.

### Helmfile Diff

`helmfile diff` helps by previewing changes to be made to the cluster before deploying. But you need real cluster connectivity from your CI to run diff, which is a security concern. And sometimes, you still need another level of visibility to "intermediate" result produced by Helmfile - rendered helmfile.yaml template.

### Debug logs, Dry-Run, Captured Releases, ...

People usually use `--helmfile --log-level=debug` for that purpose. `helmfile --dry-run` #118 is being discussed for additional visibility and better U/X.

@bitsofinfo has proposed ability to write rendered releases to the local disk in #752. This sounds like a good idea for visibility and auditability. But one gotcha is that it may lack correctness due to how Helmfile is implemented today. Helmfile as of today applies rendered releases in-place. So if Helmfile is enhanced to write a slightly different representation of releases to disk, it's theoretically possible that what Helmfile wrote and what Helmfile applied differs, which makes the whole purpose of the feature meaningless.

### GitOps

GitOps is recently emerging for better audited and reproducible deployments. But it doesn't really help Helmfile.

Highly templatized `helmfile.yaml` commited into Git for GitOps doesn't automatically provide useful git-diff - because the diff isn't based on rendered helmfile.yaml. You need either CI or your brain to render the helmfile.yaml template and take useful diff to give confidence that a new helmfile deployment doesn't break your system, which is hard.

## Solution

I'm proposing the combination of the followings:

- Embrace GitOps
- GitOps-friendly Helmfile "ReleaseSet" API
- `helmfile build` that writes a ReleaseSet
- Alternative and internal RelesetSet processor(In-place processor we have today remains for backward-compatibility

A ReleaseSet is produced from the ordinary helmfile.yaml template you all have today and looks like:

```yaml
apiVersion: helmfile.example.com/v1alpha1
kind: ReleaseSet
metadata:
  name: myapp
  annotations:
    approved: "false"
spec:
  releases:
  # or ns1/baz meaning both namespace and tillerNamespace set to ns1
  - name: bar
    chart: mychart
    namespace: ns1
[...]

PR #823, opened 2019-08-26
feat: Advanced Templating

astorath:
1. Added `helmfile build` command to print final state
Motivation: useful for debugging purposes and some CI scenarios
2. Template interpolation is now recursive (you can cross-reference release fields) like:
```yaml
templates:
  release:
    name: {{`app-{{ .Release.Namespace }}`}}
    namespace: {{`{{ .Release.Labels.ns }}`}}
    labels:
      ns: dev
```
3. Experimental: Added some boolean release fields interpolation in templates:
```yaml
templates:
  release:
    name: {{`app-{{ .Release.Namespace }}`}}
    namespace: dev
    installedTemplate: {{`{{ eq .Release.Namespace "dev" }}`}}
```
this solves: #818

4. Added more template interpolations: Labels, SetValues
5. Added template interpolation for inline Values
6. Added `helmfile list` command to print target releases in simple tabular form
7. Added release names in some `helm` output messages, e.g.: `Comparing release=%v, chart=%v`

## TODO:

- [x] add some docs
- [x] add values (full) templating
- [x] make some `yaml` serialization improvements


mumoshu:
@astorath Hey! Thanks a lot for your efforts.

I'm looking forward to review this fully once this gets out of the WIP status.

Until then I'll keep adding design comments. Firstly:

> Template interpolation is now recursive (you can cross-reference release fields) like:

How the dependencies across field templates are handled? Do you by any chance build something like a DAG of field templates?

> Experimental: Added some boolean release fields interpolation in templates:

This makes sense. An alternative way to that is to give boolean field `interface{}` types and render it when and only when the actual value type was `string`. Have you considered it? If so, would u mind sharing why you prefer doing it with another `-Template` fields?

> Added helmfile state command to print final state
> Motivation: useful for debugging purposes and some CI scenarios

Awesome!

Can I later enhance it so that the output of the `state` command can be fed into `helmfile apply` for reproducible and audited deployments?

Actually I'm talking about #780 so I'd be glad if I can rename it to `helmfile build`. Thoughts?

astorath:
@mumoshu

> How the dependencies across field templates are handled? Do you by any chance build something like a DAG of field templates?

No, just simple iterations with some sane limit. If release stops mutating: interpolation stops.

> This makes sense. An alternative way to that is to give boolean field interface{} types and render it when and only when the actual value type was string. Have you considered it? If so, would u mind sharing why you prefer doing it with another -Template fields?

Well, I do consider my approach ugly. But I don't think that messing with `HelmSpec`  (changing some field types) is a good option: all consumers would have to cast those or use some getters on public field...

As I've pointed out, the best way (IMHO) is to separate `HelmSpec` and `HelmTemplateSpec`. Maybe treat `HelmTemplateSpec` as `map[string]interface{}`. But this approach requires a lot of effort and my humble golang knowledge is not enough for such an endeavour...

Nevertheless, if you think using `interface{}` for those fields with some getter methods is a better option - this is definitely an option.

> Actually I'm talking about #780 so I'd be glad if I can rename it to helmfile build. Thoughts?

I've replaced the name.

pkg/state/state_exec_tmpl_test.go
```diff
@@ -1,36 +1,131 @@
...
+func boolPtrToString(ptr *bool) string {
+	if ptr == nil {
+		return "<nil>"
+	}
+	return fmt.Sprintf("&%t", *ptr)
+}
+
+func ptr(v interface{}) interface{} {
+	r := v
+	return reflect.ValueOf(r).Addr().Interface()
```
mumoshu: This looks neat!

mumoshu:
> No, just simple iterations with some sane limit. If release stops mutating: interpolation stops.

Understood. As long as someone tries 6+ iterations in a real world project(not sure someone ever does!), we can live with that, which makes me think that it's a great starting point.

> the best way (IMHO) is to separate HelmSpec and HelmTemplateSpec

I agree a lot!

> Maybe treat HelmTemplateSpec as map[string]interface{}

This may negate my previous comment but I started thinking that both `map[string]interface{}` and overloading makes it harder for users to enable/disable templating on the specific field(name, namespace, tillerless and so on there will be no prob. I'm curious about `values` and `set`)

 The way you've done things with the `-Template` suffix would never suffer from that, which is good.

Probably I'll adapt existing fields like `values` and `set` to be more specific about templating enablement in the future.

docs/writing-helmfile.md
```diff
@@ -89,6 +89,42 @@ releases:
...
+  ```yaml
+  # ...
+    installedTemplate: '{{`{{ eq .Release.Namespace "kube-system" }}`}}'
+    waitTemplate: '{{`{{ eq .Release.Labels.tag "safe" | not }}`}}'
+  # ...
+  ```
+- `set` block values:
+  ```yaml
+  # ...
+    set:
```
mumoshu: Accordingly to https://github.com/roboll/helmfile/pull/823#issuecomment-526800928, I'd be happy if we could enable templates only for `setTemplate` and `valuesTemplate` rather than `set` and `values` respectively.

I'll make the changes on my own before releasing this, but if you made it on your side I'd be happy to review/accept the PR! Thanks.

mumoshu: See #833

mumoshu: Merged #833. Please feel free to leave any comments about it!

PR #827, opened 2019-08-28
allow empty pattern matching and move on

### LINKED ISSUE #778: deps fails on missing globbed files
We have a root helmfile similar to 
```
helmfiles:
- helm/account/helmfile.yaml
- helm/account/{{ .Environment.Name }}/services/*/helmfile.yaml
- helm/account/{{ .Environment.Name }}/teams/*/helmfile.yaml
```

In some environments some of the above may not exist.  This works fine with helmfile sync, diff, apply etc, but fails for helmfile deps.

```
Running helmfile from helm/account/prod
in ./helmfile.yaml: no file matching helm/account/prod/teams/*/helmfile.yaml found
```

sstarcher:
Fixes #778

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -1330,7 +1330,7 @@ func (st *HelmState) ExpandedHelmfiles() ([]SubHelmfileSpec, error) {
 			return nil, err
 		}
 		if len(matches) == 0 {
-			return nil, fmt.Errorf("no file matching %s found", hf.Path)
+			continue
 		}
 		for _, match := range matches {
 			newHelmfile := hf


sstarcher:
@mumoshu anything I can do to help with this.  It was a regression between 0.6x and 0.8x

mumoshu:
@sstarcher Thanks for reminding me! The only thing I'm struggling is about why this affects  `deps` only. I thought `ExpandedHelmfiles` in concern is called from every helmfile command but I could be wrong.

sstarcher:
It happens to all of them not just dep.

mumoshu:
Ah okay probably I've misread your comment somewhere!

Issue #836, opened 2019-09-03
Layering "values" section in release templates


a-hat:
It seems that it is not possible to have multiple `values` sections defined. I would have expected helmfile to merge the values, but it takes only the last section defined.

For example, having this helmfile:

```
templates:
  default: &default
    values:
    - defaults.yaml    
  env1: &env1
    <<: *default
    values: 
    - env1.yaml

releases:
- name: demo
  <<: *env1
  values:
  - demo.yaml
```

This will only apply `demo.yaml`, but it would have been nice if also `defaults.yaml` and `env1.yaml` would be applied (in this order).

If I am not missing something, and this is a valid feature request, I would be willing to have a go at implementing the feature.

mumoshu:
@a-hat Hey! Thanks for writing this up. I believe this is a valid feature request that should ideally be addressed asap.

What's blocking me from doing so is that I'm not yet sure how we can differentiate when to merge and not to merge `values` and how the merge to be done.

For example, in your use-case `values` should be merged, but some existing helmfile users won't like that behavior if forced. So we may need an option to toggle merging.

Regarding how the merge is done - some people want arrays nested within values to be appended some are not(=want them overrode). We may need an option for that as well.

The prior art in this area is [spruce](https://github.com/geofffranks/spruce). K8s "strategic-merge patch" can be leveraged if it's reusable as a go library.

Do you have any suggestion/idea on the two points?

a-hat:
Thanks @mumoshu for your feedback.

That is indeed a good point, sometimes it might be necessary to overwrite the `values` instead of merging. I think the right place to decide whether to replace or overwrite the values from the previous layer is when defining a `values` block. Maybe with another keyword like `valuesMerge`, this would also ensure backwards compatibility. Defining both `values` and `valuesMerge` should result in an error.

```
templates:
  default: &default
    values:
    - defaults.yaml    
  env1: &env1
    <<: *default
    values: 
    - env1.yaml
  env2: &env2
    <<: *env2
    valuesMerge: 
    - env2.yaml

releases:
- name: demo
  <<: *env1
  valuesMerge:
  - demo.yaml
```

Which would result in 

```
- env1.yaml
- env2.yaml
- demo.yaml
```

I am not sure about array merging, maybe this could be a global flag? Alternatively, if it would be possible to use the values generated by the previous layers, the existing array values could be manually inserted when needed.

PR #837, opened 2019-09-03
doc: Update documentation about layering release values

a-hat:
`setTemplate` and `valuesTemplate` do not work, I think it is correct without the `Template` suffix.

I also added a note to the documentation regarding this issue https://github.com/roboll/helmfile/issues/836


mumoshu:
@a-hat Hey! Thanks for your support.

This part of the documentation specifically explain how `Release Templating` works. Are you sure you've tested it with the `Release Template` feature, or just regular values without nested template expressions in it?

mumoshu:
The point is that this shouldn't work:

```
    setTemplate:
    - name: '{{`{{ .Release.Name }}`}}'
      values: '{{`{{ .Release.Namespace }}`}}'
```

As until #823 nested template expressions that are starting with:

```
{{`
```

and ending with:

```
`}}
```

are rendered as-is without template expressions being evaluated.

a-hat:
Hi @mumoshu!

After reading the documentation I tried something like this (note the `valuesTemplate` in the `templates` map) to set values in a release template:

```
templates:
  default: &default
    valuesTemplate:
    - image: default    

releases:
- name: demo
  <<: *default
  chart: stable/traefik
  version: ^1.70.2
  values:
  - image: demo
```

which gives me these errors:

```
in ./helmfile.yaml: failed to read helmfile.yaml: reading document at index 1: yaml: unmarshal errors:
  line 3: field valuesTemplate not found in type state.TemplateSpec
  line 3: field valuesTemplate not found in type state.ReleaseSpec
```

 Is the documentation referring to something else which I am missing?

mumoshu:
@a-hat `valuesTemplate` should be available since helmfile v0.82.0! Which version of helmfile are you using?

Btw, in your example you don't need `valuesTemplate`. Just use `values`.

```
templates:
  default: &default
    values:
    - image: default    

releases:
- name: demo
  <<: *default
  chart: stable/traefik
  version: ^1.70.2
  values:
  - image: demo
```

a-hat:
I see, thanks for the clarification. I was using version 0.81.3, so this is why `valuesTemplate` did not work.

I updated the PR and left the section about layering `values`. Does this make sense for you?

mumoshu:
@a-hat Thanks!

Would you mind prepend `## Layering Release Values` to the sentences you've just added?

The sentences are correct. I believe it should reside within a dedicated section though.

a-hat:
Thanks for your feedback! Makes sense, I have updated the doc.

PR #856, opened 2019-09-13
feat: `helm repo add --ca-file` via repositories definition

### LINKED ISSUE #855: feature request: support --ca-file flag to helm repo add
The `helm repo add` command has a `--ca-file` that can be used to connect to `https` repositories with self-signed certificates, which [isn't currently supported in helmfile](https://github.com/tmeneau/helmfile/blob/9d851cda3be9ee077afa6859f58410c761babf2c/pkg/helmexec/exec.go#L71-L84) an option corresponding to [Helm's --ca-file](https://github.com/helm/helm/issues/3074).

While this is kind of an edge case it did come up for me.

#### Proposal:
The helmfile's `repositories` spec would be updated to support a new `caFile` option for each repository, which (exactly like the `certFile` and `keyFile` options) just be proxied to the `helm repo add` command:

```yaml
repositories:
- name: insecure
   url: https://charts.my-insecure-domain.com
   caFile: optional_ca_crt
```

tmeneau:
This adds support for the `--ca-file` option to the `helm repo add` command via the RepositorySpec. The implementation for this is identical to the implementation for `RepositorySpec.CertFile` and `RepositorySpec.KeyFile`, and adds an example to the README.

While the use-case for this is fairly rare, it's a fairly easy implementation (unless I'm missing something here).

##### Possible Concern

Our helmfile usage is fairly vanilla, but this could be a breaking change if anyone uses helmfile as a Go library. I'm just not familiar enough with the various use-cases to know whether that's a problem, but I think that would be the main concern with this change.

resolves #855

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,11 @@ repositories:
   keyFile: optional_client_key
   username: optional_username
   password: optional_password
+# Advanced configuration: You can use a ca bundle to use an https repo
+# with a self-signed certificate
+- name: insecure
+   url: https://charts.my-insecure-domain.com
+   caFile: optional_ca_crt
 
 # context: kube-context # this directive is deprecated, please consider using helmDefaults.kubeContext
 
--- a/pkg/app/app_test.go
+++ b/pkg/app/app_test.go
@@ -1871,7 +1871,7 @@ func (helm *mockHelmExec) SetExtraArgs(args ...string) {
 func (helm *mockHelmExec) SetHelmBinary(bin string) {
 	return
 }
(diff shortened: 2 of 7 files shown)


tmeneau:
Wow, thanks for the quick merge @mumoshu! Glad to hear this shouldn't break anything for anyone. 

Also, thanks for sharing the link to your helmfile-operator project, that looks interesting.

tmeneau:
By the way, I wanted to say thank you @mumoshu for all the work on Helmfile. We recently started using it and have been loving it so far!

Thank you!

PR #861, opened 2019-09-17
doc: Added some detail on how to use override values in helmfiles section

arturo-c:
Not sure but looks like this is not a bug and is the intended use-case, so added some docs around it, hopefully it makes sense.  Let me know if I need to reword

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,7 +197,9 @@ helmfiles:
   # Override state values
   values:
   # Values files merged into the nested state's values
-  - additiona.values.yaml
+  - additional.values.yaml
+  # One important aspect of using values here is that they first need to be defined in the releases section
+  # of the origin helmfile, so in this example key1 needs to be in the values or environments.NAME.values of path/to/subhelmfile.yaml
   # Inline state values merged into the nested state's values
   - key1: val1
 - # All the nested state files under `helmfiles:` is processed in the order of definition.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -197,7 +197,10 @@ helmfiles:
   # Override state values
   values:
   # Values files merged into the nested state's values
-  - additiona.values.yaml
+  - additional.values.yaml
+  # One important aspect of using values here is that they first need to be defined in the releases section
+  # of the origin helmfile, so in this example key1 needs to be in the releases section of the chart you
```
mumoshu: This overrides state values rather than release values so the sentence should be like "key1 needs to be in the `values` or `environments.NAME.values` of `path/to/subhelmfile.yaml`"

arturo-c: @mumoshu made the change, let me know if it makes sense

README.md
```diff
@@ -197,7 +197,9 @@ helmfiles:
   # Override state values
   values:
   # Values files merged into the nested state's values
-  - additiona.values.yaml
+  - additional.values.yaml
+  # One important aspect of using values here is that they first need to be defined in the releases section
```
mumoshu:
```suggestion
  # One important aspect of using values here is that they first need to be defined in the values section
```

mumoshu: I'd suggest this to clarify more that this `values` does nothing to do with `releases[].values`.

README.md
mumoshu: Thanks! One more comment and this seems like good to go

PR #882, opened 2019-10-02
feat: Enhance postsync event to include `.Error`

### LINKED ISSUE #826: Should postsync hook only run if sync was successful?
The current implementation of the postsync hook is only semantically different to cleanup but they are run one after the other with the same parameters.

I would expect postsync to only run if the sync was successful (which I did assume until looking into the code) otherwise I don't understand why having 2 events that run sequentially with no different parameters is necessary. Or perhaps adding the success/failure status of the release to the context so it can be used in the templating of the command that gets run.

I'm happy to make a PR if this logic sounds correct, otherwise any reasoning or workaround I could use to determine if the sync was successful would be greatly appreciated.

jradtilbrook:
This resolves #826 by adding a success flag to the event template data for use in hooks. See the related issue for use-cases.

I've made a first pass that is rather crude so please provide feedback if you would like it done a different way. I had some other ideas but I couldn't think of one that felt much cleaner so I've opted for this simplest one at this point.

### DIFF
--- a/pkg/event/bus.go
+++ b/pkg/event/bus.go
@@ -20,8 +20,8 @@ type Hook struct {
 }
 
 type event struct {
-	Name    string
-	Success bool
+	Name  string
+	Error error
 }
 
 type Bus struct {
@@ -38,7 +38,7 @@ type Bus struct {
 	Logger   *zap.SugaredLogger
 }
 
-func (bus *Bus) Trigger(evt string, success bool, context map[string]interface{}) (bool, error) {
+func (bus *Bus) Trigger(evt string, evtErr error, context map[string]interface{}) (bool, error) {
 	if bus.Runner == nil {
 		bus.Runner = helmexec.ShellRunner{
 			Dir: bus.BasePath,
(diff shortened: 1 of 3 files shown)


pkg/event/bus.go
```diff
@@ -20,7 +20,8 @@ type Hook struct {
 }
 
 type event struct {
-	Name string
+	Name    string
+	Success bool
```
mumoshu: Hey! Thanks a lot for the PR.

After thinking for a while, I think this can better be `Error error` so that it can be used like `{{ if not .Error }}...{{ else }}...{{ end }}` which seems more idiomatic to users with some golang background.

It may also open up more possibilities like using hooks for error reporting to Slack/GitHub/MS Teams/etc, as you now have access to the error caused the installation to fail.

WDYT?

jradtilbrook: Oh wow I really like that, it heaps better! I'll make those changes when I have time 🚀

jradtilbrook:
Yay, no worries!

dudicoco:
@jradtilbrook @mumoshu can you please explain how can we use this feature? It's unclear to me and it's undocumented as well.

Thanks

mumoshu:
@dudicoco Similar to https://github.com/roboll/helmfile#global-hooks, you can use `{{ .Event.Error }}` to get the error object on error

PR #892, opened 2019-10-11
EnableBashCompletion

marcfrederick:


main.go
```diff
@@ -46,6 +46,7 @@ func main() {
 	cliApp.Name = "helmfile"
 	cliApp.Usage = ""
 	cliApp.Version = Version
+	cliApp.EnableBashCompletion = true
```
mumoshu: @marcfrederick Thanks for your contribution! Good to know that urfave/cli had support for bash completion.

I'm interested in merging this, but curious how we'd instruct users to enable it.

Should we just distribute `/etc/bash_completion.d/helmfile` via this git repository, and point users to follow the steps explained in https://github.com/urfave/cli#distribution?

marcfrederick: Yeah that would probably be the simplest solution. To be honest I originally made this change for myself only and decided I might as well push it upstream, but hadn't put much thought into how that would actually work.

PR #896, opened 2019-10-17
Update README.md

naseemkullah:
I think this is simpler, right?


naseemkullah:
Just read:

> Prior to this pull request, environment values were made available through the {{ .Environment.Values.foo }} syntax. This is still working but is deprecated and the new {{ .Values.foo }} syntax should be used instead.

PR #912, opened 2019-10-27
Add vendor files for go build

chenrui333:
As moving to the go modules build, leverage the vendors directory to help with the `go build` ( `go build -mod vendor` ).

Would be nice to include it in the next formula build.
https://github.com/Homebrew/homebrew-core/blob/master/Formula/helmfile.rb


davidovich:
Why?

Managing a vendor directory is more work for everyone:

* reviewers need to skim unrelated changes for PRs
* coders need not forget to amend the vendor dir
* a copy of dependencies bloat the repository

The on by default go proxy helps with removed dependencies.

chenrui333:
@davidovich Yeah, I think you are right, thanks for refreshing me on this matter.  :)

chenrui333:
After second thought, I think still vendoring is a good thing, even though I do think the management of vendoring is kind of hard. 

But that is why we need to automate the dependencies and vendoring process though.

mumoshu:
@chenrui333 Thanks! I'm trying to understand all around it.

Are you basically saying that vendoring would help faster builds while running `brew install helmfile`?

Was wondering why we can't just extract the pre-build binary from the archive:

https://github.com/Homebrew/homebrew-core/blob/f8474a5343d1ae1b9881b361011e1efa00691650/Formula/helmfile.rb#L4

chenrui333:
for homebrew-core, it would disallow the `pre-built binary` just like policy-wise.

And the vendoring stuff is more like my personal interest though. :)

I still think it does have some value, but it might be too much maintenance overhead.

PR #920, opened 2019-10-30
Run vals against values files that contain the syntax

tduffield:
If we specify ref+ syntax in a values file, run vals against that file. This should work for both `.yaml` and `.yaml.gotmpl` files, since it will evaluate the go-templating before proceeding to the ref evaluation.

I'm not sure the best way to test this. The initial implementation didn't have a test framework for me to leverage. I built locally and tested against my helmfile setup. It appears to work. 

## Example Usage

```
# helmfile.yaml
releases:
  - name: my-app
    chart: stable/foo
    values:
      - values.yaml
      - morevalues.yaml.gotmpl
```

```
# values.yaml
foo: ref+vault://127.0.0.1:8200/mykv/foo?proto=http#/mykey
bar:
  baz: ref+vault://127.0.0.1:8200/mykv/foo?proto=http#/mykey
```

```
# morevalues.yaml.gotmpl
boo: ref+vault://127.0.0.1:8200/{{ .Environment.Name }}/foo?proto=http#/{{ .Release.Name }}
```

Signed-off-by: Tom Duffield <tom@chef.io>

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -1542,7 +1542,33 @@ func (st *HelmState) flagsForLint(helm helmexec.Interface, release *ReleaseSpec,
 
 func (st *HelmState) RenderValuesFileToBytes(path string) ([]byte, error) {
 	r := tmpl.NewFileRenderer(st.readFile, filepath.Dir(path), st.valuesFileTemplateData())
-	return r.RenderToBytes(path)
+	rawBytes, err := r.RenderToBytes(path)
+	if err != nil {
+		return nil, err
+	}
+
+	// If 'ref+.*' exists in file, run vals against the file
+	match, err := regexp.Match("ref\\+.*", rawBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	if match {
+		var rawYaml map[string]interface{}
+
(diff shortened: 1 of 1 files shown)


mumoshu:
> The initial implementation didn't have a test framework for me to leverage. I built locally and tested against my helmfile setup. It appears to work.

Thanks! Yeah local testing would be enough for now. I'm still trying to figure out how we could easily test it.

tduffield:
@mumoshu if it were me, I would add a special no-op syntax to vals, something like `vals+test://myvalue` or `vals+echo://myvalue` which just evaluates to `myvalue`. That way you can write tests that can execute vals without having to worry about interfaces or handling secrets management. All you really care about is that vals is getting called in the right places.

mumoshu:
@tduffield Wow, that sounds great! I'll give it a try asap(Or even appreciate it if you could submit a pr to vals!

mumoshu:
@tduffield JFYI, I've just added an impl for the `echo` provider in https://github.com/variantdev/vals/commit/9d1714c4386c65fe120e9427ec230599748aa27c. Hoping it works as advertised 😄

PR #933, opened 2019-11-03
Use kind for setting up E2E clusters

mumoshu:


mumoshu:
For your specific case, `kind` seems to require more time to set up and a bit faster e2e testing. Overall, minikube makes the CircleCI job faster. So I'd keep using minikube for now.

Issue #932, opened 2019-11-02
feat: Double-render-free Helmfile config template


mumoshu:
# TL;DR;

I want to add a new helmfile.yaml field to make templating helmfile configs easier and more powerful than ever.

# Problem

Helmfile's double-rendering has opened a wide variety of use-cases that requires you to write a dynamic Helmfile configs.

The fact it works by recursively rendering the template twice to avoid the chicken-and-egg.

The problem is that Helmfile needs to load environment values to render the template but to load the values you need to render the template at first. It breaks in many ways and something results in behaviors that are hard to understand. See [the incomplete list of those issues](https://github.com/roboll/helmfile/issues?utf8=✓&q=label%3Adouble-rendering-gotcha) to get more context.

# Solution

I'd like to propose an alternative format for Helmfile config that doesn't suffer from the chicken-and-egg problem.

Let's see.

`helmfile.yaml`:

```yaml
values:
- foo: bar

environments:
  prod:
    value:
    - prod.yaml

template:
  # HELPERS: Template snippets that are reusable from multiple Helmfile configs
  #
  # Any files listed here can be called with {{ tempalte "BASENAME" }}
  # See https://golang.org/pkg/html/template/#ParseFiles
  helpers:
  - helpers/release.tpl
  # RELEASES: This is the alternative version of `releases` that is rendered by the template
  # In addition to `.Environment.Values`, you can acess template snippets declared above.
  releases: |
  - {{ template "release.tpl" (dict "name" "myapp1") | nindent 2}}
  - {{ template "release.tpl" (dict "name" "myapp1") | nindent 2}}
```

`helpers/release.tpl`:

```
name: {{ .Name }}
chart: ...
{{ if .Namespace }}
namespace: {{ .Namespace }}
{{ end }}
```

Notice that the new `template`, `template.helpers` and `template.releases` fields are just plain YAML primitives! This means that Helmfile can load the whole file as a valid YAML to environment values, and then render the template. We don't need the double-rendering anymore, which makes the result more predictable.

davidovich:
I also think double-rendering has made it's time. I'm very glad that @mumoshu tried the experiment, and this is OSS at it's best: build upon ideas to find better ones. 

On the proposal:

Instead of `template.helpers` would you consider using: `template.context`, `template.template-context` or `template.render-context` or variations on the context theme ? I like `template.template-context` as it hints on the usage of the `{{ template ... }}` feature of go that you mention.

[...]

PR #937, opened 2019-11-06
Enable readFile in first-pass renderer

jdemeyer:
It's not clear to me why `readFile` is disabled in the first-pass renderer. The implementation talks about "side-effect template calls" but I don't really see why reading a file is considered a side effect.

For me, the main thing is that I don't want to see the message
```
could not deduce `environment:` block, configuring only .Environment.Name. error: failed to read iae.yaml.part.0: reading document at index 1: yaml: unknown anchor 'iae' referenced
```
in my log files: it sounds like an error while it's not.

But when you think about it, using `readFile` seems useful in combination with the feature that the two-pass renderer adds (using values defined in `environment` blocks).

### DIFF
--- a/pkg/tmpl/context_funcs.go
+++ b/pkg/tmpl/context_funcs.go
@@ -31,9 +31,6 @@ func (c *Context) createFuncMap() template.FuncMap {
 		funcMap["exec"] = func(string, []interface{}, ...string) (string, error) {
 			return "", nil
 		}
-		funcMap["readFile"] = func(string) (string, error) {
-			return "", nil
-		}
 	}
 
 	return funcMap


mumoshu:
@jdemeyer Thanks!

Sorry but unfortunately, this is unlikely to happen... The first render is really for reading `values` and `environments` and nothing else. And `{{ readFile }}` affecting that is a debugging and user-support nightmare.

> For me, the main thing is that I don't want to see the message

Do you see it only when you've enabled debug logging, right...?
Then, the main problem seems to be you wanna some indication that the error is tolerable and handled correctly.

jdemeyer:
> Do you see it only when you've enabled debug logging, right...?

No, it's logged on the "info" level. If you would change that to "debug", that would also work for me.

jdemeyer:
...or even just make it more clear that it's essentially a harmless message that doesn't indicate a problem which needs to be fixed.

mumoshu:
@jdemeyer Thanks. Doing both would be nice!

Do you have any idea how the message would look like so that it could tell you its harmless?

mumoshu:
Or the right path in mid-term would be #932 which allows us to eventually remove the double-rendering.

evangelinepicache:
Fix my request

evangelinepicache:
Pull reqwest

evangelinepicache:
Pull requrst

Fradulentus:
Что именно обсуждаем?

PR #938, opened 2019-11-06
Allow combining valuesTemplate and values, same for setTemplate

jdemeyer:
Right now, one has to use either `valuesTemplate` or `values`. This is unexpected, it would be much more natural to allow using both at the same time. This implements that. The same for `setTemplate` and `set`.

### DIFF
--- a/pkg/state/release.go
+++ b/pkg/state/release.go
@@ -93,7 +93,7 @@ func (r ReleaseSpec) ExecuteTemplateExpressions(renderer *tmpl.FileRenderer) (*R
 		result.Labels[key] = s.String()
 	}
 
-	if result.ValuesTemplate != nil && len(result.ValuesTemplate) > 0 {
+	if len(result.ValuesTemplate) > 0 {
 		for i, t := range result.ValuesTemplate {
 			switch ts := t.(type) {
 			case map[interface{}]interface{}:
@@ -117,7 +117,9 @@ func (r ReleaseSpec) ExecuteTemplateExpressions(renderer *tmpl.FileRenderer) (*R
 			}
 		}
 
-		result.Values = result.ValuesTemplate
+		var newvals []interface{}
+		newvals = append(newvals, result.ValuesTemplate...)
+		result.Values = append(newvals, result.Values...)
 	}
 
 	for i, t := range result.Values {
@@ -139,7 +141,7 @@ func (r ReleaseSpec) ExecuteTemplateExpressions(renderer *tmpl.FileRenderer) (*R
 		result.Secrets[i] = s.String()
 	}
 
-	if result.SetValuesTemplate != nil && len(result.SetValuesTemplate) > 0 {
+	if len(result.SetValuesTemplate) > 0 {
 		for i, val := range result.SetValuesTemplate {
 			{
 				// name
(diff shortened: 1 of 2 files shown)


pkg/state/release.go
```diff
@@ -117,7 +117,7 @@ func (r ReleaseSpec) ExecuteTemplateExpressions(renderer *tmpl.FileRenderer) (*R
 			}
 		}
 
-		result.Values = result.ValuesTemplate
+		result.Values = append(result.ValuesTemplate, result.Values...)
```
mumoshu: This has some possibility to mutate result.Values.Template. I think we should use this intead:

```
result.Values = append([]string{}, result.Values...)
result.Values = append(result.Values, , result.ValuesTemplate...)`

File: pkg/state/release.go [hunk 2]
```diff
@@ -178,7 +178,7 @@ func (r ReleaseSpec) ExecuteTemplateExpressions(renderer *tmpl.FileRenderer) (*R
 			}
 		}
 
-		result.SetValues = result.SetValuesTemplate
+		result.SetValues = append(result.SetValuesTemplate, result.SetValues...)
```
mumoshu: The same

mumoshu:
@jdemeyer Thanks! What was your use-case? I have never had a chance to use `values` and `valuesTemplate` at the same time.

jdemeyer:
I made the requested changes.

The use case is just that we have charts which take many values. Some of these have a fixed structure and can be defined in the template (as explained in https://github.com/roboll/helmfile/blob/master/docs/writing-helmfile.md), but some values really need to customized on a case-by-case basis and cannot be templated. So I would use `valuesTemplate` for the former and `values` for the latter. With this PR, this works as expected.

mumoshu:
@jdemeyer Thanks! LGTM overall.

I'm currently working hard to stabilize the `needs`/DAG feature I introduced recently. So I'd be glad if you could give me a day or two until this gets merged.

Until then, I'd appreciate it if you could add some test to ensure this works and not regress.

Fradulentus:
Всем привет.

jdemeyer:
> Until then, I'd appreciate it if you could add some test to ensure this works and not regress.

Done.

pkg/state/release.go
```diff
@@ -178,7 +180,9 @@ func (r ReleaseSpec) ExecuteTemplateExpressions(renderer *tmpl.FileRenderer) (*R
 			}
 		}
 
-		result.SetValues = result.SetValuesTemplate
+		var newvals []SetValue
+		newvals = append(newvals, result.SetValuesTemplate...)
+		result.SetValues = append(newvals, result.SetValues...)
```
mumoshu: @jdemeyer Hey! This isn't a blocker, but can I reverse the order of values later?

I think the general usage of `values` in combination with `valuesTemplate` is that, `values` is the default values across environments, and `valuesTemplate`  are used for overlaying specific configuration per environment(or alike). So it should be `Values + ValuesTemplate`, rather than `ValuesTemplate + Values` we have now.

jdemeyer: I'm not sure that changing the order *later* is such a good idea. Better change it *now*, before people start relying on it.

Ideally, release-specific `values` should come last (after `valuesTemplate`) but default `values` should come first (before `valuesTemplate`). I haven't checked how hard it would be to implement this.

jdemeyer: Actually, after reading the code, it seems that my change only affects the parsing of `values` and `valuesTemplate` within the `release:` block and it doesn't affect default or environment-specific values. If this analysis is correct, I think that it is doing the right thing.

mumoshu: Could you clarify a bit? There's really no "release specic default" values.

Also, environment values has generally nothing to with release specific values.

You can derive release values from env values, but that happens only when you write some go template within helmfile.yaml.

mumoshu: "Later" here means a day or two if we could settle on something :) So that we won't break someone's helmfile.yaml.

PR #963, opened 2019-11-14
feat: Automatically enable Helm v3 mode

### LINKED ISSUE #923: Automatically enable Helm v3 mode
Currently `HELMFILE_HELM3` must be set to a non-empty value to enable the v3 mode. It works but not intuitive. Can't we do better?

My quick idea is that Helmfile could run `helm version -c`(`-c` is the shorthand for `--client` which skips contacting Tiller for the version number. It's v2 only) to detect the major version of the helm binary being used?

That is, if `helm version -c` returns 0 it should be v2 and if 1 and the error message was `Error: unknown shorthand flag: 'c' in -c` it should be v3.

```
$ helm version -c
Client: &version.Version{SemVer:"v2.14.3", GitCommit:"0e7f3b6637f7af8fcfddb3d2941fcc7cbebb0085", GitTreeState:"clean"}
$ echo $?
0
$ helm3b5 version -c
Error: unknown shorthand flag: 'c' in -c
$ echo $?
```

drakedevel:
Runs `helm version` in helmexec.New, and exposes a method on Interface to allow other packages to use the detected version. Preserves compatibility with previous HELMFILE_HELM3 mechanism.

Resolves #923

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -136,7 +136,7 @@ jobs:
     - run:
         name: Install helm
         environment:
-          HELM_VERSION: v3.0.0-rc.2
+          HELM_VERSION: v3.0.0
         command: |
           HELM_FILENAME="helm-${HELM_VERSION}-linux-amd64.tar.gz"
           curl -Lo ${HELM_FILENAME} "https://get.helm.sh/${HELM_FILENAME}"
--- a/Dockerfile.helm3
+++ b/Dockerfile.helm3
@@ -11,10 +11,10 @@ FROM alpine:3.10
 
 RUN apk add --no-cache ca-certificates git bash curl jq
 
-ARG HELM_VERSION=v3.0.0-rc.2
+ARG HELM_VERSION=v3.0.0
 ARG HELM_LOCATION="https://get.helm.sh"
 ARG HELM_FILENAME="helm-${HELM_VERSION}-linux-amd64.tar.gz"
-ARG HELM_SHA256="b6fff8e01aa6cd9a4541bd48172bb53b9a0ae38d7e7783a8e0fcc1db63802aaa"
+ARG HELM_SHA256="10e1fdcca263062b1d7b2cb93a924be1ef3dd6c381263d8151dd1a20a3d8c0dc"
 RUN wget ${HELM_LOCATION}/${HELM_FILENAME} && \
     echo Verifying ${HELM_FILENAME}... && \
     sha256sum ${HELM_FILENAME} | grep -q "${HELM_SHA256}" && \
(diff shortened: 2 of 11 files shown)


drakedevel:
Whoops, that CI failure is real -- the `TestHelmState_flagsForUpgrade` passes in a `ShellRunner` and tries to execute the system `helm` (which worked on my system but thankfully failed in CI). I'll pull in `mockRunner` for that test first thing tomorrow.

.circleci/config.yml
```diff
@@ -136,7 +136,7 @@ jobs:
     - run:
         name: Install helm
         environment:
-          HELM_VERSION: v3.0.0-rc.2
+          HELM_VERSION: v3.0.0
```
mumoshu: I think this resulted in the integration test to fail:

image

Actually, I've also encountered this while bumping this along with the helm version in Dockerfile.

The fix was https://github.com/roboll/helmfile/pull/965/files#diff-452b35dd32cfe6d2251b4e8a4f7e1fc0R49

drakedevel: Ah, thanks for the fix!

The integration test actually failed once before I made that change -- I tried bumping to 3.0.0 but was nearly midnight and I missed quite a few places :stuck_out_tongue: I'll rebase onto #965.

drakedevel:
@mumoshu all green! I rebased on top of the un-merged #965; the changes from that appear here since that PR hasn't been merged yet.

I also added a tiny test for the Helm 3 detection logic for good measure.

I did want to quickly call one thing out about this change: as we saw in the integration test failures, this PR makes helmfile blow up on Helm 3 <3.0.0-rc.3. Is that OK? Confidently detecting the versions in the middle is possible (although pretty ugly), but I was thinking that since all affected versions are weeks-old prereleases it's probably better to just tell folks to upgrade to a released version.

mumoshu:
> this PR makes helmfile blow up on Helm 3 <3.0.0-rc.3. Is that OK?

@drakedevel Good point! I think that's okay.

Btw, just to be clear, does it fail on rc.2 and lesser? I was thinking that you've already done a great job on keeping the check for `HELMFILE_HELM3` at https://github.com/roboll/helmfile/pull/963/files#diff-766154150f2383ce34784cd65a8f9862R51 which keeps rc.2 and lesser working.

But again, this PR looks great regardless of it actually works for rc.2 and lesser.

Thanks a lot for your contribution!

drakedevel:
> Btw, just to be clear, does it fail on rc.2 and lesser?

@mumoshu that's right -- but only if you don't specify `HELMFILE_HELM3`  (since that bypasses the `version` check completely). So I guess that's actually less of a big deal -- you had to already been using that env var, so upgrading won't actually break anything.

> Thanks a lot for your contribution!
My pleasure, thanks for the fast reviews!

PR #981, opened 2019-11-18
Sending basePath to yamls renderer

anih:
Hi,

I would like to propose this change because we faced a problem when we have app X that depends on app Y and we wanted to put Y in X dependencies:
```
helmfiles:
  - path: git::https://github/anih/Y.git@helmfile/helmfile.yaml?ref=master
```

But in Y we have 
```
helmfiles:
  - path: git::https://github/anih/generic-app.git@helmfile/helmfile.yaml?ref=master
    values:
      - extraValuesTemplates:
        - {{ requiredEnv "PWD" }}/values/values-file.yaml.gotmpl
```
unfortunately after removing `{{ requiredEnv "PWD" }}` i was getting error like 
```
values file matching "/project/.helmfile/cache/https_gihub_generic-app_git.ref=master/generic-api/values/values-file.yaml.gotmpl" does not exist in "/project/.helmfile/cache/https_gihub_generic-app_git.ref=master/generic-api"
```
With this patch all is working correctly.

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -240,8 +240,8 @@ func (a *App) within(dir string, do func() error) error {
 	return appErr
 }
 
-func (a *App) visitStateFiles(fileOrDir string, do func(string, string) error) error {
-	desiredStateFiles, err := a.findDesiredStateFiles(fileOrDir)
+func (a *App) visitStateFiles(absFileOrDir string, do func(string, string) error) error {
+	desiredStateFiles, err := a.findDesiredStateFiles(absFileOrDir)
 	if err != nil {
 		return appError("", err)
 	}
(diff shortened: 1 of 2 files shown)


mumoshu:
@anih Hey! I Thanks a lot for the PR.

I finally took some time to read this.

Just to confirm, is this change basically adds two things to helmfile?

- A template function called `basePath` to obtain the absolute path to the directory of the helmfile.yaml being processed
- Change the helmfile internal to use absolute pathes to locate files, so that the `basePath` becomes absolute one, where it was formerly relative

anih:
Hi,
Thank you for looking into it. Yes it's adding what you wrote. Probably rendering values from file it's done too late and path for `values/values-file.yaml.gotmpl` is taken from helmfile fro app X instead of Y app where this values file was added.

Issue #698, opened 2019-06-16
feat: version constraint.


nauxliu:
If I use the features introduced by the new version of helmfile, I can update the `required_version ` so my colleague knows that they need to upgrade the helmfile.

```yaml
required_version: ">= v0.78"

repositories:
  - name: stable
    url: https://kubernetes-charts.storage.googleapis.com

releases:
  - name: foo
    namespace: bar
```

mumoshu:
Resolved via #1016. Thanks for your contribution 🎉

PR #1016, opened 2019-12-05
feat: Persist Helmfile version in helmfile.lock for version compatibility check

nauxliu:
Reslove #698

### DIFF
--- a/pkg/state/chart_dependency.go
+++ b/pkg/state/chart_dependency.go
@@ -3,7 +3,9 @@ package state
 import (
 	"fmt"
 	"github.com/Masterminds/semver"
+	goversion "github.com/hashicorp/go-version"
 	"github.com/r3labs/diff"
+	"github.com/roboll/helmfile/pkg/app/version"
 	"github.com/roboll/helmfile/pkg/helmexec"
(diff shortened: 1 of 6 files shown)


nauxliu:
I can't find the CI logs on the UI, the build and test are working on my local machine.

nauxliu:
CI passed after re-push.

pkg/state/chart_dependency.go
```diff
@@ -404,6 +409,25 @@ func (m *chartDependencyManager) Resolve(unresolved *UnresolvedDependencies) (*R
 		return nil, false, err
 	}
 
+	// Make sure go run main.go works.
+	if version.Version != "" || lockedReqs.Version != "" {
```
mumoshu: Shouldn't this better be `version.Version != "" && lockedReqs.Version != ""`?

Otherwise L414 seems to fail due to an empty version when the lock file doesn't contain any version number, which is very likely `helmfile.lock` is generated by an older version of helmfile that doesn't include this feature.

nauxliu: Yes, you are absolutely correct.

Issue #899, opened 2019-10-19
--state-values-set for array or key with dot


eddycharly:
@mumoshu is this possible to pass a key containing a dot in its name to --state-values-set ?
also, does it support array access ?
i'm trying to do something like this :
```
    --state-values-set stacks.monitoring.prometheus.alertmanagerFiles.alertmanager\.yml.receivers[0].email_configs[0].auth_password=${MONITORING_SES_SECRET_KEY} \
    --state-values-set stacks.monitoring.prometheus.alertmanagerFiles.alertmanager\.yml.receivers[0].email_configs[0].auth_username=${MONITORING_SES_ACCESS_KEY} \
    --state-values-set stacks.monitoring.prometheus.alertmanagerFiles.alertmanager\.yml.receivers[0].email_configs[0].smarthost=${MONITORING_SES_ENDPOINT} \
```

mumoshu:
@eddycharly Hey!

I believe that it isn't implemented. It would be a great feature request though!

eddycharly:
@mumoshu I'm looking at implementing it.
From what is saw, helmfile is using urfave to parse the command line arguments.
Such a parsing does not seem to be natively supported by urfave, then i guess i could write a new flag parser for urfave.
Do you think it is the correct approach ?
Any hints on how to implement the feature welcome ;)

mumoshu:
@eddycharly Hey! Yes, that sounds like a correct approach.

Regarding the implementation, you'd need to implement something to handle index access to arrays in https://github.com/roboll/helmfile/blob/4294a5de0604226e3434b19c850df3db95a8dacf/main.go#L486-L489

We currently support only paths that are composed of dot-separated nested keys there. The enhancement should include a recursive loop to descend into the keys one by one, and a if-else to handle it according to that it's a regular scalar access (e.g. `.afield`) or array index access(e.g. `ary[0]`)

junpostrado:
--state-values-set stacks.monitoring.prometheus.alertmanagerFiles.alertmanager\.yml.receivers[0].email_configs[0].auth_password=${MONITORING_SES_SECRET_KEY} \
    --state-values-set stacks.monitoring.prometheus.alertmanagerFiles.alertmanager\.yml.receivers[0].email_configs[0].auth_username=${MONITORING_SES_ACCESS_KEY} \
[...]

PR #1042, opened 2019-12-16
Improve --state-values-set parsing

eddycharly:
This pull request adresses https://github.com/roboll/helmfile/issues/899

It adds support for indexed key arguments and escaped dot arguments.
Things like `--state-values-set config\.yml.something=abc` or `--state-values-set config.something[0]=abc` can be passed in the command line.

@mumoshu this should be ready for review.

### DIFF
--- a/pkg/maputil/maputil.go
+++ b/pkg/maputil/maputil.go
@@ -1,6 +1,9 @@
 package maputil
 
-import "fmt"
+import (
+	"fmt"
+	"strings"
+)
 
 func CastKeysToStrings(s interface{}) (map[string]interface{}, error) {
 	new := map[string]interface{}{}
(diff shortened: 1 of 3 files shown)


pkg/maputil/maputil.go
```diff
@@ -60,32 +63,141 @@ func recursivelyStringifyMapKey(v interface{}) (interface{}, error) {
...
+		panic(fmt.Errorf("bug: unexpected length of key: %d", len(key)))
+	}
+
+	appender := getArg(key[0])
+
+	if len(key) == 1 {
+		appender.assign(m, value)
+	} else {
+		remain := key[1:]
+		appender.shift(m, func(m map[string]interface{}) {
```
mumoshu: I feel like I'd prefer:

```
nested := appender.getMap(m)
Set(nested, remain, value)
```

But please correct me this appender/assign/shift is a common idioms for writing algorithm like this 😃

eddycharly: Not sure why i implemented it this way, probably because i was in a functional mood ;)

I will change that right away.

eddycharly: done

pkg/maputil/maputil.go
```diff
@@ -60,32 +63,133 @@ func recursivelyStringifyMapKey(v interface{}) (interface{}, error) {
...
+	}
+	return r
+}
+
+func Set(m map[string]interface{}, key []string, value string) {
+	if len(key) == 0 {
+		panic(fmt.Errorf("bug: unexpected length of key: %d", len(key)))
+	}
+
+	appender := getArg(key[0])
```
mumoshu: I'd prefer this called something akin to `cursor` or `pointer` or `reference` so that it looks like:

```
cur := getCursor(key[0])

if len(key) == 1 {
    cursor.set(m, value)
} else {
    remain := key[1:]
    nested := cur.getMap(m)
    Set(nested, remain, value)
}
```

eddycharly: done ;)

mumoshu: I'm not fully sure, but perhaps this function can be shortened into this? 

```
func Set(m map[string]interface{}, key []string, value string) {
	if len(key) == 0 {
		panic(fmt.Errorf("bug: unexpected length of key: %d", len(key)))
	}
	
	for len(key) > 1 {
        	m, key = getCursor(key[0]).getMap(m), key[1:]
	}
	getCursor(key[0]).set(m, value)
}
```

mumoshu:
@eddycharly Ahh sorry I should have made comments at once! You are so responsive.

I think https://github.com/roboll/helmfile/pull/1042#discussion_r358215534 is the last comment and after that I think I'm fully ready to merge this. Great job!

pkg/maputil/maputil.go
```diff
@@ -60,32 +63,133 @@ func recursivelyStringifyMapKey(v interface{}) (interface{}, error) {
...
 	}
+}
+
+func (a indexedKeyArg) set(m map[string]interface{}, value string) {
+	t := a.getArray(m)
+	t[a.index] = value
+	m[a.key] = t
+}
+
+func getCursor(key string) arg {
```
mumoshu: Your usage of the strategy pattern here was eye-opening to me! If it was me I would have written this in switch-case nested in a for-loop, but yours is MUCH more readable. Good job.

pkg/maputil/maputil.go
eddycharly: I think so, i thought that the recursion was not necessary but didn't take time to review it.
Sorry, i just pushed the change.

pkg/maputil/maputil.go
eddycharly: Thank you for your review, i think you catched all the points i left behind ;)

PR #1046, opened 2019-12-20
feat: Option to pass apiVersions to `helm diff` and `helm template`

### LINKED ISSUE #1014: .Capabilities.APIVersions returns wrong results
The helm built-in object `Capabilities.APIVersions` returns incorrect results when running with helmfile in my cluster.

For demonstration purposes, I have the following `test.yaml` file, which prints the results from `APIVersions` using `incubator/raw`:

```
templates:
- |
  apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: test
    labels:
      test: |
        {{ .Capabilities.APIVersions }}
```

Here is a `helmfile.yaml` using this template:

```
repositories:
- name: incubator
  url: https://kubernetes-charts-incubator.storage.googleapis.com
  
releases:
- name: test
  namespace: monitoring
  chart: incubator/raw
  version: 0.2.3
  values:
  - test.yaml
```

When I run `helm install test incubator/raw -f test.yaml --dry-run` I get the correct list of API Versions available in my cluster. `helmfile template` outputs a different set of Versions with many entries missing.

```
helmfile version v0.94.1
# kubectl version (AWS EKS)
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.3", GitCommit:"b3cbbae08ec52a7fc73d334838e18d17e8512749", GitTreeState:"clean", BuildDate:"2019-11-13T11:23:11Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.8-eks-b8860f", GitCommit:"b8860f6c40640897e52c143f1b9f011a503d6e46", GitTreeState:"clean", BuildDate:"2019-11-25T00:55:38Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
# helm version
version.BuildInfo{Version:"v3.0.0", GitCommit:"e29ce2a54e96cd02ccfce88bee4f58bb6e2a28b6", GitTreeState:"clean", GoVersion:"go1.13.4"}
```

a-hat:
resolves #1014

This makes it possible to pass the API Capabilities to helmfile when executing a task that does not render against an actual cluster (diff, template, apply).

The list of api versions are currently passed via `helmDefaults`. Maybe it would also make sense to override in specific `environments`, but this is not implemented.

This is my first contribution to a golang project, any feedback will be appreciated :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,9 @@ helmDefaults:
   tlsKey: "path/to/key.pem"
   # limit the maximum number of revisions saved per release. Use 0 for no limit (default 10)
   historyMax: 10
+  # list of api versions to pass to `helm diff` and `helm template` via the --api-versions flag
+  apiVersions:
+  - example/v1
 
 
 # The desired states of Helm releases.
--- a/README.md
+++ b/README.md
@@ -94,10 +94,6 @@ helmDefaults:
   tlsKey: "path/to/key.pem"
   # limit the maximum number of revisions saved per release. Use 0 for no limit (default 10)
   historyMax: 10
-  # list of api versions to pass to `helm diff` and `helm template` via the --api-versions flag
-  apiVersions:
-  - example/v1
-
 
 # The desired states of Helm releases.
 #
@@ -270,6 +266,18 @@ bases:
 - environments.yaml
 - defaults.yaml
 - templates.yaml
+
+#
+# Advanced Configuration: API Capabilities
+#
+# Some helmfile tasks render releases locally without querying an actual cluster (diff, apply, template),
+# and in this case `.Capabilities.APIVersions` cannot be populated.
+# When a chart queries for a specific CRD, this can lead to unexpected results.
+# 
+# Configure a fixed list of api versions to pass to helm via the --api-versions flag:
+apiVersions:
+- example/v1
+
 ```
 
 ## Templating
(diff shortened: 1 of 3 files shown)


mumoshu:
@a-hat Thanks! Code LGTM. But can we reconsider where to put it?

I think it would better added in the root of helmfile.yaml, so it looks like:

```
apiVersions:
- ...

helmDefaults:
  ...

releases:
  ...
```

The reason is that `helmDefaults` implies that you can overwrite it per each `release` under `releases`. I believe it doesn't need to be overwritable per release by its nature.

a-hat:
Hi @mumoshu 

Thanks for your feedback. Yes, this makes sense. I moved the `apiVersions` to the root of helmfile.

a-hat:
Hi @mumoshu 

Apparently helm-diff does not support the `--api-versions` flag. I only tested with `helm template`, sorry for not catching that.

I will try to submit a PR to helm-diff for supporting the flag.

Currently `helmfile apply` or `helmfile diff` fail when `apiVersions` is set in helmfile. Should we revert this code until helm-diff supports the flag?

a-hat:
Here is the PR in helm-diff

https://github.com/databus23/helm-diff/pull/175

Issue #1050, opened 2019-12-23
`helmfile test` doesn't specify namespace


RaymondKYLiu:
Hi, 
I appreciate the powerful tool, and I want to feedback an issue.

When I use the command `helmfile test`, it doesn't pass the specific namespace defined from `helmfile.yaml` to the `helm test` command's argument.

It's easy to reproduce it, I just `helm create demo-nginx` create the default chart and my `helmfile.yaml` is below

```yaml
# helmfile.yaml
environments:
  test:
  stag:
  prod:
releases:
  - name: demo-nginx
    namespace: demo
    chart: ../../../charts/demo-nginx/
    version: 0.1.0
```
Deploy it
```shell
$ helmfile --environment test --file helmfile.yaml --log-level Info --selector namespace=demo,name=demo-nginx sync
```

Test it with --log-level Debug
```shell
$ helmfile --environment test --file helmfile.yaml --log-level Debug --selector namespace=demo,name=demo-nginx test
...

1 release(s) matching namespace=demo,name=demo-nginx found in helmfile.yaml

receiving result 0
worker 1/1 started
Testing demo-nginx
exec: helm test demo-nginx --timeout 300s
exec: helm test demo-nginx --timeout 300s: 
sending result for release: demo-nginx

sent result for release: demo-nginx

worker 1/1 finished
received result for 0
err: release "demo-nginx" failed: helm exited with status 1:
  Error: release: not found
changing working directory back to "/home/raymond/Projects/k8s-cluster/k8s/deployments"
changing working directory back to "/home/raymond/Projects/k8s-cluster"
in k8s/deployments/helmfile.yaml: in .helmfiles[15]: in demo/demo-nginx/helmfile.yaml: release "demo-nginx" failed: helm exited with status 1:
  Error: release: not found
```

As you can see, the `exec: helm test demo-nginx --timeout 300s` didn't contain `--namespace demo` argument.

A workaround is passing argument with `--args` e.g.
```shell
$ helmfile --environment test --file helmfile.yaml --log-level Debug --selector namespace=demo,name=demo-nginx test --args "--namespace demo"

receiving result 0
worker 1/1 started
Testing demo-nginx
exec: helm test demo-nginx --timeout 300s --namespace demo
...
```
I believe it would be great if the feature be implemented :)

mumoshu:
Good catch! This should be implemented.
Would you mind submitting a pull request for it?

RaymondKYLiu:
Hi @mumoshu,
[...]

PR #1098, opened 2020-02-12
Fix status command for helm 3

rinrailin:
fixes same as https://github.com/roboll/helmfile/issues/1050 , but for `helmfile status` command.


eshepelyuk:
@rinrailin @mumoshu i think this should be reopened
https://sweetops.slack.com/archives/CE5NGCB9Q/p1582088381195000

PR #1100, opened 2020-02-13
Upgrade deps

chenrui333:
relates to https://github.com/Homebrew/homebrew-core/pull/50129

Upgrade deps for helmv2, v3 and golang


chenrui333:
Looks like circleci golang v1.13.8 image is not quite ready yet.

PR #1102, opened 2020-02-17
Document HELMFILE_HELM3

nauxliu:


mumoshu:
@nauxliu Hey! Thanks for the PR. But is this still required for you to enable the Helm 3 mode?

Since #963, Helmfile should automatically run `helm version -c` to detect the version number of helm being run and enables the helm 3 mode when necessary.

nauxliu:
Oh, I didn't notice that PR. I thought I still need to set the environment manually. Thanks!

Issue #1111, opened 2020-02-20
Environment Secrets not available under .Environment.Secrets.KEY but under .Environment.Values.KEY


tstmrk:
When I configure an (encrypted) secret file under environments
I expect that the secrets will be available under .Environment.Secrets.KEY  (according to the docs and examples)
But instead they are available under .Environment.Values.KEY 

`production:
    values:
    - environment/production/values.yaml
    - myChartVer: 1.0.0
    ## `secrets.yaml` is decrypted by `helm-secrets` and available via `{{ .Environment.Secrets.KEY }}`
    secrets:
    - environment/production/secrets.yaml`

When I use  .Environment.Secrets I get an error:

`can't evaluate field Secrets in type environment.Environment`

mumoshu:
They're available under `.Environment.Values`.
Where in the docs did you see the `.Environment.Secrets.KEY` notation? We need to fix that!

Ibh9mZdI6YIDyJUn48IZ:
Ah, all right! It's in at line 247 @ https://github.com/roboll/helmfile/blame/master/README.md

tstmrk:
Ah, all right! It's in at line 247 @ https://github.com/roboll/helmfile/blame/master/README.md

tstmrk:
https://github.com/roboll/helmfile/pull/1116

mumoshu:
Resolved via #1116. Thanks for your PR @tstmrk!

PR #1116, opened 2020-02-22
fix explanation about secrets in .Environment.Values

tstmrk:
Just fixed it myself ;-)


tstmrk:
https://github.com/roboll/helmfile/issues/1111

Issue #1125, opened 2020-02-27
Failed docker build 


kuzm1ch:
```
docker build -f Dockerfile.helm3 . 
docker build -f Dockerfile .
```
failed on step  `RUN make static-linux`

```
go: finding github.com/dimchansky/utfbom v1.1.0
env CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GOFLAGS=-mod=vendor go build -o "dist/helmfile_linux_amd64" -ldflags '-X github.com/roboll/helmfile/pkg/app/version.Version=v0.101.0' 
build github.com/urfave/cli: cannot load github.com/urfave/cli: open /workspace/helmfile/vendor/github.com/urfave/cli: no such file or directory
make: *** [Makefile:33: static-linux] Error 1
The command '/bin/sh -c make static-linux' returned a non-zero code: 2
```

mumoshu:
@kuzm1ch Could you try `go mod vendor` followed by `make image` or `make image/helm3`?

https://github.com/roboll/helmfile/blob/0186254e798e087d42153e58c08ec018663e99fb/.circleci/config.yml#L58-L59

PR #1131, opened 2020-03-02
WIP: (#1125) ask question

kuzm1ch:
I have question about issue with docker build failed. 
Is the are a reason to use `go mod vendor` instead off `go mod download`? 
I can advise as follow.

### DIFF
--- a/Dockerfile.helm3
+++ b/Dockerfile.helm3
@@ -2,6 +2,8 @@ FROM golang:1.13.7-alpine3.11 as builder
 
 RUN apk add --no-cache make git
 WORKDIR /workspace/helmfile
+COPY go.mod go.sum ./
+RUN go mod download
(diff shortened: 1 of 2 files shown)


Dockerfile.helm3
```diff
@@ -2,6 +2,8 @@ FROM golang:1.13.7-alpine3.11 as builder
 
 RUN apk add --no-cache make git
 WORKDIR /workspace/helmfile
+COPY go.mod go.sum ./
+RUN go mod download
```
mumoshu: The point of `vendor` was to avoid running `go mod download` (or plain `go build` that results in downloading deps not in the cache) every time we run a docker build.

kuzm1ch: thanks for the reply, close it

PR #1133, opened 2020-03-04
Add Sportradar to USERS.md

yurrriq:


yurrriq:
Thank you! I'm glad we're finally on the list :sunglasses:

PR #673, opened 2019-06-11
feat: experimental integration with helm-x

mumoshu:
This enhances helmfile so that it can:

- Treat K8s manifests directories and Kustomize projects as charts
- Add adhoc chart dependencies on sync/diff/template without forking or modifying chart(s) (#649)
- Add adhoc patches(JSON Patch or Strategic Merge Patch supported) to be applied to the K8s resources before sync/diff/template, without forking or modifyin chart(s) (#650)

The usage is as outlined in https://github.com/mumoshu/helm-x/tree/master/examples/helmfile.

Add any or all of `dependencies:`, `jsonPatches:` and `strategicMergePatches:` so that it adds additional flags to `helm` calls that is only supported by `helm x`.

```yaml
releases:
- name: kustomize
  chart: ../kustomize
- name: manifests
  chart: ../manifests
- name: foo
  chart: incubator/raw
  dependencies:
  - alias: bar
    chart: incubator/raw
  values:
  - values.yaml
  - bar:
      enabled: true
      resources:
      - apiVersion: v1
        kind: Pod
        metadata:
          name: bar
        spec:
          containers:
          - command:
            - sleep
            - 1000
            image: alpine:3.9.4
            imagePullPolicy: IfNotPresent
            name: bar
  jsonPatches:
  - target:
      version: v1
      kind: Pod
      name: foo
    patch:
    - op: replace
      path: /spec/containers/0/command
      value:
      - sleep
      - "123"
  strategicMergePatches:
  - apiVersion: v1
    kind: Pod
    metadata:
      name: bar
    spec:
      containers:
      - name: bar
        command:
        - sleep
        - "234"
```

You can alternatively provide `source: path/to/patch.yaml` for `jsonPatches` and `strategicMergePatches` items to externalize. Add a `.gotmpl` suffix like you would do for values files to enable templating.

To make this feature work, you must point `--helm-binary` to the `helm-x` binary like below:

```
$ helmfile --helm-binary ~/.helm/plugins/helm-x/bin/helm-x --log-level debug apply
```

The easiest way to grab the helm-x binary is to install it as a helm plugin:

```
$ helm plugin install https://github.com/mumoshu/helm-x
```

See [helm-x](https://github.com/mumoshu/helm-x) for more details on it.

The integration should ideally be automatic. That is, it shouldn't force you to set `--helm-binary`. But I had no other way to not bloat helmfile's codebase to just add this experimental feature.

Resolves #649
Resolves #650

PR #1172, opened 2020-04-04
feat: GA of Kustomize and K8s manifests support

### LINKED ISSUE #876: Support for multiple charts in single release (similar to alias in helm)
Currently we map single release in helmfile by using helm sub charts i.e.
```
subchart1: worker-type1
subchart2: worker-type2
chart1:
  subchart1-alias1:
    overide-variables1
  subchart2-alias1:
    overide-variables1
  subchart2-alias2:
    overide-variables2
```
where a chart refers to a subchart multiple times by using aliases in `requirements.yaml`. We then manage the deployment of chart1 by mapping it to a release in helmfile.

While this pattern works it is very difficult to iterate over this pattern. Figure this scenario:
* we want to update the subchart1 - for every end-to-end test on this we will need to update the chart1 to point to the new version of subchart1 and then update helmfile to point to new versions of chart1.
* we want the release to map directly to an application - for every application we want to have a single release so we can easily map and graph the unique names. while we can still work-around this by using a unique label across a single helmfile declaring multiple releases - but that would not be as consistent.

An ideal solution would be to flatten the structure and have no subchart and let helmfile manage the alias declaration. Effectively in helmfile release we call something like:
```
helmDefaults:
  tillerNamespace: {{ .Environment.Values.tiller }}
  kubeContext: {{ .Environment.Values.kubectx }}

environments:
  env1:
    values:
    - ../../envs/env1.yaml
  env2:
    values:
    - ../../envs/env2.yaml

releases:
  - name: app 
    namespace: {{ .Environment.Values.namespace }}
    charts:
    - alias: app-worker1
       chart: repo-name/app-worker1
       version: 0.1.4
    - alias: app-worker2
       chart: repo-name/app-worker1
       version: 0.1.4
    secrets:
      - secrets/app.{{ .Environment.Values.env }}.yaml
    values:
      - values/app.yaml.gotmpl
```
--_app.yaml.gotmpl_
```
app-worker1:
  override-values:
    aa: 112asas
    bb: 121212
app-worker2:
  override-values:
  cc: 12412
  dd: 124sdf
```

It would be very convenient to have a feature similar to helm `alias` in helmfile so we can call multiple instances of a single chart.

mumoshu:
This is the GA version of the helm-x integration #673 developed last year.

Benefits? You get all the followings without an extra helm plugin:

- Ability to add ad-hoc chart dependencies/aliases, without forking the chart (Fixes #876 )
- Ability to patch resulting K8s resources before installing the helm chart
- Ability to install a kustomization as a chart (Requires `kustomize` binary to be available in `$PATH`
- Ability to install a directory of K8s manifests as a chart
- etc.


lukasmrtvy:
any progress ?

dudicoco:
@mumoshu I can not get this feature to work.

helm version `version.BuildInfo{Version:"v3.2.1", GitCommit:"fe51cd1e31e6a202cba7dead9552a6d418ded79a", GitTreeState:"clean", GoVersion:"go1.13.10"}`

First attempt:
```
releases:
- chart: my-repo/amazing-app
  version: 1.0.4
  installed: true
  name: amazing-app
  namespace: devops
  labels:
    chart: amazing-app
    name: amazing-app
    namespace: devops
  jsonPatches:
  - target:
      version: v1
      kind: Pod
      name: static-web
    patch:
    - op: replace
      path: /spec/containers/0/command
      value:
      - sleep
      - "123"
```

output:
```
$ helmfile diff
I0527 15:36:00.959212   23656 chartify.go:236] using requirements.yaml:
dependencies:
  - name: base
    version: "^1.0.0"
    repository: "@my-repo"
    alias: amazing-app
in ./helmfile.yaml: [exit status 1

COMMAND:
  helm dependency build /var/folders/fj/scl5dj6d13b7mb34kxzwwfp80000gn/T/4001514311826455215/amazing-app

OUTPUT:
  manager.go:107: warning: a valid Helm v3 hash was not found. Checking against Helm v2 hash...
  Error: the lock file (requirements.lock) is out of sync with the dependencies file (requirements.yaml). Please update the dependencies]
```
Removing the `jsonPatches` block solves this issue.

Second attempt:
```
releases:
- name: foo
  chart: incubator/raw
  dependencies:
  - alias: bar
    chart: incubator/raw
  values:
  - bar:
      enabled: true
      resources:
      - apiVersion: v1
        kind: Pod
        metadata:
          name: bar
        spec:
          containers:
          - command:
            - sleep
            - 1000
            image: alpine:3.9.4
            imagePullPolicy: IfNotPresent
            name: bar
  jsonPatches:
  - target:
      version: v1
      kind: Pod
      name: static-web
    patch:
    - op: replace
      path: /spec/containers/0/command
      value:
      - sleep
      - "123"
```

Output:
```
$ helmfile diff
I0527 15:39:03.873601   23756 chartify.go:236] using requirements.yaml:
dependencies:
- name: raw
  repository: https://kubernetes-charts-incubator.storage.googleapis.com/
  condition: bar.enabled
  alias: bar
  version: "*"
I0527 15:39:13.813020   23756 replace.go:45] options: {false [] []  }
in ./helmfile.yaml: [invalid state: no files rendered]
```
Removing the `jsonPatches` and 'dependencies' blocks solves this issue.

PR #1188, opened 2020-04-07
feat: add tmpl function `required`

### LINKED ISSUE #572: Add "required" function identical to helm's
Feature Request:

`helm` has a `required` template function documented [here](https://github.com/helm/helm/blob/master/docs/charts_tips_and_tricks.md#know-your-template-functions). This is similar to, but more powerful than, `requiredEnv`. I would like `helmfile` to support the same function.

My use case is that I am writing releases where a value can come from multiple environment variables (as a way of implementing layers of configuration), so I cannot use `requiredEnv`, but I want to mark a value as required and provide a useful error message. For example:
```
dnsZone: {{ required "Either DNS_ZONE or CLUSTER_NAME required" (coalesce (env "DNS_ZONE") (env "CLUSTER_NAME")) }}
```

RaymondKYLiu:
Resolves #572 

Force error test with this `values.yaml.gotmpl` and without any environment variables `CLUSTER_NAME`  and `DNS_ZONE` has been set
```yaml
# values.yaml.gotmpl
dnsZone: {{ required "Either DNS_ZONE or CLUSTER_NAME required" (coalesce (env "DNS_ZONE") (env "CLUSTER_NAME")) }}
```

error message will be shown as:
> in helmfile.yaml: in .helmfiles[23]: in helmfile.yaml: failed to render values files "values.yaml.gotmpl": failed to render [values.yaml.gotmpl], because of template: stringTemplate:2:12: executing "stringTemplate" at <required "Either DNS_ZONE or CLUSTER_NAME required" (coalesce (env "DNS_ZONE") (env "CLUSTER_NAME"))>: error calling required: Either DNS_ZONE or CLUSTER_NAME required

### DIFF
--- a/pkg/tmpl/context_funcs.go
+++ b/pkg/tmpl/context_funcs.go
@@ -26,6 +26,7 @@ func (c *Context) createFuncMap() template.FuncMap {
 		"get":            get,
 		"getOrNil":       getOrNil,
 		"tpl":            c.Tpl,
+		"required":       Required,
 	}
 	if c.preRender {
 		// disable potential side-effect template calls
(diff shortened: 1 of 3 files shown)


RaymondKYLiu:
@mumoshu I'm not sure if #1185 will affect expected behavior or not. Maybe I can cherry-pick your change from #1185 .

PR #1190, opened 2020-04-08
Add conditional release support

### LINKED ISSUE #781: feat: First-class Support for Conditional Release
I'm seeing emerging use of `{{ if eq .Environment.Name "theenv" }}` in helmfile.yaml templates for making releases optional for a subset of environments.

To me, helmfile templates are necessarily evil for 

I'm considering to add Helmfile a first-class support for that use-case by adding `if` configuration key to `helmfile.yaml`:

```yaml
releases:
- name: istio
  chart: istio.io/istio
  if:
     environment:
     - dev
     - staging
```

Any thoughts?

acaire:
Add conditional release support. Fixes #781

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -156,6 +156,8 @@ type ReleaseSpec struct {
 	CleanupOnFail *bool `yaml:"cleanupOnFail,omitempty"`
 	// HistoryMax, limit the maximum number of revisions saved per release. Use 0 for no limit (default 10)
 	HistoryMax *int `yaml:"historyMax,omitempty"`
+	// Condition, when set, evaluate the mapping specified in this string to a boolean which decides whether or not to process the release
+	Condition string `yaml:"condition,omitempty"`
 
 	// MissingFileHandler is set to either "Error" or "Warn". "Error" instructs helmfile to fail when unable to find a values or secrets file. When "Warn", it prints the file and continues.
 	// The default value for MissingFileHandler is "Error".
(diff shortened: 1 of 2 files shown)


pkg/state/state.go
```diff
@@ -1282,14 +1284,18 @@ func (st *HelmState) GetReleasesWithOverrides() []ReleaseSpec {
...
+	}
+	rs, err := markExcludedReleases(st.GetReleasesWithOverrides(), st.Selectors, values)
 	if err != nil {
 		return nil, err
 	}
 	return rs, nil
 }
 
-func markFilteredReleases(releases []ReleaseSpec, selectors []string) ([]Release, error) {
+func markExcludedReleases(releases []ReleaseSpec, selectors []string, values map[string]interface{}) ([]Release, error) {
```
mumoshu, line 1298: Ah nice! "exclude" seems nicer here because I was always struggling about if "filter" means "filter in" or "filter out" 👍

acaire: Haha! Funny you mention that because I thought the same - I even had to [look it up](https://www.google.com/search?q=define%3Afilter) because I thought the usage was wrong, but surely enough it means to get rid of unwanted items!

itscaro:
@mumoshu @acaire If I'm not wrong, `helmfile status` does not respect this conditional flag:

https://github.com/roboll/helmfile/blob/master/pkg/state/state.go#L1785-L1801

mumoshu:
@itscaro As far as I remember, no. They should be filtered beforehand here

https://github.com/roboll/helmfile/blob/1ef9b29f6dce765e3a083e90467581ad2e6fccc6/pkg/app/app.go#L353-L366

`SetFilter(true)` should result in call the logic added in this PR.

Did you find `helmfile status` not respecting condition, anyway?

itscaro:
@mumoshu I have 2 envs, one has less charts enabled. 
running `helmfile lint` or `helmfile status`, helmfile does action on disabled charts.
I run `helmfile apply` then `helmfile status`, the first one deploys only enabled charts.

itscaro:
I've just do a quick debug:
At this step I still have all my charts in `st.ReleaseSetSpec.Releases` and `st.ReleaseSetSpec.Env.values` contains the right boolean. 
https://github.com/roboll/helmfile/blob/1ef9b29f6dce765e3a083e90467581ad2e6fccc6/pkg/app/app.go#L810

I've just had a look at the code, I don't know how it works yet. Maybe `st.ReleaseSetSpec.Releases` is never changed and is filtered later in the flow?

PR #1198, opened 2020-04-11
put tiller on path for use in helm-tiller plugin.

doughgle:
Put tiller binary on the path in helm v2.X container image so that helm-tiller plugin will use it instead of duplicating the download of helm tarball and performing version check.


doughgle:
@mumoshu if you're okay, please merge.

Issue #1201, opened 2020-04-13
Values are not populated in second pass renderer if preliminary state parsing had failed


dany74q:
Hey hey !

I've tackled a small issue (maybe ? would love your feedback) when executing helmfile on our uber-templated helmfile hierarchy that we've built in the past week.

To illustrate it, here's a small reproducer:

```yaml
# helmfile.yaml
environments:
  default:
    values:
      - ./values.yaml.gotmpl
---

{{ readFile "common.yaml.gotmpl" }}

releases:
  - name: {{ .Values.releaseName }}
    <<: *default


# common.yaml.gotmpl
templates:
  default: &default
    chart: charts/{{ .Values.chartName }}


# values.yaml.gotmpl 
releaseName: pasten
chartName: pasten-chart
```

I spot the following behavior:
- Using .Values.releaseName fails with: `in helmfile.yaml: error during helmfile.yaml.part.1 parsing: template: stringTemplate:5:20: executing "stringTemplate" at <.Values.releaseName>: map has no entry for key "releaseName"`
- Chaging it to `.Environment.Values.releaseName` fixes it, and it renders correctly

After debugging this, what I've found is:

- We mock `readFile` in the preliminary state to return an empty string
- This makes the prestate rendering fail, because the yaml anchor is undefined
- The failure is tolerated and we continue towards the second pass rendering
- When the prestate fails, the values map stays empty, while the environment values are correctly populated due to the previous part (layer)

This can be easily overcome by using (the deprecated / not recommended) `.Environment.Values`.

If that's in fact an issue (albeit probably an esoteric one), I think that a potential solution could be to populate the values regardless of whether or not the prestate had succeeded.

Would love to hear your thoughts.

Thanks !

Nuru:
I have run into situations where `helmfile` gives bad error message. I'm going to wait until this is fixed to see if that fixes the problem I'm seeing before opening an issue, but if you have the inclination to look into it:

What seems to be happening is that some error in first-pass rendering does not trigger a reported error, but leads to problems with the setup for the second-pass, and then we get a second-pass error like `<.Values.releaseName>: map has no entry for key "releaseName"` that is no help in figuring out the first-pass error, which could be, for example, an improperly indented comment that causes the template to stop processing prematurely.

dany74q:
Works great in 0.111.0 btw - thanks for approving !

PR #1202, opened 2020-04-13
Fix: populate .Values regardless of prestate success

dany74q:
This commit proposes a potential solution for
https://github.com/roboll/helmfile/issues/1201

The gist is that, if prestate rendering fails, for any reason,
we do not populate the .Values in the second pass renderer.

I think that what have been expected in this case is to populate the
.Values irregardless.

pkg/app/two_pass_renderer.go
- Migrated to use finalEnv.GetMergedValues()

pkg/environment/environment.go
- Introduced GetMergedValues, which merges the environment's defaults
and current values, and then casts the keys to string;
This was previously defined in HelmState.Values() - however, as this
method is only concerned with the environment, I think it's more
appropriate for it to sit here.

pkg/state/state_exec_tmpl.go
- Extracted out HelmState.Values() to environment.go, see above

### DIFF
--- a/pkg/app/two_pass_renderer.go
+++ b/pkg/app/two_pass_renderer.go
@@ -116,15 +116,13 @@ func (r *desiredStateLoader) twoPassRenderTemplateToYaml(inherited, overrode *en
 		r.logger.Debugf("first-pass rendering result of \"%s\": %v", filename, *finalEnv)
 	}
 
-	vals := map[string]interface{}{}
+	vals, err := finalEnv.GetMergedValues()
+	if err != nil {
+		return nil, err
+	}
+
 	if prestate != nil {
 		prestate.Env = *finalEnv
-		vals, err = prestate.Values()
-		if err != nil {
-			return nil, err
-		}
-	}
-	if prestate != nil {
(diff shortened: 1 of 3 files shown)


pkg/app/two_pass_renderer.go
```diff
@@ -116,15 +116,13 @@ func (r *desiredStateLoader) twoPassRenderTemplateToYaml(inherited, overrode *en
...
-		if err != nil {
-			return nil, err
-		}
+	vals, err := finalEnv.GetMergedValues()
+	if err != nil {
+		return nil, err
 	}
+
 	if prestate != nil {
+		prestate.Env = *finalEnv
```
mumoshu, line 125: Just asking - do you by any chance figure out why we mutate prestate here? (I lost my memory and I understand you've kept this in favor of the L121 before this change.

dany74q: I think we can safely skip the mutation actually - we don't really continue to use the prestate past this point;
I kept it b/c I wasn't sure if the mutation was intentional or not - didn't want to break anything (:

PR #1132, opened 2020-03-03
Bump golang to v1.14.2

chenrui333:

PR #1219, opened 2020-04-19
Another attempt to bump golang to v1.14.2 

renehernandez:
Based on the proposed changes from #1132 

It seems like the original PR has issues with some tab/spaces changes affecting the results of the tests


renehernandez:
@mumoshu Could you check why the PR is not being built by CircleCI? 

I set up a CircleCi project against my fork and the pipeline passed with the upgrade to 1.14 at https://app.circleci.com/pipelines/github/renehernandez/helmfile

chenrui333:
@renehernandez sorry for the delay, I think I think I made the PR thru. If you dont mind, we can close this in favor of that one. :)

PR #1221, opened 2020-04-20
feat(template): added secret template function

aldor007:
This adds tmpl `fetchSecretValue` and `expandSecretRefs`  functions by:
- Adding:
    - `fetchSecretValue`, `expandSecretRefs`  function in tmpl package
    - gomock for tests purpose
- Changing:
    - move init of vals package to function (so the same instance can be used for template values and rendering the whole template)

`secret` template function allows to use `vals` module for getting secrets

**Why?**

The current implementation of secrets from vault etc has an issue with rendering files like below:
```yaml
# env/dev/values.yaml
secret: ref+vault://key/#val
tenantId: ref+vault://key/#tenant
```

```yaml
# sevice.yaml.gotmpl
run.shell: |
  echo {{ .Values.secret }} | sudo passwd
loginUrl: https://something/{{ .Values.tenantId }}/oauth2
```
This will produce error like
```bash
failed to render values files "values/service.yaml.gotmpl": expand vault://key/#/val | sudo passwd 
```
// vals package is trying to get incorrect key 
Without knowing about implementations it is not easy to figure out that you can use it this way.

This PR introduces `secret` template function that uses the same module for retrieving secrets
Proposed solution: 
```yaml sevice.yaml.gotmpl
run.shell:
  echo {{ .Values.secret | secret }} | sudo passwd
loginUrl: https://something/{{ .Values.tenantId | secret }}/oauth2
```

 I'm waiting for feedback :)


mumoshu:
@aldor007 Hey! Thanks for your contribution.

Everyone asks this but I'm still unsure - are we really sure that this means `run.shell` with the secret embedded is written to the disk? Is that non-security issue for you? 🤔

mumoshu:
For your example, why you can't embed `run.shell` in the container image and only pass the secret via the container envvars, or treat the whole `run.shell` as a secret?

aldor007:
> For your example, why you can't embed run.shell in the container image and only pass the secret via the container envvars, or treat the whole run.shell as a secret?

> I think that's much safer.

In some cases, without modifying the base helm chart it is not possible,  And this was just a fast example.

A new example below:
How can I encode to base64 using secrets from vault? 
Spinnaker helm chart for example
https://github.com/helm/charts/blob/bd32c5248c21891d35576106f98d7b2a98fc952f/stable/spinnaker/values_saml.yaml#L36
``yaml
# env/values.yaml
spinnaker
   store:
     password: ref+vault://spinnaker/#password
```
With `secret` function
```yaml
# spinnaker.yaml.gotmpl
    password.txt: {{ .Values.spinnaker.store.password | secret | b64enc }}
````

The may problem that I have with the current implementation is that replacing of values is done outside template execution. So you can't do additional operation on secrets from vault etc.
Syntax suggests that manipulation is possible but implementation don't allow this.
IMO this PR is only extending what it is already in helmfile.

**Comment by mumoshu** [comment]
@aldor007 Thanks! OK, I'm convinced.

My next concern is that this may be a maintenance nightmare cuz I need to tell people when to use and not use this new function, as `ref+`s are automatically expanded if a ref is a whole value in for a specific key in a YAML hash. But that's another story... Hopefully, the helmfile community would help that. It's not a blocker.

Regarding the function name, I'd suggest `expandSecretRef` as that's how it works - expand a secret ref into a secret value. Otherwise, `fetchSecretValue` would be OK.

It shouldn't be `fetchSecret` as it doesn't get the secret itself. Rather it queries for a specific secret by key/ID of the secret and "returns the value" contained in the secret.

I'd also suggest not to name it like `getSecretValue`, as to me "fetch" sounds more like fetching from the remote secrets store. On the other hand, "get" rather sound like its reading values that are already there.

**Comment by mumoshu** [comment]
Also, if we go this way, I'd suggest providing an another function that named like `expandSecretRefs` as well, that takes a `map[string]interface{}` like `.Values`, and replaces refs included in all the values to corresponding secret values.

**Comment by aldor007** [comment]
Okay. 
I will prepare an update, later today (in about 10h)

**Comment by mumoshu** [comment]
If we're happy to add `expandSecretRefs` as well, I'd suggest we use `fetchSecretValue` for the single string arg variant. That way it would be very clear that they have different use-cases and abilities as their names are very much diferent.

**Comment by mumoshu** [comment]
@aldor007 Alright / No rush. Thank you so much for your efforts!

**Comment by aldor007** [comment]
@mumoshu I've update PR 
I hope I've fulfilled your requests. 
To help community to better understand this feature I've added some documentation :)

**Comment by DmytroKozlovskyi** [comment]
Hi, I faced, looks like, with the same issue today. I'm trying deploy userlist.txt for pgbouncer and have a variable in the template:

`userlist: '"db-user-name" "ref+vault://secret/data/my-service/{{ .Values.environment }}/ie?address={{ .Values.vaultAddress }}#/db_pgbouncer_password"'`

but when trying to do diff/apply I got the error `no value found for key db_pgbouncer_password"` because the last quote added to the vault URI. 
Does this PR fix this issue also?

**Comment by aldor007** [comment]
hi @DmytroKozlovskyi 
Yes, it should.

Below some pseudocode 
```yaml
userlist: '"db-user-name" " {{ printf "ref+vault://secret/data/my-service/%s/ie?address=%s#/db_pgbouncer_password"'  .Values.environment .Values.vaultAddress | fetchSecretValue }} 
```

PR #1226, opened 2020-04-26
Support for createNamespace

### LINKED ISSUE #891: Helm3 doesn't automatically create namespace
Helm3 doesn't automatically create namespace - see https://v3.helm.sh/docs/faq/#automatically-creating-namespaces

How can we solve this with helmfile, so that we don't have to manually create them?

### LINKED ISSUE #1140: Feature Request: Support for Upcoming Helm3 —create-namespace
## what

Next release of Helm will hopefully have support for automatic namespace creation. 

Would like this natively supported by Helmfile. 

## why

This is a pain point when migrating from helm2 to heln3. Using the flag restores the behavior of helm2.

## references 
https://github.com/helm/helm/pull/7648

cdunford:
- createNamespace is a new attribute that can be added to helmDefaults
  or an individual release to enforce the creation of a release namespace
  during sync if the namespace does not exist. This leverages helm's
  (3.2+) --create-namespace flag for the install/upgrade command. If
  running helm < 3.2, the createNamespace attribute has no effect.

Resolves #891
Resolves #1140

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -114,6 +114,8 @@ type HelmSpec struct {
 	CleanupOnFail bool `yaml:"cleanupOnFail,omitempty"`
 	// HistoryMax, limit the maximum number of revisions saved per release. Use 0 for no limit (default 10)
 	HistoryMax *int `yaml:"historyMax,omitempty"`
+	// CreateNamespace, when set to true (default), --create-namespace is passed to helm3 on install/upgrade (ignored for helm2)
+	CreateNamespace *bool `yaml:"createNamespace,omitempty"`
 
 	TLS       bool   `yaml:"tls"`
 	TLSCACert string `yaml:"tlsCACert,omitempty"`
@@ -158,6 +160,8 @@ type ReleaseSpec struct {
 	HistoryMax *int `yaml:"historyMax,omitempty"`
 	// Condition, when set, evaluate the mapping specified in this string to a boolean which decides whether or not to process the release
 	Condition string `yaml:"condition,omitempty"`
+	// CreateNamespace, when set to true (default), --create-namespace is passed to helm3 on install (ignored for helm2)
+	CreateNamespace *bool `yaml:"createNamespace,omitempty"`
 
 	// MissingFileHandler is set to either "Error" or "Warn". "Error" instructs helmfile to fail when unable to find a values or secrets file. When "Warn", it prints the file and continues.
 	// The default value for MissingFileHandler is "Error".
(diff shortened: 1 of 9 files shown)


pkg/state/state_test.go
```diff
@@ -584,10 +680,13 @@ func TestHelmState_flagsForUpgrade(t *testing.T) {
...
 			}
-			helm := helmexec.New("helm", logger, "default", &mockRunner{})
+			helm := &exectest.Helm{
+				Version: tt.version,
+			}
+
 			args, err := state.flagsForUpgrade(helm, tt.release, 0)
 			if err != nil {
-				t.Errorf("unexpected error flagsForUpgade: %v", err)
+				t.Errorf("unexpected error flagsForUpgrade: %v", err)
```
mumoshu, line 689: Thanks for the correction 👍

pkg/state/state.go
```diff
@@ -1635,6 +1639,12 @@ func (st *HelmState) flagsForUpgrade(helm helmexec.Interface, release *ReleaseSp
 		flags = append(flags, "--cleanup-on-fail")
 	}
 
+	if helm.IsVersionAtLeast(3, 2) &&
```
mumoshu, line 1642: Nice! We do need `createNamespace` to be valid and configurable only when Helm is 3.2.0 or later. Thanks for doing this properly and elegantly.

pkg/state/state.go
```diff
@@ -1635,6 +1639,12 @@ func (st *HelmState) flagsForUpgrade(helm helmexec.Interface, release *ReleaseSp
...
+		(release.CreateNamespace != nil && *release.CreateNamespace ||
+			release.CreateNamespace == nil && (st.HelmDefaults.CreateNamespace == nil || *st.HelmDefaults.CreateNamespace)) {
+		flags = append(flags, "--create-namespace")
+	}
```
mumoshu, line 1646: Can we add an error message for the user who tried to use createNamespace with a version of Helm that doesn't support it? Probably like this:

```suggestion
	if (release.CreateNamespace != nil && *release.CreateNamespace ||
		release.CreateNamespace == nil && (st.HelmDefaults.CreateNamespace == nil || *st.HelmDefaults.CreateNamespace)) {
		if !helm.IsVersionAtLeast(3, 2) {
			return nil, fmt.Errorf("releases[].createNamespace requires Helm 3.2.0 or greater")
		}
		flags = append(flags, "--create-namespace")
	}
```

Morriz:
Nice, this would also solve the issue of charts attempting to create a namespace for their own deployment (valid use case for helm3), like istio-operator. With the new flag we should be able to disable initial namespace creation, right?

Morriz:
Any idea in which release this will land?

mumoshu:
Please see the releases page!

allanlewis:
I don't have permissions to create a namespace, so I think this change is causing new deployments to fail for me: can we please make this optional? (My perms allow me to create anything in my current context's namespace, but not a namespace.)

mumoshu:
@allanlewis Helmfile does that by default because it has been the default behavior since Helm 2. To disable namespace creation, you can set `createNamespace: false`

allanlewis:
> To disable namespace creation, you can set `createNamespace: false`

Thanks! Would that be in `Chart.yaml`?

mumoshu:
Nope, helm's `--createNamespace` maps helmfile's `createNamespace` so you write it in helmfile.yaml

allanlewis:
> ...you write it in helmfile.yaml

Got it, thanks :+1:

PR #1227, opened 2020-04-26
Enhance createNamespace error handling

cdunford:
- An error is raised if createNamespace is set explicitly and the
  helm version being used is not 3.2+

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -1639,10 +1639,14 @@ func (st *HelmState) flagsForUpgrade(helm helmexec.Interface, release *ReleaseSp
 		flags = append(flags, "--cleanup-on-fail")
 	}
 
-	if helm.IsVersionAtLeast(3, 2) &&
-		(release.CreateNamespace != nil && *release.CreateNamespace ||
-			release.CreateNamespace == nil && (st.HelmDefaults.CreateNamespace == nil || *st.HelmDefaults.CreateNamespace)) {
-		flags = append(flags, "--create-namespace")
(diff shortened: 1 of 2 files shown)


pkg/state/state.go
```diff
@@ -1639,10 +1639,14 @@ func (st *HelmState) flagsForUpgrade(helm helmexec.Interface, release *ReleaseSp
...
 
-	if helm.IsVersionAtLeast(3, 2) &&
-		(release.CreateNamespace != nil && *release.CreateNamespace ||
-			release.CreateNamespace == nil && (st.HelmDefaults.CreateNamespace == nil || *st.HelmDefaults.CreateNamespace)) {
-		flags = append(flags, "--create-namespace")
+	if release.CreateNamespace != nil && *release.CreateNamespace ||
+		release.CreateNamespace == nil && (st.HelmDefaults.CreateNamespace == nil || *st.HelmDefaults.CreateNamespace) {
+		if helm.IsVersionAtLeast(3, 2) {
+			flags = append(flags, "--create-namespace")
+		} else if release.CreateNamespace != nil || st.HelmDefaults.CreateNamespace != nil {
```
mumoshu, line 1646: Great! I realized that the code I proposed in the previous PR was wrong cuz it consistently fails when the createNamespace is omitted AND Helm is pre 3.2. Yours is absolutely correct.

cdunford: Yeah I found that out pretty quickly 😅

frank-berlin: could be enhanced see #1246

PR #1233, opened 2020-04-29
Allow skipping chart-related operations when `installed: false`

### LINKED ISSUE #1232: When `installed` is `false` chart should not be fetch
Hello,

I'm facing a problem when I'm set the `installed` boolean to false because, helmfile still fetch the chart on the repository.

I have several environments in which I define the component versions. For some environments some components are not required and so I don't set any version and I would like to not install them, so I write a helmfile like that:

```yaml
release:
- name: releaseA
  chart: A
  installed: {{ not (empty .Values.chartAVersion }}
  version: {{ .Values.chartAVersion | default "noVersion" }}
```

But helmfile still wants to download chartA:noversion, which doesn't exists so it fails.
My question is why helmfile download the chart ? I think it's unecessary if `installed` is false since it will do nothing with it am I right ?

FYI I loop over all my helmfiles for each environments that's why I need something like that.
If you have another solution please give me an hint.

guillomep:
When install is false, we do not need to update the repositories and get
the chart.

Fix #1232

### DIFF
--- a/pkg/app/context.go
+++ b/pkg/app/context.go
@@ -17,6 +17,15 @@ func NewContext() Context {
 func (ctx Context) SyncReposOnce(st *state.HelmState, helm state.RepoUpdater) []error {
 	var errs []error
 
+	hasInstalled := false
+	for _, release := range st.Releases {
+		hasInstalled = hasInstalled && (release.Installed == nil || *release.Installed)
+	}
+
+	if !hasInstalled {
+		return errs
+	}
+
 	allUpdated := true
 	for _, r := range st.Repositories {
 		_, exists := ctx.updatedRepos[r.Name]
--- a/pkg/app/context.go
+++ b/pkg/app/context.go
@@ -19,7 +19,7 @@ func (ctx Context) SyncReposOnce(st *state.HelmState, helm state.RepoUpdater) []
 
 	hasInstalled := false
 	for _, release := range st.Releases {
-		hasInstalled = hasInstalled && (release.Installed == nil || *release.Installed)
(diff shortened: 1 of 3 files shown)


guillomep:
I don't understand the error in `ci/circleci: test` with `make pristine`

pkg/app/context.go
```diff
@@ -17,6 +17,15 @@ func NewContext() Context {
 func (ctx Context) SyncReposOnce(st *state.HelmState, helm state.RepoUpdater) []error {
 	var errs []error
 
+	hasInstalled := false
```
mumoshu, line 20: This should be set to `true` on init. Otherwise L22 never turns true

mumoshu: ...Which seems to be the case of the integration test failure

mumoshu:
> I don't understand the error in ci/circleci: test with make pristine

It's a code-format-related error. Try running `make fmt` and `git add -p && git commit` the changes so that it will resolve.

pkg/app/context.go
guillomep: Yes I made a mistake, I invert the condition

mumoshu:
@guillomep I've changed the PR title as actually it does work not only for repo update.

dudicoco:
@guillomep @mumoshu i'm running `helmfile template` and it fetches the charts which have `installed: false`. Any ideas?

guillomep:
In fact, I noticed that it does fetch the chart even with `installed: false` with the `template` command or the `lint` one

PR #1234, opened 2020-04-29
Correct ordering between releases and sub-helmfiles on destroy/delete

### LINKED ISSUE #1192: destroy ordering not reversed when mixing helmfiles and inline releases
Hi! Thanks for writing helmfile!

I've a `helmfile` that mixes inline `releases` and `helmfiles`, like so:

```yaml
helmfiles:
  - path: ./init.yaml

releases:
  - name: my-app
    ...
```

At `apply` time `init.yaml` releases are created first, and then `my-app` gets created.

At `destroy` time, the same order takes place.

I've seen https://github.com/roboll/helmfile/issues/797 which made me wonder if `destroy` should follow reverse order in this case too. Is this a bug?

nejtr0n:
This fix apply helmfiles in reversed order in destroy action.

Fixes #1192

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -475,9 +475,44 @@ func (a *App) visitStates(fileOrDir string, defOpts LoadOpts, converge func(*sta
 		}
 		st.Selectors = opts.Selectors
 
-		noMatchInHelmfiles, err = a.reverseOrder(d, f, st, opts, defOpts, converge, !opts.Reverse)
-		if err != nil {
-			return err
+		visitSubHelmfiles := func() error {
+			if len(st.Helmfiles) > 0 {
+				noMatchInSubHelmfiles := true
+				for i, m := range st.Helmfiles {
+					optsForNestedState := LoadOpts{
+						CalleePath:        filepath.Join(d, f),
+						Environment:       m.Environment,
+						Reverse:           defOpts.Reverse,
+						RetainValuesFiles: defOpts.RetainValuesFiles,
+					}
+					//assign parent selector to sub helm selector in legacy mode or do not inherit in experimental mode
+					if (m.Selectors == nil && !isExplicitSelectorInheritanceEnabled()) || m.SelectorsInherited {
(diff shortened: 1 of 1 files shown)


pkg/app/app.go
```diff
@@ -527,6 +507,41 @@ func (a *App) visitStates(fileOrDir string, defOpts LoadOpts, converge func(*sta
 	return nil
 }
 
+// apply reverse order in destroy action
+func (a *App) reverseOrder(d, f string, st *state.HelmState, opts, defOpts LoadOpts, converge func(*state.HelmState) (bool, []error), needReverse bool) (bool, error) {
```
mumoshu, line 511: I think this would better named `visitSubHelmfiles`

pkg/app/app.go
```diff
@@ -527,6 +507,41 @@ func (a *App) visitStates(fileOrDir string, defOpts LoadOpts, converge func(*sta
...
+	noMatchInHelmfiles := true
+	if len(st.Helmfiles) > 0 && needReverse {
```
mumoshu, line 513: Can we move this conditional block to the call-side?

Like

```
if !opts.Reverse {
	noMatchInHelmfiles, err = a.visitSubHelmfiles(d, f, st, opts, defOpts, converge)
	if err != nil {
		return err
	}
}

templated, tmplErr := st.ExecuteTemplates()

// snip

if !opts.Reverse {
	noMatchInHelmfiles, err = a.visitSubHelmfiles(d, f, st, opts, defOpts, converge)
	if err != nil {
		return err
	}
}

noMatchInHelmfiles = noMatchInHelmfiles && !processed

return context{app: a, st: templated, retainValues: defOpts.RetainValuesFiles}.clean(errs)
```

nejtr0n:
Rename function and move it to inline closure. All checks passed.
Tested it on 
```
helmfiles.d/
  01-first.yaml
  02-second.yaml

helmfile.yaml
releases:
  third
  fourth
```

Order was reversed to
```
fourth
third
second
first
```
(fourth and third run concurently)

Issue #1213, opened 2020-04-17
duplicate release "my-release" found in "": there were 5 releases named "my-release" matching specified selector


victornoel:
Hi,

Since helm 3, the release name is scoped to the namespace, so now in my helmfile, I have multiple releases with the same name but in different namespaces.

When I try to use helmfile apply or sync, I get the following error:
```
duplicate release "my-release" found in "": there were 5 releases named "my-release" matching specified selector
```

victornoel:
Using helmfile v0.106.2

victornoel:
Some more information:
- until then one can use `-l namespace=xxx` to deploy each release one by one
  - btw this `-l namespace` is not documented
- the diff works when doing `-i apply` but then it can't apply with the error above

mumoshu:
Ah good catch!

I think we missed adding proper Helm 3 support to Helm 3 around this duplication detection and DAG features.

For duplication detection I believe we need to use `r.Namespace` instead of the tiller namespace when Helm 3 is used (isHelm3 == true):
https://github.com/roboll/helmfile/blob/e0a793b7c5a8bb271ce2918cf1a0de30dac834a5/pkg/app/app.go#L704-L706

For the DAG feature(releases[].needs) we need to use the namespace instead of the tiller namespace here:

https://github.com/roboll/helmfile/blob/69feadc3603c8890a6bb02c2afd6c6b00bb3b720/pkg/state/state.go#L440-L456

mumoshu:
Would you mind submitting a PR to fix those two?

victornoel:
@mumoshu I will try :)

victornoel:
@mumoshu so I've started to work on that, sorry, I'm wrapping my head around how things are organised (and also Go ^^). I was wondering, about `releaseToID`, since it already use the value of `r.Namespace` to build the id, isn't that already enough to have unique ids? I ask because it's going to change a lot just to pass the `helm` object around to have it inside `releaseToID`, so if I can avoid it, the fix should be simpler.

mumoshu:
@victornoel Ah good catch! Yeah, if you could, I think reusing `ReleaseToID` would do the job.

https://github.com/roboll/helmfile/blob/d71ac1574809164e7bb053126745b156d490d991/pkg/state/state.go#L453

victornoel:
[...]

PR #1235, opened 2020-04-29
Use namespace for release unicity with helm3 (#1213)

victornoel:
This is for #1213 so that the releases are considered scoped to their namespace and not to tiller namespace.

Note the open question here: https://github.com/roboll/helmfile/issues/1213#issuecomment-621402678

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -682,7 +682,7 @@ func (a *App) visitStatesWithSelectorsAndRemoteSupport(fileOrDir string, converg
 	return a.visitStates(fileOrDir, opts, converge)
 }
 
-func processFilteredReleases(st *state.HelmState, converge func(st *state.HelmState) []error) (bool, []error) {
+func processFilteredReleases(st *state.HelmState, helm helmexec.Interface, converge func(st *state.HelmState) []error) (bool, []error) {
 	if len(st.Selectors) > 0 {
 		err := st.FilterReleases()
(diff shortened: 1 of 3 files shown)


mumoshu:
@victornoel Thanks! The change LGTM. Regarding the test failure, it's due to that there is now an extra call to helm(helm.IsHelm3) where it wasn't expected before this change.

This is where you can start:

https://github.com/roboll/helmfile/blob/eeb61e61746a5197e2ed722ff23c980a899ea534/pkg/app/app_test.go#L90

In `expectNoCallsToHelm` we currently create `&noCallHelmExec{}`, which panics on any call to helm.

Instead of `noCallHelmExec` which panics on any helm call, you should create a mock that respond to only `IsHelm3`, which is what you added in this PR.

You'd create a new struct for that, which would look like:

```
type versionOnlyHelmExec struct {
   isHelm3 bool
}

func (helm *versionOnlyHelmExec) IsHelm3() bool {
	helm.doPanic()
	return helm.isHelm3
}
```

So that you can use `&versionOnlyHelmExec{isHelm3: false}` instead of `&noCallHelmExec{}` for `TestVisitDesiredStatesWithReleasesFiltered_ReleaseOrder`.

victornoel:
@mumoshu I think I've managed to do something correct. I also added some tests :)

victornoel:
@mumoshu thanks for the help ^^

Issue #1104, opened 2020-02-18
Add support for helm upgrade --debug


desaintmartin:
When doing `helmfile --loglevel debug sync`, shouldn't we give the --debug switch to helm?

mumoshu:
Yeah that makes sense. How about `helmfile --debug` sets `--loglevel debug` for helmfile and enable helmfile to pass `--debug` to all the helm invocations?

desaintmartin:
Sure! Let me do it in a few days.

AlexsJones:
I've found the codebase quite difficult to find an entrypoint into this. I looked at adding it on the helmexec but the way arguments are appended seems to vary. 

My initial gut feeling was to pass this in as a field on the helm exec, but it's not immediately obvious how to wire this in.

Is there a way we can acces cli context in the helmexec more directly? 
```go
func configureLogging(c *cli.Context) error {
	// Valid levels:
	// https://github.com/uber-go/zap/blob/7e7e266a8dbce911a49554b945538c5b950196b8/zapcore/level.go#L126
	logLevel := c.GlobalString("log-level")
```

```go
func New(helmBinary string, logger *zap.SugaredLogger, kubeContext string, runner Runner, debug bool) *execer {
	execer := &execer{
		helmBinary:       helmBinary,
		isHelm3:          detectHelm3(helmBinary, logger, runner),
		logger:           logger,
		kubeContext:      kubeContext,
		runner:           runner,
		decryptedSecrets: make(map[string]*decryptedSecret),
	}

	if debug  {
		execer.SetExtraArgs("--debug")
	}

	return execer
}
```

mumoshu:
@AlexsJones Thanks for trying!

@machine424 added `--debug` to only enable passing `--debug` to helm commands in #1236. 

With that, you'll be able to use `helmfile --debug --log-level=debug SUBCOMMAND` for the desired behaviour.

If you could still try to add the feature to enable `log-level=debug` when `--debug` is provided, I believe you can just add the following three lines in `configureLogging`:

```
if c.GlobalBool("debug") {
	logLevel = "debug"
}
```

mumoshu:
Oops, sorry. Just realized that @machine424 has already done that in his PR!

Kirill-Garbar:
https://github.com/roboll/helmfile/issues/1289

PR #1236, opened 2020-04-30
feat: Add global `--debug` flag for passing `--debug` to helm

machine424:
Added a global flag `—debug`  that sets the logging level to debug and passes the global flag `—debug` to Helm through subcommands’ `—args`. See: https://github.com/roboll/helmfile/issues/1104

- Helm accepts the redundancy of the global flag `--debug`
- argparser.GetArgs gets rid of the redundancy which seems to be used to parse all —args
- what do you think about adding an Args field to `state.ReleaseSpec` ?
- what about adding a `Debug` (as we do for `—atomic` and `—wait` e.g.) into `state.HelmSpec` `state.ReleaseSpec` for more granularity ?
- sth still prevents the debug output from being displayed into stdout
- tests

### DIFF
--- a/main.go
+++ b/main.go
@@ -13,7 +13,6 @@ import (
 	"github.com/roboll/helmfile/pkg/state"
 	"github.com/urfave/cli"
 	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
 )
 
 var logger *zap.SugaredLogger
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -22,14 +21,11 @@ func configureLogging(c *cli.Context) error {
...
 	logLevel := c.GlobalString("log-level")
-	if c.GlobalBool("quiet") {
+	if c.GlobalBool("debug") {
+		logLevel = "debug"
+	} else if c.GlobalBool("quiet") {
 		logLevel = "warn"
 	}
-	var level zapcore.Level
-	err := level.Set(logLevel)
-	if err != nil {
```
machine424, line 30: `helmexec.NewLogger` handles this.

mumoshu:
> what do you think about adding an Args field to state.ReleaseSpec ?

Big no-no! I'm even wanting to deprecate `--args`. It's really a mess because we can't define which args should be passed to which helm commands.

> what about adding a Debug (as we do for —atomic and —wait e.g.) into state.HelmSpec state.ReleaseSpec for more granularity ?

Sounds not bad, but what's the exact use-case? 🤔

mumoshu:
@machine424 Anyway, I believe the primary goal of this PR - adding `--debug` to helmfile and its implementation - looks GREAT.

So can we just merge this and discuss about other bullet points in separate issues.

Would that be great for you as well?

PR #1245, opened 2020-05-06
Error when subhelmfiles do not exist

craigfurman:
This prevents typos in sub-helmfile paths from silently preventing operations from running.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,9 @@ helmfiles:
   # The nested-state file is locally checked-out along with the remote directory containing it.
   # Therefore all the local paths in the file are resolved relative to the file
   path: git::https://github.com/cloudposse/helmfiles.git@releases/kiam.yaml?ref=0.40.0
+# If set to "Error", return an error when a subhelmfile points to a
+# non-existent path. The default behavior is to print a warning and continue.
+missingFileHandler: Error
 
 #
 # Advanced Configuration: Environments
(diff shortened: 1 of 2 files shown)


craigfurman:
I verified that this does what it sets out to do using a local build, but I couldn't find an obvious place to add a unit test for this behavior. This is my first time contributing to helmfile, so perhaps I didn't look hard enough!

mumoshu:
@craigfurman Thanks! I remember that this is here for backward-compatibility since the very early days of Helmfile.

It'd be nice if you could take a look on #827 and #778 for the context.

I DEFINITELY agree that the current behaviour is not good, though!

Can we instead add a new option `missingFileHandler: Error` in the top-level of helmfile.yaml to switch to this behaviour?

If you could still try to contribute that, here's how:

- You'd add a new go struct field named `missingFileHandler string` to the `HelmState` struct
- In L1793 of the pkg/state/state.go you modified, you instead write something like:

```
if len(matches) == 0 {
	if st.MissingFileHandler == "Error" {
		return nil, fmt.Errorf("no matches for path: %s", hf.Path)
	}
	continue
}
```

craigfurman:
Thanks for looking @mumoshu!

It's unfortunate to introduce another flag, but I can see the value in preserving backwards compatibility.

I've updated this PR now - can you take another look?

PR #1288, opened 2020-05-30
Add integration test for Kustomize inetgration

ento:
### Context

I was working on a helmfile.yaml that applies json patches to a chart that uses `{{ .Release.Namespace }}` in its templates. ([kube-state-metrics to be precise](https://github.com/helm/charts/blob/ed7c211dc21f8f4435c802782a76ffdf4309e400/stable/kube-state-metrics/templates/_helpers.tpl#L45).)

### What happened
When I applied it, its value was set to the kube context that was active at the time. 

### Expected
As the helmfile.yaml had explicitly specified the namespace for the chart, I expected `{{ .Release.Namespace }}` to be resolved to that namespace, not the context's.

### Notes
I'm new to Go and have a feeling that storing a helper 'script' under `test/` and building it as an executable may not be considered best practice. So: please know that this is coming from a place of ignorance.

Also: will not force-push anymore after I mark this as ready for review (unless there's request to squash commits).

I'd be happy to split out the parts that add assertions on the result of `helmfile template` to a separate PR, if that'd be easier to review. I've kept the commits that show my journey through helm-diff, kustomize, and the official yaml package so that it's easier to follow why I ended up with the current implementation.

### DIFF
--- a/pkg/state/helmx.go
+++ b/pkg/state/helmx.go
@@ -31,6 +31,8 @@ func (st *HelmState) PrepareChartify(helm helmexec.Interface, release *ReleaseSp
 
 	opts.ChartVersion = release.Version
 
+	opts.Namespace = release.Namespace
+
 	dir := filepath.Join(st.basePath, release.Chart)
 	if stat, _ := os.Stat(dir); stat != nil && stat.IsDir() {
 		if exists, err := st.fileExists(filepath.Join(dir, "Chart.yaml")); err == nil && !exists {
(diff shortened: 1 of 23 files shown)


pkg/state/helmx.go
```diff
@@ -31,6 +31,8 @@ func (st *HelmState) PrepareChartify(helm helmexec.Interface, release *ReleaseSp
 
 	opts.ChartVersion = release.Version
 
+	opts.Namespace = release.Namespace
```
mumoshu, line 34: Thanks for your contribution!

FYI, this specific line is added via #1300 for fixing #1299, which is certainly the primary goal of this pull request.

I'll review other parts of this shortly!

PR #1354, opened 2020-07-13
Print command output in line

AClarkie:
This is an attempt to log helm exec output as the helm exec command is running. It only prints if the --debug flag is set.

I'd like to completely remove runner.Output() but it's used here:
https://github.com/roboll/helmfile/blob/cd427d9714924ed8fa21fc96a779a87c5b76de3c/pkg/tmpl/context_funcs.go#L101

Hoping someone can help with that.


pkg/helmexec/runner.go
```diff
@@ -34,7 +38,87 @@ func (shell ShellRunner) Execute(cmd string, args []string, env map[string]strin
...
+			}
+		}
+		wg.Done()
+	}()
+	go func() {
+		wg.Wait()
+		close(combined)
+	}()
+
+	for t := range combined {
```
mumoshu, line 97: I think this for-range loop needs to be within a dedicated go routine. Otherwise doesn't it block indefinitely?

`close(combined)` is required to finish this block, which requires the `scannerStdout` and `scannerStderr` to close, but to close those two scanners `c.Wait` needs to be run concurrently.

AClarkie: Hey thanks for the comments, no it doesn't block indefinitely but reads the output from the combined channel as its written to by the go routines that handle stdout and stderr..

I've tested the build binary on a few my local helmfile deployments and it's working well

mumoshu: My fault! I just saw `err = c.Start()` in L58 which invalidates my comment. Thanks for confirming

AClarkie:
> @AClarkie Code LGTM.
> 
> But do you have any specific reason not to change the original `Output` to print immediate debug logs like this?
> 
> I mean, if the only difference between Output and OutputInLine is ability to stream debug logs, I think I'm okay with just enhancing Output for that.

The original plan was to edit Output but it's being used on the following line:
https://github.com/roboll/helmfile/blob/cd427d9714924ed8fa21fc96a779a87c5b76de3c/pkg/tmpl/context_funcs.go#L101

And I wanted to clarify it's usage there and get a suggestion on how I should update it, is that something you can help with?

AClarkie:
Hey @mumoshu - any thoughts on my question?

mumoshu:
@AClarkie As long as `helmexec.Output` keeps returning either some captured stdout or an error object, that's okay. I think that's the only contact between helmfile/pkg/tmpl/context_funcs.go and the helmexec.Output func.

And in my understanding, your `OutputInLine` works exactly like that, except it is able to stream stdout/err via debug logs. Am I correct?

AClarkie:
> @AClarkie As long as `helmexec.Output` keeps returning either some captured stdout or an error object, that's okay. I think that's the only contact between helmfile/pkg/tmpl/context_funcs.go and the helmexec.Output func.
> 
> And in my understanding, your `OutputInLine` works exactly like that, except it is able to stream stdout/err via debug logs. Am I correct?

Yep that's correct

mumoshu:
@AClarkie Yeah I'm willing to revisit/merge this once I've finished stabilizing Helmfile v0.125.x!

mumoshu:
@AClarkie So I'll try to merge `OutputInLine` you've added to the existing `Output` today, based on our discussion above, and will merge this regardless of the result soon. Stay tuned!

mumoshu:
@AClarkie I reread your code and understood OutputInLIne to be almost equivalent to Output functionality-wise, except it writes to debug log.

In the last commit that I've pushed to your branch, I used `MultiWriter` in `Output` to also write to the debug log, so that we don't need to create a brand-new function.

I also added a prefix with an "execution ID" to every debug log message that is emitted, so that it becomes easier to deduce which log is from which command. It's not perfect at all. It should probably log the exact command being executed with the same prefix for correlation. But I defer it to another PR.

WDYT?

I'll merge this soon so that we can start testing it. But please feel free to submit another PR or send any feedback!

juniorz:
I am getting `segfaults`, and I believe they are due to this PR.

```
goroutine 66 [running]:
go.uber.org/zap.(*SugaredLogger).log(0x0, 0xff, 0x15ff253, 0x4, 0xc0005d3620, 0x2, 0x2, 0x0, 0x0, 0x0)
	/home/circleci/workspace/helmfile/vendor/go.uber.org/zap/sugar.go:221 +0x192
go.uber.org/zap.(*SugaredLogger).Debugf(...)
	/home/circleci/workspace/helmfile/vendor/go.uber.org/zap/sugar.go:133
github.com/roboll/helmfile/pkg/helmexec.(*logWriter).Write(0xc0000bf600, 0xc000288000, 0x5e, 0x8000, 0x5e, 0x0, 0x0)
	/home/circleci/workspace/helmfile/pkg/helmexec/log.go:25 +0x16c
io.(*multiWriter).Write(0xc0000bf620, 0xc000288000, 0x5e, 0x8000, 0x5e, 0x0, 0x0)
	/usr/local/go/src/io/multi.go:60 +0x87
io.copyBuffer(0x1b66740, 0xc0000bf620, 0x1b66c00, 0xc00016c050, 0xc000288000, 0x8000, 0x8000, 0x0, 0xc0005d37a8, 0x405b6f)
	/usr/local/go/src/io/io.go:407 +0x1fb
io.Copy(...)
	/usr/local/go/src/io/io.go:364
os/exec.(*Cmd).writerDescriptor.func1(0x165c698, 0xc0005d37c8)
	/usr/local/go/src/os/exec/exec.go:311 +0x63
os/exec.(*Cmd).Start.func1(0xc000033760, 0xc0000bf6a0)
	/usr/local/go/src/os/exec/exec.go:441 +0x27
created by os/exec.(*Cmd).Start
	/usr/local/go/src/os/exec/exec.go:440 +0x632
```

`logWriter.log` is `nil`: https://github.com/roboll/helmfile/blob/cd0ecc52638bbfbae93f148a5a0cdb2ad544dab8/pkg/helmexec/log.go#L25

Which I believe is due to a `nil` `ShellRunner.Logger`: https://github.com/roboll/helmfile/pull/1354/files#diff-0c6caf47bdc81e026dc633312456c1fdR40

juniorz:
This PR seems to assume the invariant `App.Logger` (used to initialize `ShellRunner.Logger`)  is non `nil`, however I could quickly find anything in the code that seems to enforce that: `conf.Logger()` might be `nil`, and `App.Init` does not do anything about this possibility.

mumoshu:
@juniorz Hey! Thanks for reporting. I think you're correct. But how can I reproduce your issue? I was thinking about why the e2e test have passed if it always panic.

mumoshu:
@juniorz Do you have any hooks in your helmfile.yaml? Then it's going to be fixed in #1445.

juniorz:
Thank you!

PR #1373, opened 2020-07-22
Add disableValidation and disableOpenAPIValidation per release

mumoshu:
`disableOpenAPIValidation: true` might be useful for workaround for broken CRDs that is known to be exist in older OpenShift versions, and `disableValidation: true` is confirmed to allow installing charts like prometheus-operator that tries to install CRDs and CRs in the same chart.

Strictly speaking, for the latter case I believe you only need `disableValidation: true` set during the first installation, but for the ease of operation I shall suggest you to always set it.

Obviously turning validation mostly(disableOpenAPIValidation) or entirely(disableValidation) result in deferring any real error until sync time. We need completely client-side validation that is able to read CRDs and use it for validating any CRs to catch any error before sync. But it worth an another (big) issue.

> Note that this feature requires helm-diff greater than or equal to v3.1.2

Fixes #1124

PR #1374, opened 2020-07-22
Add ability to use go-getter for fetching remote manifests directory as chart

mumoshu:
This, in combination with #1172, allows you to use `go-getter`-supported URL for K8s manifests on `chart`, so that Helmfile automatically fetches it and then turning it into a temporary local chart, which is then installed by Helmfile as similar as standard Helm charts.

An example usecase of this is to install cert-manager CRDs which is distributed separately from the chart:

```
releases:
- name: cert-manager-crds
  chart: git::http://github.com/jetstack/cert-manager.git@deploy/crds?ref=v0.15.2
```

I'm adding this based on discussion with @lukasmrtvy. He was trying to install cert-manager and prometheus-opreator with Helmfile, and this combined with #1373 should do the job. Thanks for the input!

### DIFF
--- a/pkg/app/app_test.go
+++ b/pkg/app/app_test.go
@@ -2373,16 +2373,16 @@ func TestTemplate_SingleStateFile(t *testing.T) {
 		"/path/to/helmfile.yaml": `
 releases:
 - name: myrelease1
-  chart: mychart1
+  chart: stable/mychart1
 - name: myrelease2
-  chart: mychart2
+  chart: stable/mychart2
 `,
 	}
 
 	var helm = &mockHelmExec{}
 	var wantReleases = []mockTemplates{
-		{name: "myrelease1", chart: "mychart1", flags: []string{"--namespace", "testNamespace", "--set", "foo=a", "--set", "bar=b", "--output-dir", "output/subdir/helmfile-[a-z0-9]{8}-myrelease1"}},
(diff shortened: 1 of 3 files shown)


osterman:
@mumoshu you're on a roll! this is awesome and fixes a big painpoint for us.

lukasmrtvy:
@mumoshu  

It will probably not work for cert-manager:
https://github.com/jetstack/cert-manager/blob/master/deploy/crds/crd-challenges.yaml#L6, due to internal templating..



Here is my workaround:

```
  hooks:
  - events: ["presync"]
    command: "/bin/sh"
    showlogs: true
    args: ["-c", "kubectl get --request-timeout='5m' -f https://github.com/jetstack/cert-manager/releases/download/{{`{{ .Release.Version }}`}}/cert-manager.crds.yaml || kubectl apply --request-timeout='5m' --validate=false -f https://github.com/jetstack/cert-manager/releases/download/{{`{{ .Release.Version }}`}}/cert-manager.crds.yaml"] 
  - events: ["postuninstall"]
    command: "/bin/sh"
    showlogs: true
    args: ["-c","kubectl delete --request-timeout='5m' --ignore-not-found -f https://github.com/jetstack/cert-manager/releases/download/{{`{{ .Release.Version }}`}}/cert-manager.crds.yaml"]
```
cavetas: https://cert-manager.io/docs/installation/upgrading/upgrading-0.15-0.16/#issue-with-older-versions-of-kubectl

Isnt it possible to use something like:
```
releases:
- name: cert-manager-crds
  chart: file::https://github.com/jetstack/cert-manager/releases/download/v0.16.1/cert-manager.crds.yaml
  forceGoGetter: true
```

?

PR #1390, opened 2020-08-01
Fix false-positive duplicate release with kubeContext

mumoshu:
Helmfile has been incorrectly showing releases with the same name but in different kubeContexts as duplicates. This fixes that.


mumoshu:
@jwitko Thanks for the feedback and the review ☺️

Issue #760, opened 2019-07-16
Unable to use .Release.Name in values.gotmpl


albertmichaelj:
I have a use case where I want the `.Release.Name` to be used as a variable in a `values.gotmpl` file. However, no matter what I do, the values.gotmpl file will not render `.Release.Name`. Is this expected behavior? Are there suggestions on how to work around this?

My use case is that I have multiple deployments of the same app on the same kubernetes cluster with non-namespaced objects (specifically PVs and filepaths on an NFS server) that need to be given unique names. I am using the release name as a unique identifier to add as a prefix to these non-namespaced objects. So, if you can't use `.Release.Name` in a gotmpl file, is there any other way to pass a unique value to the `gotmpl` file and still run a single `helmfile -f ... sync` command?

mumoshu:
@albertmichaelj Hey!

`Release.Name` is available only while rendering helmfile.yaml, not values file templates.

As a work-around, can you extract the part of your `values.gotmpl` into `helmfile.yaml` like:

```yaml
releases:
- name: yourapp
   chart: yourchart
   values:
   - pathOnNfs: /path/to{{.Release.Name}}
```

If not, I think this is a valid feature request, as I see no downside of improving helmfile to allow `.Release.Name` within `values.gotmpl`.

albertmichaelj:
@mumoshu Thanks for the response, and it's really helpful to know that I wasn't just doing something stupid. If it's not a major change, I think being able to specify some kind of unique value that can be propagated to the `values.gotmpl` file would be very, very helpful for my use case. The problem is that I (for various reasons) need to specify `pathOnNfs` multiple places in the `values.yaml` file, and while I could conceivably set the values a bunch of times in the helmfile, that would clutter things up quickly.

This doesn't necessarily need to be the `Release.Name`. It could be something in the release like 
```yaml
releases:
- name: yourapp
    chart: yourchart
    environment:
      - customValue: "whatever"
    values:
      - values.gotmpl
```
and then in `values.gotmpl` you could have something like:
```yaml
foo: {{ .Environment.Custom.Values.customValue }}
```
This would be more flexible in that you could always set `customValue` to `Release.Name` using double rendering in the helmfile. The big need for me, is that I need someway to specify per release values that need to be repeated several times in the `values.gotmpl` for a variety of reasons (generally because I'm using an existing chart and I am trying to avoid forking the chart).

There isn't any functionality currently like this in helmfile, right?

[...]

PR #1424, opened 2020-08-21
feat: Allow .Release.Name to be used in gotmpl values templates

joshes:
This adds the ability to utilize `.Release` inside of gotmpl files as discussed [here](https://github.com/roboll/helmfile/issues/760).

Resolves: https://github.com/roboll/helmfile/issues/760

Added a simple test that passed once implemented and all tests are running green.

### DIFF
--- a/pkg/state/types.go
+++ b/pkg/state/types.go
@@ -17,6 +17,8 @@ type EnvironmentTemplateData struct {
 	Namespace string
 	// Values is accessible as `.Values` and it contains default state values overrode by environment values and override values.
 	Values map[string]interface{}
+	// Release is accessible as `.Release` from any template expression executed by the renderer
+	Release ReleaseSpec
 }
 
 // releaseTemplateData provides variables accessible while executing golang text/template expressions in releases of a helmfile YAML file
--- a/pkg/state/types.go
+++ b/pkg/state/types.go
@@ -17,16 +17,30 @@ type EnvironmentTemplateData struct {
 	Namespace string
 	// Values is accessible as `.Values` and it contains default state values overrode by environment values and override values.
 	Values map[string]interface{}
-	// Release is accessible as `.Release` from any template expression executed by the renderer
-	Release ReleaseSpec
 }
 
-// releaseTemplateData provides variables accessible while executing golang text/template expressions in releases of a helmfile YAML file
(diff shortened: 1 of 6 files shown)


Romiko:
Hi, is this something that will be merged in soon?

mumoshu:
@Romiko Yes. I'll merge this today or tomorrow, immediately after I've finished stabilizing the latest minor version of Helmfile (v0.125.x).

@joshes Thanks for your contribution! This LGTM overall. But please let me modify this before/after merging this a bit, so that it doesn't accidentally expose other ReleaseSpec fields.

For example, this change as its current form result in exposing `Release.Values` to the template. It sounds useful initially, but in reality, it can contain either paths to values file OR map[string]interface{} of embedded values, which makes it hard for people to actually use. I'd like to discuss about it in a dedicated feature request(if someone submits it). Same for other ReleaseSpec fields.

joshes:
@mumoshu Sounds good. That side effect did strike me as extremely useful as well initially but I hear what you're saying. :)

Thanks!

pkg/state/types.go
```diff
@@ -19,12 +19,28 @@ type EnvironmentTemplateData struct {
 	Values map[string]interface{}
 }
 
-// releaseTemplateData provides variables accessible while executing golang text/template expressions in releases of a helmfile YAML file
+// releaseTemplateData provides variables accessible while executing golang text/template expressions in release templates
+// and release values templates within a Helmfile YAML file.
 type releaseTemplateData struct {
```
mumoshu, line 24: @joshes I realized that we already have a struct type for the template data used for rendering anything related to each release! So I refactored your change without modifying EnvironmentTemplateData.

Along the way I also realized that we already and unexpectedly exposed ReleaseSpec to the whole template in some case.

Fortunately, Helmfile had unit tests([`TestHelmState_executeTemplates`](https://github.com/roboll/helmfile/blob/3f02b866401ba7ef34c4c985de13151f3e51fb7b/pkg/state/state_exec_tmpl_test.go#L25)) that helped me to figure out which ReleaseSpec fields are intended to be exposed to the template. I extracted the fields to releseTemplateDataRelease(L34-L46).

mumoshu (approved):
@joshes Thanks a lot for implementing this! I especially appreciate the addition to the unit test, which helped me easily refactor this ☺️

PR #1430, opened 2020-08-25
Update to the latest version of helm-diff plugin

benmathews:
This is needed to support the --disable-validation flag added into
helmfile


mumoshu (approved):
Ah, good catch! Thanks for your contribution @benmathews ☺️

PR #1431, opened 2020-08-26
Fix regression in helmfile deps

### LINKED ISSUE #1421: Unable to add new release
When a lock file exists, and does not contain an entry for the new release, `helmfile deps` fails with `[no resolved dependency found for "new-release"]`. This regression was seemingly introduced in [v0.125.3](https://github.com/roboll/helmfile/compare/v0.125.2...v0.125.3).

mumoshu:
Fixes #1421


yurrriq:
Thanks!

PR #1436, opened 2020-08-28
feat: `helmfile build --embed-values` to embed release values and secrets into the output

mumoshu:
This is handy when debugging complex `helmfile.yaml` and `values.yaml.gotmpl` as the resulting `helmfile.yaml` contains all the rendered values and sops-encrypted secrets in its embedded form.

It's gonna be used to resolve a terraform-provider-helmfile issue that requires a unique hash of the desired state: https://github.com/mumoshu/terraform-provider-helmfile/issues/28

Environment values and secrets are not yet supported. They are output as-is. But it should be fine as the resulting helm release won't change as long as release values/secrets are unchanged.

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -209,7 +209,7 @@ type ReleaseSpec struct {
 	Namespace string            `yaml:"namespace,omitempty"`
 	Labels    map[string]string `yaml:"labels,omitempty"`
 	Values    []interface{}     `yaml:"values,omitempty"`
-	Secrets   []string          `yaml:"secrets,omitempty"`
+	Secrets   []interface{}     `yaml:"secrets,omitempty"`
 	SetValues []SetValue        `yaml:"set,omitempty"`
 
 	ValuesTemplate    []interface{} `yaml:"valuesTemplate,omitempty"`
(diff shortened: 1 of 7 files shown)


mnatan:
>Environment values and secrets are not yet supported. They are output as-is. But it should be fine as the resulting helm release won't change as long as release values/secrets are unchanged.

I don't think it's a problem for the terraform plugin, as it's only used for cache SHA calculation and is not printed anywhere, right?

pkg/state/state.go
```diff
@@ -2447,3 +2477,42 @@ func (st *HelmState) ToYaml() (string, error) {
...
+			paths, skip, err := st.storage().resolveFile(missingFileHandler, "values", pathPrefix+t)
+			if err != nil {
+				return nil, err
+			}
+			if skip {
+				continue
+			}
+
+			if len(paths) > 1 {
+				return nil, fmt.Errorf("glob patterns in release values and secrets is not supported yet. please submit a feature request if necessary")
```
mnatan, line 2498: I like your error handling 😄

mumoshu:
> I don't think it's a problem for the terraform plugin, as it's only used for the diff cache SHA calculation and is not printed anywhere, right?

@mnatan Thanks for reviewing ❤️ Yes, I think you're correct.

Issue #1168, opened 2020-04-02
Environment values not overriding 


michaelajr:
Something has broken in the last release around env values. I use to be able to do this:

```
environments:
  default:
    missingFileHandler: Error
    values:
    - values-file-1.yaml
    - values-file-2.yaml
```

Where `values-file-2.yaml` had some values that would override values in `values-file-1.yaml`. This is no longer working. It appears `values-file-2.yaml`  is ignored (although I am not sure). It only has a single value override.

Also `missingFileHandler: Error` no longer work. It will not error on missing files.

Sorry if I missed a breaking change.

mumoshu:
@michaelajr Hey! I couldn't reproduce your issue with the following example:

```
$ tail *
==> helmfile.yaml <==
  default:
    missingFileHandler: Error
    values:
    - values.1.yaml
    - values.2.yaml

releases:
- name: example
  values:
  - {{ .Values | toYaml | nindent 4 }}

==> values.1.yaml <==
foo: FOO

==> values.2.yaml <==
bar: BAR
```

```
$ helmfile build
---
#  Source: helmfile.yaml

filepath: helmfile.yaml
helmBinary: helm
environments:
  default:
    values:
    - values.1.yaml
    - values.2.yaml
    missingFileHandler: Error
releases:
- name: example
  values:
  - bar: BAR
    foo: FOO
templates: {}
```

Could you clarify a bit more?

We did fix a bug that you couldn't override a value if it was nil as per #1153 and #1162.

Would it affect your use-case?

michaelajr:
Hi. Thanks for the quick response. Here's our real world example. 

* We keep a list of charts to install in `addons.yaml`
* Each cluster brings overrides - `addons2.yaml` in this example

```sh
$ tail -20 *
==> addons.yaml <==
addons:
  mychart:
    skip: true
    name: mychart
    namespace: kube-system
    chart: stable/mychart
    version: 1.0.0

==> addons2.yaml <==
addons:
  mychart:
    skip: false

==> helmfile.yaml <==
environments:
  default:
    missingFileHandler: Error
    values:
    - addons.yaml
    - addons2.yaml
releases:
{{- range .Values.addons }}
{{- if not .skip }}
- name: {{ .name }}
  namespace: {{ .namespace }}
  chart: {{ .chart }}
  version: {{ .version }}
{{- end }}
{{- end }}
```

[...]

PR #1456, opened 2020-09-03
Bump sprig to v3.1.0 and mergo 3.11

jalkjaer:
I have explictly added mergo 0.3.11 as 0.3.9 is considered broke (see https://github.com/imdario/mergo#important-note , and https://github.com/Masterminds/sprig#important-notes )

The other `go.mod` changes happended automagically - I assume that is a `go mod vendor` doing that

Ref https://github.com/roboll/helmfile/issues/1168#issuecomment-683345472

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.6.3 // indirect
 	github.com/hashicorp/go-version v1.2.0
 	github.com/howeyc/gopass v0.0.0-20190910152052-7cb4b85ec19c // indirect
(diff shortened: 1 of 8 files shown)


mumoshu (approved):
@jalkjaer Thanks for the PR! I've added a commit to make the test passing.

It seems like the latest version of mergo, v0.3.11, has a regression that resulted in mergo trying to merge unexported(private) fields. The mergo doc says it ignores unexported fields so I consider this as a regression on mergo side. The commit I've added work-around that by trying use mergo only for merging go struct that has exported fields only.

Thanks again for your support @jalkjaer!

Issue #1461, opened 2020-09-07
[Feature request] Load environment variables from .env files


tonsV2:
The support for reqiredEnv is really handy but it would be even more so if environment variables found in .env files was included.

Docker compose has fairly advanced support for .env files. But less could do it. Simply including variables found in .env would be a good start. Please see the below link for more information.

https://docs.docker.com/compose/environment-variables/

vsimon:
just a suggestion, for in the meantime, there is [direnv](direnv.net) + this alias gets .env files loaded at the time the helmfile command is run
`alias di='echo dotenv > .envrc && touch .env && direnv allow'`

tonsV2:
@vsimon - Thanks for your suggestion

Minor note, isn't the touch command of the alias redundant?

direnv seems like a great project but it does seem like a minor hassel to use "plain" .env files. Not only would you have to make a symlink. Eg. ln -s .env .envrc but every environment variable needs to be prefixed with `export` which differs from the way it's done with other applications.

Furthermore, having support for this method of loading environment variables would not require direnv to be installed in a pipeline. And down the road I'd like to see this implemented supporting different environments.

tonsV2:
https://github.com/direnv/direnv/issues/674

vsimon:
Yes it deals with .envrc files and the need to export every variable in the .envrc file was the reason why I mentioned the bootstrapping alias (done once per folder). The `dotenv` function is a built-in direnv function that automatically sources an .env file in the same folder (an .env file without the export prefixes).

tonsV2:
@vsimon - I see! I read dotenv as direnv. I understand the alias now. Thanks :-)

PR #1462, opened 2020-09-07
Initial support for environment variables read from .env

tonsV2:
This #1461 led to this pull request.

Not only is this my first contribution in Go this is also the very first Go code that I've written so please tell me what I've done wrong.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/hashicorp/go-version v1.2.0
 	github.com/howeyc/gopass v0.0.0-20190910152052-7cb4b85ec19c // indirect
 	github.com/imdario/mergo v0.3.11
+	github.com/joho/godotenv v1.3.0
 	github.com/konsorten/go-windows-terminal-sequences v1.0.2 // indirect
 	github.com/pierrec/lz4 v2.3.0+incompatible // indirect
 	github.com/r3labs/diff v0.0.0-20190801153147-a71de73c46ad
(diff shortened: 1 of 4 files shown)


woodcockjosh:
What is the status of this? I would like to have this feature

woodcockjosh:
Is there anyway we can load environment variable from `.env` files specified for example in `environments` list? I think that environment variables could be different based on the environment. For example `SOPS_PGP_FP` would have a different set of public keys to control the access to the secrets based on the environment. Same as docker-compose.

vpoyraz:
They way this could be worked around for the time being is to source the .env file and then in your helmfile.yaml:

```bash
...
templates:
  default: &default
    values:
      - values.yaml
      - extraArgs:
          dryRun: {{ env "DRY_RUN" }}
          logLevel: {{ env "LOG_LEVEL" }}
releases:
- name: my-release-name
  chart: my-chart
  <<: *default
```
Here I have shortened the example but, basically we can override values.yaml entries `extraArgs.dryRun` and `extraArgs.logLevel` during runtime:

```bash
source .env && helmfile apply
```

The values.yaml file would look like this:

```yaml
extraArgs: {}
  # dryRun: true
  # logLevel: true
```

mumoshu:
Sorry for the radio silence. I was revisiting all the pending PRs today and finally had a chance to revisit this.

This is mostly good, but what makes me hesitant to merge this is that this seems to automatically import envvars if there's `.env`, and you have no way to opt-out of that behavior.

Shouldn't this be an opt-in feature instead? For example, `docker run` accepts `--env-file` for basically the same thing.

tonsV2:
@mumoshu - That's a good point. Do you have any suggestions on how to opt in?

mumoshu:
@tonsV2 My idea is to add a string-file for `--env-file` like https://github.com/roboll/helmfile/blob/87e5903705925df8298c3d490350d7ddfc100c35/main.go#L46-L50, enhance the config struct to have a new function for getting the `--env-file` value like https://github.com/roboll/helmfile/blob/87e5903705925df8298c3d490350d7ddfc100c35/main.go#L869-L871, load environment values or propagate it down to the template functions factory `func (c *Context) newTemplate() *template.Template {` and godot.load there, so that your template has access to `RequiredEng` and `GetEnv` functions that made aware of envvars from the env file.

PR #1465, opened 2020-09-09
Adding hellofresh as a helmfile user

harshal-shah:


mumoshu (approved):
Great to hear! Thank you for using Helmfile ☺️

PR #1466, opened 2020-09-10
Re-add Release.Namespace in release values.yaml templates

### LINKED ISSUE #1464: Evaluating {{ .Namespace }} in a gotmpl file no longer works since v0.126.0
I have a line in one of my values files for a release like this:
```
  service:
    local:
      annotations:
        metallb.universe.tf/allow-shared-ip: {{ .Namespace }}
```
It gives this error:
```
ailed to render values files "myapp.yaml.gotmpl": failed to render [myapp.yaml.gotmpl], because of template: stringTemplate:18:48: executing "stringTemplate" at <.Namespace>: can't evaluate field Namespace in type state.releaseTemplateData
```

I checked out v0.125.0 and v0.125.9 and they don't have this issue. It must be something that changed in v0.126.0. It is also present in v0.128.0.

Maybe https://github.com/roboll/helmfile/pull/1415 is the source of the regression?

mumoshu:
It was accidentally removed in #1424, and had been unexpectedly unavailable between v0.126.0 and v0.128.1.

Fixes #1464

### DIFF
--- a/pkg/state/create_test.go
+++ b/pkg/state/create_test.go
@@ -101,10 +101,14 @@ bar: {{ readFile "bar.txt" }}
 
 	valuesFile := "/example/path/to/values.yaml.gotmpl"
 	valuesContent := []byte(`env: {{ .Environment.Name }}
-releaseName: {{ .Release.Name }}`)
+releaseName: {{ .Release.Name }}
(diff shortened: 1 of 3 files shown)


mattymo:
Thanks!

Issue #1470, opened 2020-09-11
Temporary directory generation does not take into account kubeContext uniqueness


jwitko:
There is currently an issue when deploying with releases that use dynamic values for kubeContext.  Helmfile attempts to generate a temporary directory path that takes into account all things that make a release unqiue into the dir path naming.  It does not account for KubeContext and therefore when doing things like a `lint` or `template` on dynamic releases helmfile will attempt to unpack remote charts into the same directories causing issues with overwriting files/permissions.

```
STDERR:
  Error: failed to untar: a file or directory with the name /tmp/341109564/infra-monitoring/prometheus-operator/bitnami/prometheus-operator/0.31.0/prometheus-operator already exists
```

PR #1471, opened 2020-09-11
Append KubeContext to temporary directory path

jwitko:
Fix bug where KubeContext is not taken into account in temporary directories.

Fixes https://github.com/roboll/helmfile/issues/1470


jwitko:
I was able to build a binary with these changes and have tested locally, it fixed my issue.  Also did `make clean` and `make tests` and both went without issue.

mumoshu (approved):
@jwitko This makes sense. Thank you so much for locating/fixing it 👍 

@voron Thank you also for your support in our slack 😄

jwitko:
Happy to help!  Ty for review/merge.

PR #1472, opened 2020-09-11
Update README.md - small change to Global Hooks example  

vugardzhamalov:
I've made a small change to Global Hooks example in order to remove release name reference - this was probably copy/pasted from Hooks section. As I understand release name can't be referenced in Global Hooks section.
I am not sure if there is any open issue that might be relevant to this PR.


mumoshu (approved):
Good catch! Thanks for the fix @vugardzhamalov ☺️

PR #1488, opened 2020-09-18
Support the latest Helm (>=v3.3.2) and bump the Helm version in Docker image.

wi1dcard:
Bump Helm version to the latest as it contains several security fixes.

Ref:

- <https://github.com/helm/helm/releases/tag/v3.3.2>
- <https://github.com/helm/helm/releases/tag/v2.16.11>

### DIFF
--- a/pkg/helmexec/exec.go
+++ b/pkg/helmexec/exec.go
@@ -129,7 +129,10 @@ func (helm *execer) AddRepo(name, repository, cafile, certfile, keyfile, usernam
 		helm.logger.Infof("empty field name\n")
 		return fmt.Errorf("empty field name")
 	}
-	args = append(args, "repo", "add", "--force-update", name, repository)
+	args = append(args, "repo", "add", name, repository)
+	if helm.IsHelm3() {
+		args = append(args, "--force-update")
+	}
 	if certfile != "" && keyfile != "" {
 		args = append(args, "--cert-file", certfile, "--key-file", keyfile)
 	}
--- a/pkg/helmexec/exec.go
+++ b/pkg/helmexec/exec.go
@@ -130,7 +130,7 @@ func (helm *execer) AddRepo(name, repository, cafile, certfile, keyfile, usernam
 		return fmt.Errorf("empty field name")
 	}
 	args = append(args, "repo", "add", name, repository)
-	if helm.IsHelm3() {
+	if helm.IsHelm3() && helm.IsVersionAtLeast(3, 3, 2) {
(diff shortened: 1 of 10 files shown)


wi1dcard:
Please DO NOT MERGE THIS now. The breaking changes described in the release notes of helm impact helmfile while updating existing repos:

> One breaking change was made: The default behavior of helm repo add was changed: It no longer overwrites repositories by default. The flag --force-update was added to helm repo add to allow a repo to be overwritten. The flag --no-update still exists, but does nothing. This change accords with the Helm policy for allowing minor breaking changes in the name of security.

I'm going to fix it.

mumoshu (approved):
LGTM. Thanks a lot for your contribution @wi1dcard! ☺️

pkg/helmexec/exec.go
```diff
@@ -130,6 +130,9 @@ func (helm *execer) AddRepo(name, repository, cafile, certfile, keyfile, usernam
 		return fmt.Errorf("empty field name")
 	}
 	args = append(args, "repo", "add", name, repository)
+	if helm.IsHelm3() && helm.IsVersionAtLeast(3, 3, 2) {
```
RothAndrew, line 133: Recommend adding another method `IsVersionAtMost` and adding `&& helm.IsVersionAtMost(3, 3, 3)` to this if statement. [A PR is being worked](https://github.com/helm/helm/pull/8772) to let the previous idempotent functionality return as long as the name and url are identical to what they were before.

The security issues that Helm worked to fix with this change are valid. [They are listed in the 3.3.2 release notes](https://github.com/helm/helm/releases/tag/v3.3.2)

My golang fu is not strong, but I can work on doing a PR as it seems to be a small change.

RothAndrew: This would require an addition to the documentation that changing the URL of an existing named repo will cause an error, but IMO that's a good thing.

RothAndrew: Actually, I'm seeing another issue. I frequently use the helm git plugin and pin to a specific tag or ref in the git repo. Changing that ref would be considered a change in the URL, which would throw the error.

Thoughts?

voron: Git ref changing is content changing, generally speaking, thus throwing an error is a valid option. Looks like we need `-force-update` configurable from helmfile to shift a decision to user side.

wi1dcard: See #1494.

PR #1490, opened 2020-09-18
Drop duplicated entry in extra functions list from readme

andrew-demb:


mumoshu (approved):
Good catch! Thanks @andrew-demb ☺️

PR #1493, opened 2020-09-21
Fix helm version checking

tyrken:
Noticed this incorrect algorithm while browsing the changelog.  Should really use the built in semver package, but my golang-fu wasn't up to making that big a change...


mumoshu:
@tyrken Hey! This helped me realize our test suite wasn't sufficient to catch it, and resulted in #1490 that uses the semver library for parsing it in a more robust way.
I'm closing this as included in #1490. Thanks again for your contribution ☺️

Issue #1489, opened 2020-09-18
`helmfile repos` attempts to re-add already added repo


bskim45:
Hi, I just updated helmfile from `v0.128.0` to `v0.129.3` and strange behavior happens when I try to attempt repo-related actions. Am I missing something?

`helmfile` version: `0.129.3`
`helm` version: `3.3.2`

helmfile.yaml
```
repositories:
- name: stable
  url: https://kubernetes-charts.storage.googleapis.com

...
```

output:
```
$ helmfile repos                                                                                                       
Adding repo stable https://kubernetes-charts.storage.googleapis.com
in ./helmfile.yaml: command "/usr/local/bin/helm" exited with non-zero status:

PATH:
  /usr/local/bin/helm

ARGS:
  0: helm (4 bytes)
  1: repo (4 bytes)
  2: add (3 bytes)
  3: stable (6 bytes)
  4: https://kubernetes-charts.storage.googleapis.com (48 bytes)

ERROR:
  exit status 1

EXIT STATUS
  1

STDERR:
  Error: repository name (stable) already exists, please specify a different name

COMBINED OUTPUT:
  Error: repository name (stable) already exists, please specify a different name
```

mumoshu:
@bskim45 Recently Helmfile started to rely on helm's behaviour that `helm repo add`ing the existing repo updates it(`helm repo up` but for the specific repo). Which version of Helm are you using?

bskim45:
Just found out this is due to the breaking change in recent helm release `v3.3.2`: https://github.com/helm/helm/releases/tag/v3.3.2

Commands like `deps` and `sync` are also affected to this change. Should we have to make some changes according to the recent helm update?

mumoshu:
@bskim45 Thanks. Good to know. Yeah maybe we'd need to add `--force-update` for v3.3.2 or greater for our use-case.

I couldn't find any info for changes on `deps`  and `sync`. Would you mind sharing that?

bskim45:
Nothing special. What I meant to say is that `deps` and `sync` commands are also failing since `repo` operation also called in those two.

mumoshu:
@bskim45 Ah, I see! Thanks for clarifying.

Xtigyro:
+1

Xtigyro:
> 
> 
> @bskim45 Thanks. Good to know. Yeah maybe we'd need to add `--force-update` for v3.3.2 or greater for our use-case.
> 
[...]

PR #1494, opened 2020-09-22
Append --force-update for specific helm versions.

wi1dcard:
This PR contains 1 bug fix and 1 change:

- Fix the helm version comparison. As @tyrken mentioned in #1493, I'm trying to use the semver package to parse, compare, and process the helm version string.
- Use option `--force-update` in `helm repo add`, ONLY when the Helm version is `>= 3.3.2` and `< 3.3.4` (ref #1489 and helm/helm#8777).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.14
 
 require (
 	github.com/Azure/azure-sdk-for-go v35.0.0+incompatible // indirect
-	github.com/Masterminds/semver v1.4.2
+	github.com/Masterminds/semver/v3 v3.1.0
 	github.com/Masterminds/sprig/v3 v3.1.0
 	github.com/aryann/difflib v0.0.0-20170710044230-e206f873d14a
 	github.com/go-test/deep v1.0.3
--- a/go.sum
+++ b/go.sum
@@ -181,6 +181,7 @@ github.com/bgentry/go-netrc v0.0.0-20140422174119-9fd32a8b3d3d/go.mod h1:6QX/PXZ
 github.com/bgentry/speakeasy v0.1.0/go.mod h1:+zsyZBPWlz7T6j88CTgSN5bM796AkVf0kBD4zp0CCIs=
 github.com/bitly/go-hostpool v0.0.0-20171023180738-a3a6125de932 h1:mXoPYz/Ul5HYEDvkta6I8/rnYM5gSdSV2tJ6XbZuEtY=
 github.com/bitly/go-hostpool v0.0.0-20171023180738-a3a6125de932/go.mod h1:NOuUCSz6Q9T7+igc/hlvDOUdtWKryOrtFyIVABv/p7k=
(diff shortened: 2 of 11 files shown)


voron:
As @RothAndrew mentioned [here](https://github.com/roboll/helmfile/pull/1488#discussion_r492061666), it may be a valid case to use `--force-update` with helm versions 3.3.4+ too, when you're really need to update repository via helmfile. My propose is to get following syntax in helmfile:
```
helmDefaults:
  repoForceUpdate: false
...
repositories:
  - name: git-based-repo
    url: git+ssh://git@github.com/roboll/helmfile@?ref=<FREQUENTLY_CHANGED_OR_DYNAMIC_REF>
    forceUpdate: true
```
WDYT?
We don't need this option with helm <3.3.2, but with 3.3.2+ (and 3.3.4+ too) it may be better to let helmfile user decide what to do with it.

wi1dcard:
Hi @voron , I didn't realized that we can use a URL with the specific ref of version. Hmmm, I think that would be a great case of showing why we need to leave the option and let the users decide. What about the default value of `repositories.*.forceUpdate`? We may have 3 choices:

1. The default value of `repositories.*.forceUpdate` is always false.
2. The default value is determined based on the Helm version, `>=3.3.2 && <3.3.4` -> `true`, otherwise -> `false`.
3. The default value is determined based on the Helm version, `>=3.3.2` -> `true`, otherwise -> `false`.
4. The default value of `repositories.*.forceUpdate` is `helmDefaults.repoForceUpdate`, and the default value of latter one can be one of above.

Personally I prefer `4th + 1st` or just the third one. Because:

With `4th + 1st`, we keep the default behavior as secure as possible without adding any additional flags, and the users who use the specific helm versions can simplely change the behavior for all repos in one place `helmDefaults.repoForceUpdate`.

With the third, we can add the `--force-update` under the hood and do not have to "break" anything. Users won't notice that there's any change either using helm < 3.3.2 or helm > 3.3.4.

What do you think?

cc @mumoshu

voron:
I vote for `4th+1st`:
```
The default value of repositories.*.forceUpdate is helmDefaults.repoForceUpdate
The default value of helmDefaults.repoForceUpdate is false
```
 It's `helm` "responsibility" to break this thing, I don't see reasons to hide it from helmfile user. We need to provide an ability to adjust this option when required, w/o any "smart" staff IMHO.

mumoshu:
I vote for 4th + 1st, too. I thought 4th + 2nd or 3rd would be the best before seeing https://github.com/helm/helm/pull/8777 - apparently the Helm team "fixed" the regression issue very quickly 👍

mumoshu (approved):
This LGTM as its current state and I'm merging this as it only enhances the current implementation. Thank you so much for your continued effort @wi1dcard!

I'll leave the implementation of the ability to expose `--force-update` configurable to another PR(s).

PR #1498, opened 2020-09-24
feat: Upgrade kubectl, upgrade helm

naphta:
As per the [version skew of kubectl](https://kubernetes.io/docs/setup/release/version-skew-policy/#kubectl) I picked v1.18.9 as to support both the latest stable and as many older versions as possible.

I upgraded Helm because using the latest release I was getting strange issues with helm-diff being unable to unmarshal YAML which I couldn't replicate locally (on the same versions of helmfile, helm, and plugins), but could within the Docker release. Upgrading Helm to the latest patch fixed it for me.

Changes:

* Bump helm to v3.3.4.
* Bump kubectl to v1.18.9
* Update Dockerfile to be a little more verbose to help understand which command in a chain failed.

Signed-off-by: Jake Hill <jake@naphta.uk>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,11 +11,12 @@ FROM alpine:3.11
 
 RUN apk add --no-cache ca-certificates git bash curl jq
 
-ARG HELM_VERSION=v2.16.12
+ARG HELM_VERSION="v2.16.12"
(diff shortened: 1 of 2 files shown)


mumoshu (approved):
@wi1dcard's #1546 introduces the proper support for helm 3.3.4+ so this is good to go now. Thanks for your contribution @naphta!

PR #1500, opened 2020-09-25
Clarify namespace prefix in needs

baurmatt:
Before this, the documentation didn't mention that the release needs to
be prefixed with the namespace if the release had a namespace configured.


mumoshu (approved):
LGTM. Thanks for your contribution, @baurmatt ☺️

PR #1510, opened 2020-10-01
Fix: forceNamespace needs shouldRun is true

mdaisuke:
Hi,

I found `releases[].forceNamespace` doesn't work okay, so tried to fix it.

## how it works

```sh
$ cat <<EOF > sample-helmfile.yaml
releases:
- name: scaler
  forceNamespace: this-is-forced-namespace
  chart: stable/cluster-autoscaler
EOF

$ helmfile -f sample-helmfile.yaml template
```
It's gonna insert `metadata.namespace` with value `forceNamespace` if a chart has some manifest templates which doesn't have `metadata.namespace`.

cc @mumoshu


mumoshu (approved):
LGTM. Thanks a lot for your contribution, @mdaisuke! ☺️

PR #1516, opened 2020-10-08
Add SettleMint to the USERS list

roderik:
The SettleMint platform allows enterprises to spin up k8s clusters and deploy production-grade blockchain networks and additional services. Helmfile is in charge of deploying these networks and services on demand out of the self-service management UI.


mumoshu (approved):
Great to hear! Thanks for using Helmfile in production @roderik 😄

PR #1526, opened 2020-10-11
Support for azure acr helm repositories

javiplx:
This adds some basic support for helm repositories hosted on azure container registry (not OCI but [classic ones](https://docs.microsoft.com/en-us/cli/azure/acr/helm?view=azure-cli-latest). Add a new element to RepositorySpec to state that is externally managed and runs the az-cli command instead of the helm one.

### DIFF
--- a/pkg/app/app_test.go
+++ b/pkg/app/app_test.go
@@ -2418,7 +2418,7 @@ func (helm *mockHelmExec) SetExtraArgs(args ...string) {
 func (helm *mockHelmExec) SetHelmBinary(bin string) {
 	return
 }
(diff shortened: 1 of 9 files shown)


mumoshu (approved):
LGTM! Would you mind rebasing this onto the latest commit, though?

javiplx:
Just to solve conflict? In that case I prefer merge from upstream. Any problema with that?

mumoshu:
@javiplx That's fine, too!

javiplx:
Error is connection related while building image, but I have no permission to retry.

mumoshu:
Triggered a rerun of the failed job.

PR #1535, opened 2020-10-13
Update USERS.md

Stono:
Adding AT as a User :-) 
Thanks for your OSS efforts!


mumoshu (approved):
@Stono Thanks a lot for using Helmfile ☺️

PR #1542, opened 2020-10-16
Add helmDefaults.repoForceUpdate and repositories[].forceUpdate configs.

wi1dcard:
Add `helmDefaults.repoForceUpdate` and `repositories[].forceUpdate` in helmfile.yaml. With this PR, users can specify whether they'd like to enable `--force-update` globally, individually, or just leave it disabled (by default).

Related PRs or issues:

- <https://github.com/roboll/helmfile/pull/1494>
- <https://github.com/roboll/helmfile/issues/1489>
- <https://github.com/helm/helm/pull/8777>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -153,7 +153,7 @@ jobs:
     - run:
         name: Install helm
         environment:
-          HELM_VERSION: v3.3.3
+          HELM_VERSION: v3.3.4
         command: |
           HELM_FILENAME="helm-${HELM_VERSION}-linux-amd64.tar.gz"
           curl -Lo ${HELM_FILENAME} "https://get.helm.sh/${HELM_FILENAME}"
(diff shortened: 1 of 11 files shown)


wi1dcard:
Found a problem here: helm seems not to update repos without the `--force-update` option, even though the URL specified in helmfile.yaml and existing helm repos are the same:

```
Adding repo stable https://kubernetes-charts.storage.googleapis.com/
"stable" already exists with the same configuration, skipping

Adding repo gitlab https://charts.gitlab.io
"gitlab" already exists with the same configuration, skipping
```

I don't see any good solution for now. Either we add the option `--force-update` (which could leave potential vulnerabilities), or don't add `--force-update` (which let helm just ignore adding or updating the repo if it already exists) is not the great option I think.

Probably we need to perform a `helm repo update` once `helm repo add`s are done. Is there any better idea?

wi1dcard:
This PR should be able to pick up and implement now as Helm has added `helm repo update <repo_name>`. I'll rebase it to the latest later and finish it as soon as I'm available.

yxxhero:
see: https://github.com/roboll/helmfile/issues/2148

PR #1543, opened 2020-10-16
fix: README improvement

isindir:
Documentation improvement - took some time for me to realise that `devel` should be used instead of `args` to enable pulling development versions of charts from harbor:

- Fixing README


isindir:
Fixing issue #1544 - can't link these together other than via this comment

mumoshu (approved):
LGTM. Thanks for the addition @isindir ☺️

PR #1549, opened 2020-10-19
mention Arch Linux support

shibumi:
Hi,
I have pushed helmfile to the official Arch Linux repositories. Thanks for your work, I use it in production :)


mumoshu (approved):
Glad to hear! Thanks for your support @shibumi ☺️

PR #1564, opened 2020-11-02
Fix `helmfile --selector x=y template` with `needs`

### LINKED ISSUE #1552: Rendering Helm chart template for a single release, ignoring `needs`
Is it possible to run `helmfile template --selector name=x` to output the template for a release which has `needs`?

I updated to `0.132.0` but still can't get around `in helmfile.yaml: "x" depends on nonexistent release "y"` 😿

I'm trying to run some tests like `kubeval` and `kube-score` on the "final" release output as the YAML can be quite different based on the values in the Helmfile if the chart has conditionals in the template, so it would be much more useful to let Helmfile do its magic on it rather than just using the output of the vanilla chart with `helm template`.

mumoshu:
I had mistakenly inverted the necessary flag to turn off the legacy processing on selectors that should be used only for helmfile commands that does not support DAGs/needs.

Fixes #1552

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -233,7 +233,7 @@ func (a *App) Template(c TemplateConfigProvider) error {
 		}
 
 		return
-	}, SetFilter(true))
+	})
 }
 
 func (a *App) WriteValues(c WriteValuesConfigProvider) error {


daaain:
Hey @mumoshu, I'd love to test this, but not sure how to install from an unmerged PR 😳

mumoshu:
@daaain Hey! I've released v0.132.2 for this. Would you mind giving it a shot?

daaain:
Works perfectly (tried 1.133.0 now that it's the latest), thanks @mumoshu 🙇

PR #1580, opened 2020-11-10
fix: support arm binaries

### LINKED ISSUE #1433: ARM / RPi support
Hi Roboll,

I wondered if you'd considered adding ARM / RPi support to helmfile? I'm asking as we've had [a request from the arcade community](https://github.com/alexellis/arkade/pull/194) to make helmfile downloadable via `arkade get` and I noticed you have some multi-arch binaries, but nothing for armhf (RPi) or ARM64 (server-grade)?

Would you consider adding it? Is there a reason that hasn't been suggested before? (I checked the issue tracker)

Alex

jstrachan:
fixes #1433

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ integration:
 .PHONY: integration
 
 cross:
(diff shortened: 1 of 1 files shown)


mumoshu (approved):
Thanks for the fix @jstrachan ☺️

PR #1582, opened 2020-11-11
Update readme stable/incubator URLs

bycEEE:
Small update to the readme repository examples since Google is no longer officially supporting hosting the stable/incubator helm charts. The official helm docs are using these repos now, eg. https://helm.sh/docs/intro/quickstart/ and https://github.com/helm/charts.


mumoshu (approved):
LGTM. Thanks for the fix @bycEEE ☺️

PR #1584, opened 2020-11-12
Fix openSUSE repo URL.

manfred-h:
Apologies for not getting it right the first time...


mumoshu (approved):
LGTM. Thanks for the fix @manfred-h!

PR #1587, opened 2020-11-15
Fix incorrect help text of some CLI options.

wi1dcard:
Some of the commands show the incorrect options help text, for example `helmfile deps -h`:

```
...

OPTIONS:
   --args value                   pass args to helm exec
   --skip-repos helm repo update  skip running helm repo update before running `helm dependency build`
```

The trailing `helm repo update` after `--skip-repos` shouldn't be there.

The reason is that the CLI library `urfave/cli` has a feature called "placeholder values". If we backquote a string in the option usage, it will be considered as a "placeholder" for the value of that option.

This PR corrects the help text by replacing the backquotes with double-quotes:

```
...

OPTIONS:
   --args value  pass args to helm exec
   --skip-repos  skip running "helm repo update" before running "helm dependency build"
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -54,7 +54,7 @@ func main() {
 		},
 		cli.StringFlag{
 			Name:  "environment, e",
-			Usage: "specify the environment name. defaults to `default`",
+			Usage: `specify the environment name. defaults to "default"`,
 		},
 		cli.StringSliceFlag{
 			Name:  "state-values-set",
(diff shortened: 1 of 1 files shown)


mumoshu (approved):
LGTM. Thanks for the fix @wi1dcard!

// I'll also review your another PR soon 🙏

PR #1588, opened 2020-11-15
Fix the logic of helmfile deps and add tests.

wi1dcard:
## Context

`helmfile.yaml`:

```yaml
repositories:
  - name: bitnami
    url: https://charts.bitnami.com/bitnami/

releases:
  - name: example # this chart contains a dep
    chart: ./charts/example
```

`./charts/example/Chart.yaml`:

```yaml
apiVersion: v2
name: example
description: A Helm chart for Kubernetes

type: application
version: 0.1.0
appVersion: 1.16.0

dependencies:
  - name: redis
    version: ^10.7.0
    repository: alias:bitnami
```

When I run `helmfile deps --skip-repos` with the helmfile.yaml above, it skips updating the deps of the local chart `example`:

```
$ helmfile deps --skip-repos
There are no repositories defined in your helmfile.yaml.
This means helmfile cannot update your dependencies or create a lock file.
See https://github.com/roboll/helmfile/issues/878 for more information.
```

What I was expecting is that helmfile updates all the dependencies of the chart, like what helmfiles does without the option `--skip-repos` (but skips adding repos of course):

```
$ helmfile deps
Adding repo bitnami https://charts.bitnami.com/bitnami/
"bitnami" has been added to your repositories

Building dependency release=darwinia-bridge, chart=charts/darwinia-bridge
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "bitnami" chart repository
Update Complete. ⎈Happy Helming!⎈
Saving 1 charts
Downloading redis from repo https://charts.bitnami.com/bitnami/
Deleting outdated charts

There are no repositories defined in your helmfile.yaml.
This means helmfile cannot update your dependencies or create a lock file.
See https://github.com/roboll/helmfile/issues/878 for more information.
```

## The Problem

While I was digging into this issue, I found that the problem is not caused by `--skip-repos` itself.

In fact, logics in `UpdateDeps`:

<https://github.com/roboll/helmfile/blob/f6bf885fb76d71e4008601532aabd2841c4f3b20/pkg/state/state.go#L1928-L1937>

will never get executed, as `release.Chart` is already "normalized path", which means, it won't be considered as a local chart anymore. Here is where the chart get "normalized":

<https://github.com/roboll/helmfile/blob/f6bf885fb76d71e4008601532aabd2841c4f3b20/pkg/app/run.go#L62-L74>

`releaseToChart` contains the paths proceeded by `normalizeChart(st.basePath, chartPath)`, and `rel.Chart = chart` overrides the original chart paths in releases.

In order to fix that, I've decided not to break the current behavior of `PrepareCharts` but change the logic of determining if it's a local chart.

Firstly, I changed `if isLocalChart(release.Chart)` to `if pathExists(release.Chart)`. However, I changed it again to the func `DirectoryExistsAt` to make it mockable in unit tests. 

### Add `SkipDeps` in struct `ChartPrepareOptions`

`PrepareCharts` skips executing `helm dep build` when `SkipRepos` is true:

<https://github.com/roboll/helmfile/blob/f6bf885fb76d71e4008601532aabd2841c4f3b20/pkg/state/state.go#L943-L948>

<https://github.com/roboll/helmfile/blob/f6bf885fb76d71e4008601532aabd2841c4f3b20/pkg/state/state.go#L1086-L1090>

Therefore, if I run `helmfile deps --skip-repos`:

1. PrepareCharts not executing `st.runHelmDepBuilds`. ✅
2. Then execute `UpdateDeps`. ✅

If run `helmfile deps`:

1. PrepareCharts executing `st.runHelmDepBuilds`. ❌ (Unnecessary)
2. Then execute `UpdateDeps`. ✅

PrepareCharts shouldn't use `SkipRepos` as the condition of running `helm dep build` or not. So I added another flag `SkipDeps` in struct to control whether PrepareCharts needs to build the deps.

### Helper functions in testfs

<https://github.com/roboll/helmfile/blob/f6bf885fb76d71e4008601532aabd2841c4f3b20/pkg/testhelper/testfs.go#L54-L62>

Determining absolute paths with `strings.Contains` sounds not ideal. Changed to `strings.HasPrefix(path, "/")`.

<https://github.com/roboll/helmfile/blob/f6bf885fb76d71e4008601532aabd2841c4f3b20/pkg/testhelper/testfs.go#L64-L74>

`path[0] == '/'` can lead to nil pointer error with empty string `path`. Changed to `strings.HasPrefix(path, "/")`.

### Add and fix tests

- Move several test cases from `TestHelmState_UpdateDeps` to `Test_normalizeChart`.
- Mock the fs and add test cases in `TestHelmState_UpdateDeps`.
- Add `TestDeps` in `pkg/app/app_test.go`.
- Fix tests in `pkg/remote/remote_test.go` with respect to the changes of `testfs`.
- Output logs when skipping updating deps for remote chart.

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -105,6 +105,7 @@ func (a *App) Deps(c DepsConfigProvider) error {
 	return a.ForEachState(func(run *Run) (_ bool, errs []error) {
 		prepErr := run.withPreparedCharts("deps", state.ChartPrepareOptions{
 			SkipRepos:   c.SkipRepos(),
+			SkipDeps:    true,
 			SkipResolve: true,
 		}, func() {
 			errs = run.Deps(c)
@@ -122,7 +123,10 @@ func (a *App) Repos(c ReposConfigProvider) error {
 	return a.ForEachState(func(run *Run) (_ bool, errs []error) {
 		var reposErr error
 
-		err := run.withPreparedCharts("repos", state.ChartPrepareOptions{SkipRepos: true}, func() {
+		err := run.withPreparedCharts("repos", state.ChartPrepareOptions{
+			SkipRepos: true,
+			SkipDeps:  true,
+		}, func() {
 			reposErr = run.Repos(c)
 		})
 
@@ -140,7 +144,10 @@ func (a *App) Repos(c ReposConfigProvider) error {
 
 func (a *App) DeprecatedSyncCharts(c DeprecatedChartsConfigProvider) error {
 	return a.ForEachState(func(run *Run) (_ bool, errs []error) {
-		err := run.withPreparedCharts("charts", state.ChartPrepareOptions{SkipRepos: true}, func() {
+		err := run.withPreparedCharts("charts", state.ChartPrepareOptions{
+			SkipRepos: true,
+			SkipDeps:  true,
+		}, func() {
 			errs = run.DeprecatedSyncCharts(c)
 		})
 
@@ -166,7 +173,10 @@ func (a *App) Diff(c DiffConfigProvider) error {
 
 		var errs []error
 
-		prepErr := run.withPreparedCharts("diff", state.ChartPrepareOptions{SkipRepos: c.SkipDeps()}, func() {
+		prepErr := run.withPreparedCharts("diff", state.ChartPrepareOptions{
+			SkipRepos: c.SkipDeps(),
+			SkipDeps:  c.SkipDeps(),
+		}, func() {
 			msg, matched, affected, errs = a.diff(run, c)
 		})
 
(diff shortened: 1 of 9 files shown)


mumoshu (approved):
LGTM. Thank you so much for your continued effort @wi1dcard ☺️ 

> Firstly, I changed if isLocalChart(release.Chart) to if pathExists(release.Chart). However, I changed it again to the func DirectoryExistsAt to make it mockable in unit tests.

Note: I initially thought this can break it when `chart` referenced a local tarball e.g. `chart: envoy-1.9.3.tgz`. But after seeingin L974 in pkg/state/state.go, the move to `directoryExistsAt` might be a good thing, as you can't run `helm dep build` on a tarball, obviously.. 😃

PR #1594, opened 2020-11-20
Don't append --tiller-namespace flag when running helm3

jayme-github:
When `helmDefaults.tillerNamespace` is set, helmfile will append `--tiller-namespace` to the helm flags regardless of the helm version.


mumoshu (approved):
@jayme-github Ah, thanks for fixing this! LGTM 👍

PR #1604, opened 2020-11-26
fix: allow helmfiles to be corrected marshalled to YAML so they can be manipulated by other programs

chrismellard:


mumoshu (approved):
@chrismellard Thanks for the fix! LGTM.

BTW, I had no much time to make Helmfile easily consumable as a Go library so far. As you seem to have experience in it, I'd appreciate any feedback(s) from you! (like extracting APIs into its own pkg like `api`

Issue #1134, opened 2020-03-04
What's the best practice to work with helmfile inside the container


jizusun:
I really love the out-of-the-box docker images at https://quay.io/roboll/helmfile
And the Dockerfiles: https://github.com/roboll/helmfile/blob/master/Dockerfile.helm3

My approach is to mount my project folder (or only the helm chart folder) into the container (from my `Makefile`):

```makefile
helmfile-install: 
  docker run --rm -it \
	-v $(PROJECT_ROOT_DIR):$(PROJECT_ROOT_DIR) \
	-w `pwd` \
	-v $$HOME/.kube:$$HOME/.kube \
	-e KUBECONFIG=$${KUBECONFIG} \
        quay.io/roboll/helmfile:helm3-v0.102.0 helmfile -f charts/helmfile.yaml apply
```

However, the user in the container is `root`, so after `helmfile apply`, i will get `*.tgz` with permissions `user: root, group: root` under the sub-folder. To work with it out of the container, i have to use `sudo`.

Is it better if we add a non-root user in the `Dockerfile`?
Like the following:

```
ARG username=user
RUN addgroup -g 1000 ${username} \
    && adduser -u 1000 -G ${username} -s /bin/sh -D ${username}
USER ${username}
```

I would like to create a pull request if any maintainer agree with me 😸 
Thanks.

mumoshu:
@jizusun Hey! I have never tried to use the helmfile image that way, but I've experimented a bit and came up with this working command:

```
docker run -it --rm -u "$(id -u):$(id -g)" -v "$(pwd):$(pwd)" -w "$(pwd)" quay.io/roboll/helmfile:v0.106.1 bash -c "echo foo > foo.txt"
```

So without modifying the image, it seems like `docker run -u UID:GID` helps altering the owner uid/gid.

Would this work for you, too?

jizusun:
> @jizusun Hey! I have never tried to use the helmfile image that way, but I've experimented a bit and came up with this working command:
> 
> ```
> docker run -it --rm -u "$(id -u):$(id -g)" -v "$(pwd):$(pwd)" -w "$(pwd)" quay.io/roboll/helmfile:v0.106.1 bash -c "echo foo > foo.txt"
> ```
> 
> So without modifying the image, it seems like `docker run -u UID:GID` helps altering the owner uid/gid.
> 
> Would this work for you, too?

Oh thank you @mumoshu  ❤️ 

With the command you mentioned the file permission is not a problem any more.
[...]

PR #1607, opened 2020-12-01
Add documentation on how to use the docker image

felipecrs:
References #1134

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,4 +46,6 @@ RUN helm plugin install https://github.com/databus23/helm-diff && \
 
 COPY --from=builder /workspace/helmfile/dist/helmfile_linux_amd64 /usr/local/bin/helmfile
 
-CMD ["/usr/local/bin/helmfile"]
+WORKDIR /wd
+
+CMD ["/usr/local/bin/helmfile", "--help"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,6 +46,4 @@ RUN helm plugin install https://github.com/databus23/helm-diff && \
 
 COPY --from=builder /workspace/helmfile/dist/helmfile_linux_amd64 /usr/local/bin/helmfile
 
-WORKDIR /wd
-
 CMD ["/usr/local/bin/helmfile", "--help"]
--- a/README.md
+++ b/README.md
@@ -377,23 +377,23 @@ The [Helmfile Docker images are available in Quay](https://quay.io/roboll/helmfi
 
 ```sh-session
 # helm 2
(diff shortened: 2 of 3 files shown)


felipecrs:
Do we need really to mount the `helm` configuration (`~/.helm` for v2 and `~/.config/helm` for v3)? Don't the `helmfile.yaml` itself already contain all the needed configuration?

Dockerfile
```diff
@@ -46,4 +46,6 @@ RUN helm plugin install https://github.com/databus23/helm-diff && \
 
 COPY --from=builder /workspace/helmfile/dist/helmfile_linux_amd64 /usr/local/bin/helmfile
 
-CMD ["/usr/local/bin/helmfile"]
+WORKDIR /wd
```
mumoshu, line 49: This makes sense, but I'd like to keep this `WORKDIR /`, as I've seen someone uses the image like:

```
docker -v ${PWD}/helmfile.yaml:/helmfile.yaml ...
```

And this change can be a breaking change for those.

README.md
```diff
@@ -365,12 +365,40 @@ If you wish to treat your enviroment variables as strings always, even if they a
...
 - Windows (using [scoop](https://scoop.sh/)): `scoop install helmfile`
 - macOS (using [homebrew](https://brew.sh/)): `brew install helmfile`
 
+### Running as a container
+
+The [Helmfile Docker images are available in Quay](https://quay.io/roboll/helmfile). There is no `latest` tag, since the `0.x` versions can contain breaking changes, so make sure you pick the right tag. Example using `helmfile 0.135.0`:
+
+```sh-session
+# helm 2
+$ docker run --rm --net=host -v "${HOME}/.kube:/root/.kube" -v "${HOME}/.helm:/root/.helm" -v "${PWD}:/wd" quay.io/roboll/helmfile:v0.135.0 helmfile sync
```
mumoshu, line 380:
```suggestion
$ docker run --rm --net=host -v "${HOME}/.kube:/root/.kube" -v "${HOME}/.helm:/root/.helm" -v "${PWD}:/wd" --workdir /wd quay.io/roboll/helmfile:v0.135.0 helmfile sync
```

This does make sense, but as I've commented at https://github.com/roboll/helmfile/pull/1607/files#r540608083, I'd like to use `/` as the default workdir and explicitly set `--workdir /wd` for this "recommended" use-case.

I ABSOLUTELY agree we should have a great default, but I believe the image is already used widely in production and our users won't love breaking changes :)

Dockerfile
felipecrs: Indeed this is a breaking change for some. However, helmfile supports referencing multiple helmfiles, and mounting one by one can be tricky.

README.md
felipecrs: Makes sense, to prevent a breaking change. Perhaps in the future. :)

mumoshu (approved):
LGTM. Thank you so much for your contribution @felipecrs!

PR #1612, opened 2020-12-04
fix: support arm64 images

jstrachan:


mumoshu (approved):
Thanks for the addition @jstrachan! ☺️

PR #1614, opened 2020-12-07
Fix selector syntax in example

lifelofranco:
The current example in the documentation for using the selector flag doesn't work. It looks like the order matters.

`helmfile sync --selector job=dbmigrator` gives me `Incorrect Usage: flag provided but not defined: -selector`.


mumoshu (approved):
Yes, the order matters. Thanks for the fix @lifelofranco!

PR #1616, opened 2020-12-09
chore: suppress empty values on serializing several helmfile values

chrismellard:
To keep a more minimalist helmfile when we're serializing our helmfiles to from programs to disk it would be nice to keep them with a trimmed set of fields


mumoshu (approved):
LGTM. Thanks for the improvement @chrismellard ☺️

PR #1626, opened 2020-12-21
Include kustomize binary in helm3-based Docker image

dalbani:
In order to use the newly included Kustomize Transformers functionality (#1592), I propose to include the `kustomize` binary in the Docker image.
I understand that it will make the image size significantly larger, but is that a problem?
The alternative would be to provide another flavor of the image, e.g. `helm3-kustomize-v0.135.0`, but is it worth the hassle?

### DIFF
--- a/Dockerfile.helm3
+++ b/Dockerfile.helm3
@@ -36,6 +36,15 @@ RUN set -x && \
     chmod +x kubectl && \
     mv kubectl /usr/local/bin/kubectl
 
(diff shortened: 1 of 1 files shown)


mumoshu (approved):
Thanks for your contribution! Yeah, I believe this makes sense. Let's try and wait until anyone complains.

// I can imagine that anyone may be using the helmfile image as the base for their custom image and not wanting the official image to contain anything that they don't use(i.e. kustomize). But we can't verify that without releasing and waiting for complaints :)

PR #1629, opened 2020-12-23
feat: added in oci repository flag and added helm methods to pull and export charts

chrismellard:
added features to allow helmfile users to work with OCI compliant registries.


pkg/helmexec/exec.go
```diff
@@ -157,6 +157,13 @@ func (helm *execer) UpdateRepo() error {
 	return err
 }
 
+func (helm *execer) RegistryLogin(name string, username string, password string) error {
+	helm.logger.Info("Logging in to registry")
+	out, err := helm.exec([]string{"registry", "login", name, "--username", username, "--password", password}, map[string]string{})
```
chrismellard, line 162: Going to need to move this off the command line and use std-in before we merge

chrismellard:
@mumoshu ready for your review on this now when you have the time 👍

varunpalekar:
@mumoshu Please help us on PR review

adrianharabula:
@mumoshu please help us with this, default install of jx3 on azure, cannot be used without this.

mumoshu (approved):
LGTM. Thanks everyone for patience. Awesome work @chrismellard!

pkg/state/state.go
```diff
@@ -957,7 +1007,7 @@ func (st *HelmState) PrepareCharts(helm helmexec.Interface, dir string, concurre
 				skipDepsGlobal := opts.SkipDeps
 				skipDepsRelease := release.SkipDeps != nil && *release.SkipDeps
 				skipDepsDefault := release.SkipDeps == nil && st.HelmDefaults.SkipDeps
-				skipDeps := !isLocal || skipDepsGlobal || skipDepsRelease || skipDepsDefault
+				skipDeps := !isLocal || skipDepsGlobal || skipDepsRelease || skipDepsDefault || !isOCI
```
nauxliu, line 1010: Hi @chrismellard Why skip deps for non-OCI releases? I'm running into some issues with this change, I'm not familiar with OCI, can you explain your intention? Thank you!

mumoshu:
In #1667, I'm adding a few fixes on top of this OCI repo support. I'd appreciate it if you could also review it!

pkg/state/state.go
mumoshu: @nauxliu I think we should invert it, so that dep updates are skipped when the chart is from an OCI repo(i.e. it's a remote chart). The fix is included in #1667 and I'd appreciate if you could review it!

PR #1635, opened 2020-12-30
docs: update helm command name

caleb15:
helm delete is now helm uninstall. See https://helm.sh/docs/helm/helm_delete/


mumoshu (approved):
LGTM. Thanks for the fix @caleb15!

PR #1641, opened 2021-01-07
Fix typo "verion"

muffl0n:


mumoshu (approved):
Thanks for the fix @muffl0n ☺️

PR #1644, opened 2021-01-08
Bump variantdev/vals to 0.12.0

mumoshu:
For Azure KeyVault support for the fix for AWS SSM provider.


Morriz:
Hey @mumoshu may I ask if this is related to encryption? That is delegated to helm-secrets right?

PR #1661, opened 2021-01-28
fix: fixed up OCI repos so they support paths and URLs that differ from chart names contained there

chrismellard:


mumoshu (approved):
Thanks @chrismellard!

mumoshu:
@chrismellard I'm adding a few fixes on top of the OCI repo support via #1667. I'd appreciate it if you could also review it!

PR #1662, opened 2021-01-28
fix: dont sent RegistryLogin password via args

cameronbraid:
Password is already being sent via stdin, so change the args to use password from stdin, and don't send password via args


mumoshu (approved):
LGTM. Thanks for the fix @cameronbraid ☺️

PR #1667, opened 2021-02-02
Fix OCI support

mumoshu:
I have seen various issues related to the OCI repository support recently added to Helmfile.
This is the patch that should fix all the issues Im aware of until now.

If you've seen issues on helmfile v0.138.0x, I'd appreciate it if you could help testing this.

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -986,6 +986,8 @@ func (st *HelmState) PrepareCharts(helm helmexec.Interface, dir string, concurre
 				}
 				chartFetchedByGoGetter := chartPath != chartName
 
+				var isOCI bool
+
 				if !chartFetchedByGoGetter {
 					ociChartPath, err := st.getOCIChart(release, dir, helm)
 					if err != nil {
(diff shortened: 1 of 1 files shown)


mumoshu:
Anyone knows an example chart that is stored in a public OCI repository which is good for use within our integration test suite?

chrismellard (approved):
lgtm

chrismellard:
Not aware of any public OCI repositories. But if this merges we can quickly test out over in jenkins-x where we're working with private OCI registries there. Not ideal testing cross project like this but the only real way would be to set up some more comprehensive integration tests on this repo

hervelemeur:
@mumoshu since helm chart repositories respect OCI specs, you should be able to use any of [https://artifacthub.io/packages/helm/](https://artifacthub.io/packages/helm/) as public OCI repository AFAIK.

jdolitsky:
@mumoshu - you can create a test account on [bundle.bar](https://bundle.bar/). Here are instructions on how to push Helm chart: https://bundle.bar/docs/supported-clients/helm/

mumoshu:
Thank you everyone for reviewing!

@chrismellard I'll cut a patch release soon.

@hervelemeur @jdolitsky I'll definitely try Artifact Hub and bundle.bar 👍

PR #1671, opened 2021-02-05
Add ability to control skipping Deps and Repos separately

kuzaxak:
Right now we have only one flag which passed to both variables, which isn't flexible as we would like to.

Main problem in our case that we are using Helmfile with envs. In each env we can have different version of remote chart. To apply that configuration in parallel we would like to split repos sync and deps build into 2 stages.

Repos sync can't be executed in parralel because it cause helm locks and `context deadline exceeded` exceptions.

### DIFF
--- a/main.go
+++ b/main.go
@@ -186,7 +186,11 @@ func main() {
 				},
 				cli.BoolFlag{
 					Name:  "skip-deps",
-					Usage: `skip running "helm repo update" and "helm dependency build"`,
+					Usage: `skip running "helm dependency build"`,
+				},
+				cli.BoolFlag{
+					Name:  "skip-repos",
+					Usage: `skip running "helm repo update" before running "helm dependency build"`,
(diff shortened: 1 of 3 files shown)


dudicoco:
@kuzaxak this is a valuable feature, why was the PR closed?

kuzaxak:
> this is a valuable feature

I found root cause for the issue we had. Problem with helm dependency cache itself. 
We are changing env variable to run helmfile in parallel. 

```
script.dir(PATH) {
      script.withEnv([
        "HELMFILE_ENVIRONMENT=${deployment.directory}",
        "HELM_REPOSITORY_CACHE=${cacheFolder}/cache",
        "HELM_REPOSITORY_CONFIG=${cacheFolder}/config.json",
      ]) {
        sh("helmfile -e ${deployment.name} diff")
      }
    }
```

PR #1673, opened 2021-02-07
fix: make --kube-context override helmDefaults

astorath:
This would allow cli flag `--kube-context` to override value in helmDefaults allowing to use different values in local development and CI context.


mumoshu (approved):
Makes sense. Perhaps this could have been discussed in github issues beforehand but I think this is pretty safe/won't break anything. Let's try!

mumoshu:
Let me tag this is `feat: ` assuming this isn't actually a bug fix :)

Issue #1632, opened 2020-12-29
Update deprecated helm-secrets repo to active fork


cvitan:
Current helm-secrets fork used in helmfile is https://github.com/futuresimple/helm-secrets which nowadays redirects to https://github.com/zendesk/helm-secrets.

This project is no longer being maintained, and they are referring to https://github.com/jkroepke/helm-secrets as the active fork to switch to.

PR #1676, opened 2021-02-10
Update reference for helm-secrets to latest maintained fork

eperdeme:
Minor reference fix to the latest place to find helm-secrets.
Tested the plugin myself and works without an problems.

Resolves https://github.com/roboll/helmfile/issues/1632


eperdeme:
Failing due to helm2 support.

Can either work around this or should we drop helm2 support ?

mumoshu:
@eperdeme Thanks for the PR! I think reverting helmfile+helm2 image to use futuresimple's is the best thing we can do at the moment.

mumoshu (approved):
LGTM. Thanks for your contribution @eperdeme ☺️

PR #1691, opened 2021-02-24
feat: change to helmfile template so it respects logging levels

chrismellard:


jstrachan:
@mumoshu any chance of a review please? its a fairly simple change ;)

mumoshu (approved):
LGTM. Thanks, @chrismellard and @jstrachan!

mumoshu:
@chrismellard @jstrachan Hey! I started to think we got to revert this. What is this change supposed to do, and for which use-case(s) do you need this patch?

I think I've completely misread this patch while reviewing- To me, this looks to break `helmfile template` or `helmfile template > all.yaml` completely, due to that its output is now flowing into stderr instead of stdout. This can be also confirmed by CI builds failing since this change.

jstrachan:
@mumoshu ah ok! the main aim was to be able to tone down the logging levels of `helmfile template` to avoid pages of output of every `helm template` command for every release + every file in every chart being written to a file - it wasn't attempting to modify the semantics of stdout/stderr.

I wonder if there's better way to keep the same logging level of helm?

mumoshu:
@jstrachan Ah gotcha! Thanks for clarifying.

> I wonder if there's better way to keep the same logging level of helm?

Yeah probably- How about forcing writes to stdout when and only when `--output-dir` is not passed to `helm template`?

PR #1698, opened 2021-02-26
fix bug #1694

### LINKED ISSUE #1694: Multiple releases with the same name
HI,

I think I found a bug, here is the context:
I have a multicluster use case. For each kube cluster I want to deploy some releases, sometimes with the same configuration, sometimes with some differences.
I came from a bash script that does that, and I want to migrate to Helmfile.

Current Helm releases have the same name on each cluster.

To facilitate the deploying of releases in these clusters, I have a Helmfile that looks like this:

```yaml
{{- define "release" }}

{{ $extraValues := dict }}
{{ if hasKey $.Context.data "extraValues" }}
{{ $extraValues = deepCopy $.Context.data.extraValues }}
{{ end }}

{{ range $pop, $popData := .Context.data.pops }}
{{ if kindIs "string" $popData }}
{{ $popData = dict "ip" $popData }}
{{- end }}

{{- $popValues := dict }}
{{- if hasKey $popData "extraValues" }}
{{ $popValues = deepCopy $popData.extraValues }}
{{- end }}

{{ $size := pluck "size" $popData $.Context.data | first }}
{{ $channel := pluck "channel" $popData $.Context.data | first }}
  - name: {{ $.Context.name }}
    chart: ../channels/{{ $channel }}
    kubeContext: {{ $pop }}
    namespace: {{ $.Context.name }}
    createNamespace: true
    skipDeps: true
    labels:
      pop: {{ $pop }}
      client: {{ $.Context.name }}
    values:
      - {{ sprigGet (sprigGet $.Values.pops $pop) "values" | toJson }}
      - "../sizes/{{ $size }}.yaml"
      - proxy-out:
          defaultOrigin: {{ $.Context.data.defaultOrigin | toJson }}
        proxy-in:
          service:
            externalIPs:
              - "{{ $popData.ip }}"
      - {{ $extraValues | toJson}}
      - {{ $popValues | toJson }}
    secrets:
      - "certificates/{{ $.Context.name }}.enc.yaml"
    atomic: true
{{- end }}
{{- end }}

releases:
{{ range $name, $data := .Values.clients }}
{{ $context := (dict "name" $name "data" $data) }}
{{ $root := dict "Values" (deepCopy $.Values) | merge (dict "Context" $context) }}
{{ template "release" $root }}
{{ end }}

```

The problem I encounter is that when I try to launch a helmfile diff it seems to mix some releases specific values accross clusters. If I try to launch each specific cluster separatly using selector everything looks good. If I change the release name to make it unique accross my cluster the problem also disapear, but as I have already releases deployed this way this isn't really a good solution...

A workaround I found is to split each cluster in it's own step using the `helmfiles:` directive, but I loose concurrent deployements accross cluster...

Can you helm me ?

xabufr:
When the same release name is used accross namespaces/kubecontexts
a bad chart name could be used

fixes #1694

### DIFF
--- a/pkg/app/run.go
+++ b/pkg/app/run.go
@@ -16,7 +16,7 @@ type Run struct {
 	helm  helmexec.Interface
 	ctx   Context
 
-	ReleaseToChart map[string]string
+	ReleaseToChart map[state.PrepareChartKey]string
 
 	Ask func(string) bool
 }
@@ -68,7 +68,12 @@ func (r *Run) withPreparedCharts(helmfileCommand string, opts state.ChartPrepare
 	for i := range r.state.Releases {
 		rel := &r.state.Releases[i]
 
-		if chart := releaseToChart[rel.Name]; chart != "" {
+		key := state.PrepareChartKey{
+			Name:        rel.Name,
+			Namespace:   rel.Namespace,
+			KubeContext: rel.KubeContext,
+		}
+		if chart := releaseToChart[key]; chart != "" {
 			rel.Chart = chart
 		}
 	}
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -886,6 +886,8 @@ type ChartPrepareOptions struct {
 
 type chartPrepareResult struct {
 	releaseName            string
+	releaseNamespace       string
+	releaseContext         string
 	chartName              string
 	chartPath              string
 	err                    error
@@ -907,6 +909,10 @@ func (st *HelmState) GetRepositoryAndNameFromChartName(chartName string) (*Repos
 	return nil, chartName
 }
 
+type PrepareChartKey struct {
+	Namespace, Name, KubeContext string
+}
+
 // PrepareCharts creates temporary directories of charts.
 //
 // Each resulting "chart" can be one of the followings:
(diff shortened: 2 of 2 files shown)


pkg/app/run.go
```diff
@@ -68,7 +68,12 @@ func (r *Run) withPreparedCharts(helmfileCommand string, opts state.ChartPrepare
 	for i := range r.state.Releases {
 		rel := &r.state.Releases[i]
 
-		if chart := releaseToChart[rel.Name]; chart != "" {
```
mumoshu, line 71: Ah, this says it all. Thanks for finding!

mumoshu (approved):
LGTM. Thank you so much for your contribution @xabufr ☺️

jwitko:
@xabufr @mumoshu if either of you has a free moment and is willing, would you mind elaborating on this issue for me?  I'm just trying to educate myself and I'm having a hard time understanding what was going wrong and how this fixes it.  Was there just a fix implemented in one place and this carries it over?

xabufr:
I will try to explain:

When Helmfile prepare your releases charts, it calls  `PrepareCharts` that returns a map that associate a release to it's chart path.

With the old version of `PrepareCharts`, the problem appears when you have more than one release with the same name but in differents kube contexts/namespaces: the map that associates a release to it's chart path uses only the release name as key, so if you have multiple releases with the same name but that deploys differents charts (name or version) in differents contexts, wrong charts may be picked (as multiple releases have the same map key).

My fix is to introduce use a key with the release kubeContext/namespace/releaseName. This permits to find the good chart path for each releases, even if your releases have the same name but in different contexts.

Is it more clear ? (cause I'm not very good at english...)

jwitko:
Yes this makes *perfect* sense and is very similar to some previous issues we ran into and fixed here.  Thank you so much for the explanation @xabufr !

Issue #1700, opened 2021-02-26
Support templated secret values


phihos:
Value files ending with `.gotmpl` support templating while secret value files ending with `.gotmpl` currently cause an error.
Since secret value files behave like normal value files after encryption it should be technically possible to implement that without breaking backward compatibility.

PR #1701, opened 2021-02-26
Support templated secret values (#1700)

phihos:
Secret files ending with `.gotmpl` are now also rendered as template.

For that I had to fix the assumed filename of the decrypted secret file, because it predicted it wrong as soon as the file did not end with `.yaml`. After that I had to make the handler method for decrypting files (`generateSecretValuesFiles`) use the same template processing method as for normal value files (`generateTemporaryReleaseValuesFiles`).

A new unit test has been added and the existing integration test has been extended.


phihos:
It took me a while, but the integration tests are now passing in CI. I had to put some effort into the CircleCI config and I think it is a nice improvement since Ubuntu 20.04 images are now used and some duplication is eliminated as well.

mumoshu:
@phihos Hey! Thanks for the PR.

I've reviewed this but I'm quite confused. Would you mind if I ask a few fundamental things?

- What is the expected use-case of this?
  - Are you trying to encrypt a template file? Or are you trying to encrypt a secret yaml file generated from a template?
  - Why would you like to do so?
- Does `helm-secrets` supports encrypting the gotmpl file if it isn't a valid YAML?

phihos:
Hi @mumoshu,
I had a template values file like this:
```yaml
# values.yaml.gotmpl
environment:
  -   name: MY_EXTERNAL_IP
      value: |
          {{ exec "./get-external-ip.sh" (list "") }}
  - name: MY_SECRET_VALUE
    value: somesecret
```
Now I had the issue of exposing secrets in my repo so I wanted to encrypt them via sops. So I thought I could just declare this file as a secret file in my helmfile. But this was not supported and I began to implement it.
My implementation works just fine with helm secrets as long as the yaml is valid like in the example above.

mumoshu:
@phihos Sorry for the delayed response I had been overwhelmed by the sea of github notifications.

> Is there an easier way to achieve this that I did not see?

Probably no- Could you confirm my idea anyway?

Initially I thought you could just use one values and one secrets yaml files like:

`values.yaml.gotmpl`:

```
environment:
  -   name: MY_EXTERNAL_IP
      value: |
          {{ exec "./get-external-ip.sh" (list "") }}
```

`secrets.yaml`:

```
_sops:
  ...

environment:
  - name: MY_SECRET_VALUE
    value: (encrypted by sops)
```

so that you could write helmfile.yaml like:
```
releases:
- name: foo
  values:
  - values.yaml
  secrets:
  - secrets.yaml
```

But it didn't work as you might have expected because helm is not able to merge two arrays (`environment` from values and secrets in this case)?

mumoshu:
And assuming my above guess is correct, I think this is almost the right way to go.

The last thing I'm concerned about currently is that if this works against non-yaml file?

Apparently your example `secrets_templated.yaml.gotmpl` says it was a valid yaml file before being encrypted. But a `yaml.gotmpl` isn't always a valid yaml file- Actually it only needs to be a valid yaml only after gotmpl is rendered.

Is this patch supposed to work in that case? (I'm just asking- If it isn't supported to work, I'd definitely appreciate you fixing it but I can also do it on my-side before/after merging)

phihos:
As you correctly assumed the example you posted does not work. I tried the same thing before implementing the feature.

The code I implemented currently only supports templates that are also valid yaml. I am not sure if that can ever be fixed because when running helmfile the secret file first needs to be decrypted and then rendered. Since helm secrets does not seem to support encrypting invalid yaml getting any template into encrypted form is not possible. Only the ones that also pass a yaml syntax check I think.

mumoshu:
@phihos I appreciate your detailed response! This looks good now. I tried my best resolving conflicts with the latest main branch in 706741d. Would you mind confirming if I've done it correctly?

mumoshu:
@phihos I'm currently struggling to adapt the added test cases to the current main branch de5f679 707dc1e, revealing that I don't fully understand how this works yet 😃

mumoshu:
@phihos Hey! I returned to this after a few days break and I've finally managed to fix the build. I'm now merging this. If you can have some time double-checking my changes this week, it's more than welcome!

Thanks anyway for your contribution ☺️

mumoshu:
@phihos Could you also confirm if the commit message I've summarized is correct?  85accf7

phihos:
@mumoshu Thanks a million for finishing this! I was more occupied this weekend than I thought I would be. I just reviewed everything and the tests look fine. Though I am not sure why you split the golden templates and secrets, but I don't mind.
You described the feature correctly in the docs. In the YAML samples the indentation seems to be a bit off in two occasions (behind the dashes of list items).
I am looking forward to your next release. I really got to love this tool :heart:

yurrriq:
Thanks for this!

PR #1706, opened 2021-03-03
fix: pull OCI charts one by one

### LINKED ISSUE #1705: Helm pull from OCI in parallel
I've found a bug related to how Helm works.

When Helm pull multiple charts in parallel, it lost some pulled charts from it's local repository, example:

```bash
export HELM_EXPERIMENTAL_OCI=1
helm chart pull CHART-A:0.2.0 & \
helm chart pull CHART-B:1.0.0
helm chart ls # Only prints CHART-B, but A and B where pulled without any error
```

A simple workaround is to disable concurrent chart pulling [here](https://github.com/roboll/helmfile/blob/master/pkg/app/run.go#L62) (I've tested it and it works)

xabufr:
fixes #1705

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -73,6 +73,11 @@ type ReleaseSetSpec struct {
 	MissingFileHandler string `yaml:"missingFileHandler,omitempty"`
 }
 
+type PullCommand struct {
+	ChartRef     string
+	responseChan chan error
+}
+
 // HelmState structure for the helmfile
 type HelmState struct {
 	basePath string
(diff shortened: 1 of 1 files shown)


pkg/state/state.go
```diff
@@ -961,6 +966,11 @@ func (st *HelmState) PrepareCharts(helm helmexec.Interface, dir string, concurre
 	}
 
 	var builds []*chartPrepareResult
+	pullChan := make(chan *PullCommand)
+	defer func() {
+		pullChan <- nil
```
mumoshu, line 971: I believe that we can just `close(pullChan)` here and change the for-loop to:

```
for pullCmd := range pullChan {
	err := helm.ChartPull(pullCmd.ChartRef)
	pullCmd.responseChan <- err
}
```

for simplicity. I'd appreciate it if you could try that that submit another pull request for cleanup purpose 😃

mumoshu (approved):
Great job! Thank you so much for reporting and fixing it @xabufr ☺️

pkg/state/state.go
xabufr: Okay, I didn't know that (I'm a noob in Go language :smile:  ), I'll do a PR for that !

PR #1711, opened 2021-03-08
allow to access oci repos without auth

### LINKED ISSUE #1710: Allow to use public oci repositories
For example, github packages repo ghcr.io allows to create public accessible artifacts.
Now helmfile exits with message
`in ./helmfile.yaml: username and password are required fields for logging in to OCI registries with helm
`

kupnu4x:
fixes #1710

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -382,13 +382,12 @@ func (st *HelmState) SyncRepos(helm RepoUpdater, shouldSkip map[string]bool) ([]
 		if shouldSkip[repo.Name] {
 			continue
 		}
-		var err error
+		var err error = nil
 		if repo.OCI {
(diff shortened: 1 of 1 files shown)


pkg/state/state.go
```diff
@@ -382,13 +382,12 @@ func (st *HelmState) SyncRepos(helm RepoUpdater, shouldSkip map[string]bool) ([]
 		if shouldSkip[repo.Name] {
 			continue
 		}
-		var err error
+		var err error = nil
```
mumoshu, line 385:
```suggestion
		var err error
```

AFAIK the zero value for `error` is nil so we can just write it like this 👍

mumoshu (approved):
LGTM. Thanks a lot for your contribution @kupnu4x!

Issue #1650, opened 2021-01-14
Support waiting for jobs (new in helm 3.5.0)


btashton:
Helm now supports waiting for jobs as of 3.5.0 we should expose this like we do with `wait`.  This would resolve #785

In helm the flag is `--wait-for-jobs` and came in as https://github.com/helm/helm/pull/8363

Issue #785, opened 2019-08-02
Wait for job to be successfully running


shikhar7:
Hi, I am trying to run the below helmfile.yaml and in case of mysql as shown below i have one job and i need to wait for the job to be successfully completed before going to another helm chart release.I am running helmfile sequentially by using --concurrency=1.

I have earlier used shell script and i used to do something like `kubectl wait --for=condition=complete --timeout=-1s job/mysql-load -n ${global_namespace}`.Is there any way to do so in helmfile.If yes can you please elaborate on that.
```
releases:
{{ if .Environment.Values.isDmpPrerequisites }}
  - name: dmp-prerequisites
    chart: ./helm-installer/dmp-prerequisites
    values:
    - values.yaml.tmpl
{{ end }}

{{ if .Environment.Values.isDynamicPv }}
  - name: dynamic-pv
    chart: ./helm-installer/dynamic-pv
    values:
    - values.yaml.tmpl
{{ end }}

{{ if .Environment.Values.isMysqlInternal }}
  - name: mysql
    chart: ./helm-installer/mysql
    values:
    - values.yaml.tmpl
    wait: true
{{ end }}

{{ if .Environment.Values.isHelp }}
  - name: help
    chart: ./helm-installer/help
    values:
    - values.yaml.tmpl
{{ end }}

environments:
  default:
    values:
    - environments/default/values.yaml
```

mumoshu:
Yep. Try setting `wait: true` for each release under `releases`, or run `kubectl wait` in [hooks](https://github.com/roboll/helmfile#hooks)!

shikhar7:
I have tried to set wait:true in all and in command how do i need to set this ...setting this way doesn't work.
```
releases:
{{ if .Environment.Values.isDmpPrerequisites }}
  - name: dmp-prerequisites
    chart: ./helm-installer/dmp-prerequisites
    values:
    - values.yaml.tmpl
{{ end }}

{{ if .Environment.Values.isDynamicPv }}
  - name: dynamic-pv
    chart: ./helm-installer/dynamic-pv
    values:
    - values.yaml.tmpl
{{ end }}

{{ if .Environment.Values.isMysqlInternal }}
  - name: mysql
    chart: ./helm-installer/mysql
    values:
    - values.yaml.tmpl
    wait: true
{{ end }}

{{ if .Environment.Values.isHelp }}
  - name: help
    chart: ./helm-installer/help
    values:
    - values.yaml.tmpl
    hooks:
[...]

PR #1715, opened 2021-03-11
Enable --wait-for-jobs flag introduced in helm3.5

jwitko:
Fixes https://github.com/roboll/helmfile/issues/1650, https://github.com/roboll/helmfile/issues/785

This PR adds:

- `--wait-for-jobs` CLI override option to helmfile binary
- `waitForJobs` helmDefaults parameter (default `false`)
- `waitForJobs` release parameter (default `false`)

`--wait-for-jobs` was introduced in Helm3.5 ( https://github.com/helm/helm/pull/8363 )

### DIFF
--- a/pkg/state/temp_test.go
+++ b/pkg/state/temp_test.go
@@ -37,39 +37,39 @@ func TestGenerateID(t *testing.T) {
 	run(testcase{
 		subject: "baseline",
 		release: ReleaseSpec{Name: "foo", Chart: "incubator/raw"},
-		want:    "foo-values-67b55dc69b",
+		want:    "foo-values-779795898b",
 	})
 
 	run(testcase{
 		subject: "different bytes content",
 		release: ReleaseSpec{Name: "foo", Chart: "incubator/raw"},
 		data:    []byte(`{"k":"v"}`),
-		want:    "foo-values-5988bf4947",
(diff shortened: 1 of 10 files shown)


pkg/state/temp_test.go
```diff
@@ -37,39 +37,39 @@ func TestGenerateID(t *testing.T) {
 	run(testcase{
 		subject: "baseline",
 		release: ReleaseSpec{Name: "foo", Chart: "incubator/raw"},
-		want:    "foo-values-67b55dc69b",
+		want:    "foo-values-779795898b",
```
jwitko, line 40: @mumoshu I'm not 100% on if I should be doing this since the outcome of the suffix has changed or if I should be striving to make the tests have identical suffix?  I'm guessing with the introduce of the additional helmDefaults it required this change?

mumoshu: > I'm guessing with the introduce of the additional helmDefaults it required this change?

@jwitko I believe so, and it's behaving as expected in terms of implementation of generateIValuesID func so you're all good 👍

mumoshu (approved):
LGTM. Thank you so much for the addition ☺️

StepanKuksenko:
thank you for this function
we want this one much ! can you tell when do you plan next release ?

PR #1721, opened 2021-03-14
Fix/helm-secrets version

astorath:
This pins `helm-secrets` version in `Dockerfile.helm3`

- feat(Dockerfile): bump helm version
- fix(Makefile): static-linux mod
- feat(Dockerfile): pin helm-secrets version


mumoshu (approved):
LGTM. Thanks for fixes @astorath!

PR #1729, opened 2021-03-18
Feature: kubectl hooks

javiplx:
This is an attempt to make less verbose simple hooks. If you like this, I'll add some documentation and tests.
This will allow replacing
```yaml
- events: [ postsync ]
  command: kubectl
  args:
    - apply
    - --kustomize=kustom
```
with
```yaml
- events: [ postsync ]
  command: kustomize
  args:
    - kustom
```

### DIFF
--- a/pkg/event/bus.go
+++ b/pkg/event/bus.go
@@ -63,10 +63,10 @@ func (bus *Bus) Trigger(evt string, evtErr error, context map[string]interface{}
 		switch class := hook.Class; class {
 			case "apply":
 				hook.Command = "kubectl"
-				hook.Args = append([]string{"apply", "-f"}, hook.Args...)
+				hook.Args = append([]string{"apply", "-n" , "{{.Release.Namespace}}", "-f"}, hook.Args...)
 			case "kustomize":
 				hook.Command = "kubectl"
-				hook.Args = append([]string{"apply", "-k"}, hook.Args...)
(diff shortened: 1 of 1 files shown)


mumoshu:
@javiplx Hey! Thanks for the PR.

I generally agree with the general direction. Sorry for nit what I think would make this better is that design this around making it more declarative while simple.

What if it looked like:

```
- events: [ postsync ]
  kubectlApply:
    args:
    - --kustomize=kustom
```

and

```
- events: [ postsync ]
  kustomizeApply:
    args:
    - kustom
```

?

One more slight benefit I think this has is that we can start with `args` but gradually grow the API to include more features out-of-the-box. If we are to support `kustomize` flag natively later, it can be like:

```
- events: [ postsync ]
  kubectlApply:
    kustomize: kustom
```

javiplx:
> ```
> - events: [ postsync ]
>   kubectlApply:
>     kustomize: kustom
> ```

I like very much this last suggestion. The only point against is that my go is quite limited and I'm not sure whether I'll be able to implement that. Actually, even adding serious tests for my initial suggestion might be challenging ...  
But give me a few days to think if the challenge can be accepted.

javiplx:
It wasn't finally that hard, and I'm more or less satisfied with the resulting code : roboll/helmfile#1736

PR #1732, opened 2021-03-19
(fix) remove usage of charts.yaml in error message

itscaro:


mumoshu (approved):
LGTM! Thanks for the improvement @itscaro ☺️

PR #1734, opened 2021-03-19
(feat) add fetch command to download charts for debug

itscaro:
Should the temporary directory for charts download be $HOME/.helmfile/cache for commands like `lint`?

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -291,6 +291,24 @@ func (a *App) Lint(c LintConfigProvider) error {
 	}, SetFilter(true))
 }
 
+func (a *App) Fetch(c FetchConfigProvider) error {
+	return a.ForEachState(func(run *Run) (ok bool, errs []error) {
+		prepErr := run.withPreparedCharts("pull", state.ChartPrepareOptions{
+			ForceDownload: true,
+			SkipRepos:     c.SkipDeps(),
+			SkipDeps:      c.SkipDeps(),
+			OutputDir:     c.OutputDir(),
+		}, func() {
+		})
+
+		if prepErr != nil {
+			errs = append(errs, prepErr)
+		}
+
+		return
+	}, SetFilter(true))
+}
+
 func (a *App) Sync(c SyncConfigProvider) error {
 	return a.ForEachState(func(run *Run) (ok bool, errs []error) {
 		prepErr := run.withPreparedCharts("sync", state.ChartPrepareOptions{
--- a/main.go
+++ b/main.go
@@ -328,6 +328,28 @@ func main() {
 			}),
 		},
 		{
+			Name:  "fetch",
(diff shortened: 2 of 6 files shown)


pkg/app/app.go
```diff
@@ -289,6 +289,24 @@ func (a *App) Lint(c LintConfigProvider) error {
 	}, SetFilter(true))
 }
 
+func (a *App) Pull(c PullConfigProvider) error {
```
mumoshu, line 292: Can we leave the name `pull` for more broader use and use e.g. a more specific variant like `prepare-charts` for now? If you prefer shorter, I think `prepare` is also fine as we already use `prepare` with the exact same meaning in our [hooks](https://github.com/roboll/helmfile#hooks)

mumoshu:
> the temporary directory for charts download be $HOME/.helmfile/cache for commands like lint?

Yes, that sounds like a great idea! I also guess doing so would probably allow us to let Helmfile reuse already prepared charts if requested.

pkg/app/app.go
itscaro: how about `fetch`? to be in line with helm :-)

`helm fetch` is alias of `helm pull`

mumoshu: @itscaro Hey! Yeah- maybe good enough?

What scares me is that it isn't exactly just `fetch` either- `PrepareChart` involves the nasty "chartify" step that runs various commands to turn even local directory containing k8s manifests/kustomize into a temporary helm chart.

https://github.com/roboll/helmfile/blob/25baebe2b18cc059389b4bb1fcd334383466f0e2/pkg/state/state.go#L1051

WDYT? Would it be okay to just call it `helmfile fetch`?

itscaro: Yes I did the change to call it `helmfile fetch` and yes local chart will be copied into the temporary directory too. The current behaviour of this command is a bit intutive, I leave the default output directory as the temporary directory which is deleted afterwards. To keep downloaded charts (or locally copied ones) `--output-dir` must be set. Tell me what do you think?

itscaro:
> > the temporary directory for charts download be $HOME/.helmfile/cache for commands like lint?
> 
> Yes, that sounds like a great idea! I also guess doing so would probably allow us to let Helmfile reuse already prepared charts un subsequent command if requested.

Yes. I will try to address this proposal in another PR.

pkg/app/app.go
mumoshu: Thanks! It would look perfect if `--output-dir` defaults to somewhere under `$HOME/.helmfile/cache`, as other Helmfile commands fetches charts into there. That way,  running `helmfile fetch` beforehand would probably make subsequent `helmfile apply` faster.

mumoshu (approved):
LGTM. Thanks for the addition @itscaro ☺️

PR #1737, opened 2021-03-25
Fix helmfile-template not writing to stdout

mumoshu:
This fixes #1691 by redirecting helm-template output to our logger when and only when --output-dir is being passed to helm-template.

See https://github.com/roboll/helmfile/pull/1691#issuecomment-805636021 for more context.


mumoshu:
@jstrachan @chrismellard Would this be what you wanted originally? I'm fixing #1691 this way so that you can still use `helmfile template | kubectl apply -f -`, while enabling to preevnt excessive logs when using `--output-dir` like `helmfile --log-level=error template --output-dir path/to/dir`.

Issue #494, opened 2019-03-15
[Feature Request] Add local Helm Chart templates to repository chart release


sboschman:
Let me start with describing the issue I am trying to solve.

Many (official) Helm charts allow you to specify annotations on object descriptions (e.g. ingress descriptions). Ingress controllers support f.e. authentication by looking at annotations on the ingress objects. E.g. `ingress.kubernetes.io/auth-type` and `ingress.kubernetes.io/auth-secret`. This `ingress.kubernetes.io/auth-secret` annotation defines the name of the k8s secret containing the username/password for basic auth.

But the charts don't offer a way to create the secret object. If the chart did include a secret template, you could have used the helm secret-plugin to render.

Manually creating the secrets feels bad...

Another option would be to create a helmfile release based on a local chart, having this local chart containing the secret template. Imho it would be a lot cleaner if this secret would be part of the same Helm release as f.e. the ingress object.

This makes me wonder if it is sensible to have a helmfile feature to merge arbitrary local Helm templates into the chart before rendering.

helmfile.yaml:
```yaml
releases:
- name: myprometheus
  chart: stable/prometheus
  values:
  - values.yaml
  secrets:
  - secret/my-prometheus-secret.yaml
  templates:
  - templates/basic-auth-secret.yaml
```

values.yaml:
```yaml
server:
  ingress:
    enabled: true
    annotations:
      ingress.kubernetes.io/auth-type: "basic"
      ingress.kubernetes.io/auth-secret: prometheus-basic-auth-secret
```

templates/basic-auth-secret.yaml:
```yaml
apiVersion: v1
kind: Secret
type: Opaque
metadata:
  name: prometheus-basic-auth-secret
  labels:
    app: {{ template "prometheus.name" . }}
    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
    heritage: {{ .Release.Service }}
    release: {{ .Release.Name }}
data:
  auth: {{ .Values.myPrometheusSecret | b64enc | quote }}
```

Feature request #392 describes how to get secrets from different backends/sources (Helm secret plugin / Vault / AWS / [KeyHub](https://topicus-keyhub.com/) :heart: ) into helmfile. This feature request is about a way to get those secrets into k8s as `Secret`s as part to the release they belong to.

**Dream mode**
Being able to add arbitrary Helm templates would allow me to do the following.

[...]

PR #1745, opened 2021-03-30
feat: Helmfile renders *.yaml.gotmpl in a K8s manifests/kustomization directory

mumoshu:
Related to https://github.com/roboll/helmfile/issues/494

This feature is mostly a built-in alternative to incubator/raw chart without external dependency and has
access to helmfile's own template functions and template data.

The expected use-case of this feature is to add arbitrary K8s resources to your deployment.

Unlike the original issue raised in #494 this doesn't enable you to add arbitary resources to a release. That's another story. But this would be a good foundation for that, too.


hilsenrat:
Thanks @mumoshu!
Do you know when we can expect a release with this feature? And documentation about how to use it? Might be really useful for us.  

cc (regarding release) @roboll

PR #1749, opened 2021-03-30
feat: --show-secrets on diff and apply commands

### LINKED ISSUE #1674: helm-diff --show-secrets
Hi,

Before I sync or apply changes, I'd like to know if some of the secrets will be changed. Implemented [here](https://github.com/databus23/helm-diff/blob/0882d8c63d782cd5ac325fe3b4e00342ae30f6a1/cmd/release.go#L76).

This is an opposite argument to the `--suppress-secrets` that we need from time to time as we want to be sure that secret data is not changed, while annotations may be changed.

I can add `--args --show-secrets` but helmfile in some cases runs `helm list --filter` command (when I have `installed` parameter in any of my releases set to `false`) that throws an error.

This is similar to https://github.com/roboll/helmfile/issues/1240 issue.

Thanks!

strainovic:
This PR adds --show-secrets argument on the diff and apply commands that is useful for debugging purposes.

Resolves #1674

### DIFF
--- a/main.go
+++ b/main.go
@@ -200,6 +200,10 @@ func main() {
 					Name:  "suppress-secrets",
 					Usage: "suppress secrets in the output. highly recommended to specify on CI/CD use-cases",
 				},
+				cli.BoolFlag{
+					Name:  "show-secrets",
+					Usage: "do not redact secret values in the output. should be used for debug purpose only",
+				},
 				cli.IntFlag{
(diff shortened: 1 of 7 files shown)


mumoshu (approved):
LGTM. Thanks for the enhancement @strainovic ☺️

jim-barber-he:
When will there be a release with this new feature?

PR #1751, opened 2021-03-31
feat: allow override chart in args

### LINKED ISSUE #1690: Support overriding chart from command line
We have a use case to build chart on demand and deploy, it would be nice if helmfile can allow overriding chart in command line.

yujunz:
Closes #1690


mumoshu (approved):
LGTM. Thanks a lot for your contribution @yujunz ☺️

Issue #1655, opened 2021-01-19
jsonPatches failed with missing dependencies of a chart


jmederer:
When running `helmfile sync` with following Yaml
```
repositories:
  - name: bitnami
    url: https://charts.bitnami.com/bitnami
    caFile: bitnami.pem

helmDefaults:
  args:
    - "--debug"
  atomic: true
  cleanupOnFail: true
  skipDeps: false
  
releases:
  - name: cassandra-bit
    namespace: cassandra
    labels:
      team: noops
    chart: bitnami/cassandra
    values:
      - values-cassandra.yaml 
    jsonPatches:
    - target:
        version: apps/v1
        kind: StatefulSet
        name: cassandra-bit
      patch:
        - op: add
          path: /spec/template/metadata/annotations
          value:
            deonPatchesloyment/timestamp: "{{ now | unixEpoch }}"
```
helmfile exited with error it cant find the only dependency of the chart.

```
Adding repo bitnami https://charts.bitnami.com/bitnami
"bitnami" has been added to your repositories

running helm fetch bitnami/cassandra --untar -d /tmp/chartify775686341
using requirements.yaml:
{}

options: {false [/tmp/values195671422] [] cassandra  true}
running helm template --debug=false --include-crds --output-dir /tmp/chartify775686341/cassandra/helmx.1.rendered -f /tmp/chartify775686341/cassandra/values.yaml -f /tmp/values195671422 --namespace cassandra cassandra-bit /tmp/chartify775686341/cassandra
patching files: [/tmp/chartify775686341/cassandra/templates/networkpolicy.yaml /tmp/chartify775686341/cassandra/templates/cassandra-secret.yaml /tmp/chartify775686341/cassandra/templates/headless-svc.yaml /tmp/chartify775686341/cassandra/templates/service.yaml /tmp/chartify775686341/cassandra/templates/statefulset.yaml]
jsonpatches/patch.0.yaml:
- op: add
  path: /spec/template/metadata/annotations
  value:
    deonPatchesloyment/timestamp: "1611071318"

generated and using kustomization.yaml:
kind: ""
apiversion: ""
resources:
- templates/networkpolicy.yaml
- templates/cassandra-secret.yaml
- templates/headless-svc.yaml
- templates/service.yaml
- templates/statefulset.yaml
patchesJson6902:
- target:
    kind: StatefulSet
    name: cassandra-bit
    version: apps/v1
  
  path: jsonpatches/patch.0.yaml

[...]

PR #1761, opened 2021-04-08
Fix chartify to take --include-crds into account

### LINKED ISSUE #1760: --include-crds does not work with jsonPatches/strategicMergePatches/transformers/etc
Extracted from https://github.com/roboll/helmfile/issues/1655#issuecomment-815292454

mumoshu:
Fixes #1760
Ref https://github.com/variantdev/chartify/commit/d8f7eac17df3e7a77b5e61f5d2fc11fdfa376d52

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,7 @@ require (
 	github.com/spf13/cobra v1.1.1
 	github.com/tatsushid/go-prettytable v0.0.0-20141013043238-ed2d14c29939
 	github.com/urfave/cli v1.22.5
-	github.com/variantdev/chartify v0.7.2
+	github.com/variantdev/chartify v0.7.3
 	github.com/variantdev/dag v0.0.0-20191028002400-bb0b3c785363
 	github.com/variantdev/vals v0.13.0
 	go.uber.org/multierr v1.6.0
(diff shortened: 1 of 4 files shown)


mumoshu:
@damoon I'd appreciate it if you could test this out! It's not going to be released soon but you could run `make build` and run `./helmfile` to test.

damoon:
@mumoshu 
First of all, thanks for fixing https://github.com/roboll/helmfile/issues/1655

I tested this and it works for `helmfile template` vs `helmfile template --include-crds`.
The issue remains, that the CRDs are still included (when using patches and using helm args to skip crds) during the installation via `helmfile sync` (and i assume `helmfile apply`).
I would like to have an option to ignore skipCRDs on a per release base via the helmfile.yaml.

My use-case needs to install the CRDs first. This is done via a self maintained chart.
rook-ceph needs the CRD (ServiceMonitor) from prometheus-monitoring the prometheus-monitoring needs the CRD from rook-ceph (s3 bucket).
Currently i fail to install the prometheus chart, because the CRD is already managed by another chart.

mumoshu:
@damoon Hey! Thanks for testing & confirming.

Regarding `sync` and `apply`, what we would need to add to helmfile seems like the equivalent to `--skip-crds` of the upstream `helm [upgrade|install]`:

```
$ helm upgrade --help | grep crd
      --skip-crds                    if set, no CRDs will be installed when an upgrade is performed with install flag enabled. By default, CRDs are installed if not already present, when an upgrade is performed with install flag enabled
```

I've opened #1770 for that. Please correct me there if you had wanted anything else!

Issue #809, opened 2019-08-16
feat: optionally scope `helmfile deps` to specific releases


NelsonJeppesen:
feat: optionally scope `helmfile deps` to specific releases
---
I'd like to have more control when updating `helmfile.lock`. Right now I can update all releases, no releases or do it manually.

### use case
I have 20+ releases in a monitoring/observe namespace. Many large and complex charts like _Prometheus_ and _Jenkins_. Running `helmfile deps` and updating everything at once is a huge amount of change. I'd like to update one or more at a time and move that change though dev, stg, prod

### examples

`helmfile deps jenkins`

`helmfile deps elastic/elasticsearch` 
if I'm using some other repo
```repositories:
  - name: elastic   
    url: https://helm.elastic.co
```

mumoshu:
Thanks for the request!

I took some time to think about this and realized that this should be better done in the upstream first.

That is, I'd enhance `helm dep up` to accept a new flag for limiting which chart in the Chart.yaml(v3)/requiremens.yaml(v2) is checked for update.

yurrriq:
We've been wanting this feature for while now. @mumoshu, do you still think the change belongs upstream? I would love to be able to use helmfile labels too, e.g. `helmfile -l myLabel=foo deps`, which seems to require a change on the helmfile side anyway.

mumoshu:
@yujunz Hey! Thanks for resurrecting this. I believe enhancing deps to support labels `helmfile -l foo=bar deps` does make sense. Do you need anything other than that?

What @NelsonJeppesen has requested seemed something totally different than what you suggested, as it was about running `helm dep up` by chart name, rather than release label or release name.

yurrriq:
Ah, I see. I'd like to select by labels and release name.

mumoshu:
@yurrriq Thanks! Okay, let's beging with that.

mumoshu:
@yurrriq PTAL at #1764

I'm still scratching my head about how we could achieve what @NelsonJeppesen has originally requested.

NelsonJeppesen:
> @yurrriq PTAL at #1764
> 
> I'm still scratching my head about how we could achieve what @NelsonJeppesen has originally requested.

Nice to see activity on this! While my original request was slightly easier for my use case, using a selector achieves the same results.

I think (it's been awhile) my thought process was that updating by chart made more sense because there can be a many to one mapping between chart and releases. We have several releases that use the same helm chart. Just a clarity thing. 

[...]

PR #1764, opened 2021-04-09
Enable helmfile-deps to limit helm-dep-up by selector

mumoshu:
Ref https://github.com/roboll/helmfile/issues/809#issuecomment-814423653

This is not what had been originally requested in #809 but anyway- This enables you to limit helm-dep-up run by helmfile-deps by selector. For example, `helmfile -l name=foo deps` should result in only the release named `foo` to be helm-dep-uped.

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -2128,9 +2128,26 @@ func (st *HelmState) ResolveDeps() (*HelmState, error) {
 
 // UpdateDeps wrapper for updating dependencies on the releases
 func (st *HelmState) UpdateDeps(helm helmexec.Interface) []error {
+	var selected []ReleaseSpec
+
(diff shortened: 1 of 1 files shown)


yurrriq:
Thanks! I'll test this "at work" (from home) tomorrow. If I want to contribute a test, should I PR and target the `scope-helmfile-deps-with-selector` branch?

mumoshu:
@yurrriq I'm merging this for early testers like you but please feel free to submit PRs to add tests against our main branch :) (That's much appreciated!

yurrriq:
I did a quick initial test, and when I pass a selector, the resulting lock file _only_ includes that release, and it deletes the lock file entries for releases that don't match the selector, which is not the desired behaviour.

PR #1766, opened 2021-04-09
Allow users other than root to use helm plugins

aperigault:
Hi,
as discussed in this issue  https://github.com/roboll/helmfile/issues/1134, I propose to allow all users to go through root directory to be able to use helm plugins in container running as non root users.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,6 +44,9 @@ RUN helm plugin install https://github.com/databus23/helm-diff && \
     helm plugin install https://github.com/aslafy-z/helm-git.git && \
     helm plugin install https://github.com/rimusz/helm-tiller
 
+# Allow users other than root to use helm plugins located in root home
+RUN chmod 751 /root
+
 COPY --from=builder /workspace/helmfile/dist/helmfile_linux_amd64 /usr/local/bin/helmfile
 
 CMD ["/usr/local/bin/helmfile", "--help"]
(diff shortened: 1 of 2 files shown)


aperigault:
Hi @mumoshu, what do you think about that?

mumoshu:
@aperigault Hey! Thanks. Honestly speaking I don't understand for what usecase this can be helpful.

How did you test this? What was your example usecase and the exact docker-run command you used to verify this?

mumoshu:
Or maybe you tried to complement https://github.com/roboll/helmfile/issues/1134#issuecomment-605624954 and tried to fix https://github.com/roboll/helmfile/issues/1134#issuecomment-610836923?

aperigault:
Hi!
My use case is to run this container as non root in kubernetes due to PSP limitations and use plugins installed in /root folder.
Yes I want to fix https://github.com/roboll/helmfile/issues/1134#issuecomment-610836923

aperigault:
Hi @mumoshu,  do you think this is a good workaround ?

mumoshu (approved):
I'm not sure if forcing everyone to use non-root user within a helmfile container is the right thing to do either, as it might limit debugging operations within the container. Or perhaps I'm still missing something? 🤔 

Probably this is a good workaround. This is almost a year old but it's better done than nothing! Thanks for your feedback and contribution. Merging.

PR #1772, opened 2021-04-12
Add --{include,skip}-needs to various helmfile commands

### LINKED ISSUE #1692: Not installing dependancies when using selector
i have some thing like 

```
releases:
  - name: cert-manager
    namespace: cert-manager
    chart: cert-manager/cert-manager
    version: v1.2.0
    missingFileHandler: Warn
  - name: prometheus
    needs:
      - cert-manager/cert-manager
    namespace: monitoring
    chart: prometheus/prometheus
    version: 13.3.1
    missingFileHandler: Warn
  - name: prometheus-adapter
    needs:
      - monitoring/prometheus
    namespace: monitoring
    chart: prometheus/prometheus-adapter
    version: 2.12.0
    missingFileHandler: Warn
```

And when i am using `helmfile --selector name=prometheus apply` , i expect cert-manager to be installed as dependency which is not happening

mumoshu:
Currently, this is going to be a breaking change for whoever relied on the helmfile's existing behavior of `helmfile -l foo=bar apply` silently ignoring the unfulfilled `needs`.

Since this change, for example, `helmfile -l name=myapp apply` with the below helmfile.yaml would fail and require you to add either `--include-needs` or `--skip-needs` to make it successful:

```
releases:
- name: myapp
  chart: charts/myapp
  needs:
  - mydb
- name: mydb
  chart: charts/mydb
```

## Affected commands

- `helmfile template` defaults to `--skip-needs=true` + `--include-needs=false` assuming the primary use of `helmfile template` is piping to `kubectl apply` and Argo CD, or saving to the disk.
- `helmfile diff`, `helmfile sync`, and `helmfile apply` defaults to `--skip-needs=false` + `--include-needs=false` assuming the user would like to be sure which is the behaviour they want given feedbacks in #1692

## Breaking change

- `helmfile -l name=myapp [apply|diff|sync]` now fail due to the unhandled unfulfilled dependency `mydb`

## Additional features

- `helmfile -l name=myapp sync --include-needs` syncs `mydb` and then `myapp`
- `helmfile -l name=myapp sync --skip-needs` syncs `myapp` only. This was the default behaviour before this change.
- `helmfile -l name=myapp diff --include-needs` diffs `mydb` and then `myapp`
- `helmfile -l name=myapp diff --skip-needs` diffs `myapp` only. This was the default behaviour before this change.
- `helmfile -l name=myapp apply --include-needs` syncs `mydb` and then `myapp`
- `helmfile -l name=myapp apply --skip-needs` syncs `myapp` only. This was the default behaviour before this change.

Resolves #1692

@bseenu @roderik @mwakkach @antaloala @FWest98 It's mostly implemented, except that I need to make the error more user-friendly. Currently it's like `"mydb" depended by "myapp" is not included`, which is hard for the user to know if they need to add either `--include-needs` or `--skip-needs` to make it successful.

### DIFF
--- a/main.go
+++ b/main.go
@@ -392,7 +392,7 @@ func main() {
 				},
 				cli.BoolFlag{
 					Name:  "include-needs",
-					Usage: `automaticlaly include releases from the target release's "needs" when --selector/-l flag is provided. Does nothing when when --selector/-l flag is not provided`,
+					Usage: `automatically include releases from the target release's "needs" when --selector/-l flag is provided. Does nothing when when --selector/-l flag is not provided`,
 				},
 				cli.BoolFlag{
 					Name:  "wait",
@@ -456,7 +456,7 @@ func main() {
 				},
 				cli.BoolFlag{
 					Name:  "include-needs",
-					Usage: `automaticlaly include releases from the target release's "needs" when --selector/-l flag is provided. Does nothing when when --selector/-l flag is not provided`,
+					Usage: `automatically include releases from the target release's "needs" when --selector/-l flag is provided. Does nothing when when --selector/-l flag is not provided`,
 				},
 				cli.BoolFlag{
 					Name:  "skip-diff-on-install",
--- a/main.go
+++ b/main.go
@@ -386,6 +386,14 @@ func main() {
 					Name:  "skip-deps",
 					Usage: `skip running "helm repo update" and "helm dependency build"`,
 				},
+				cli.BoolTFlag{
+					Name:  "skip-needs",
+					Usage: `do not automatically include releases from the target release's "needs" when --selector/-l flag is provided. Does nothing when when --selector/-l flag is not provided`,
+				},
+				cli.BoolFlag{
+					Name:  "include-needs",
+					Usage: `automaticlaly include releases from the target release's "needs" when --selector/-l flag is provided. Does nothing when when --selector/-l flag is not provided`,
+				},
 				cli.BoolFlag{
 					Name:  "wait",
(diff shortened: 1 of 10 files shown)


main.go
```diff
@@ -386,6 +386,14 @@ func main() {
 					Name:  "skip-deps",
 					Usage: `skip running "helm repo update" and "helm dependency build"`,
 				},
+				cli.BoolFlag{
+					Name:  "skip-needs",
+					Usage: `do not automatically include releases from the target release's "needs" when --selector/-l flag is provided. Does nothing when when --selector/-l flag is not provided`,
+				},
+				cli.BoolFlag{
+					Name:  "include-needs",
+					Usage: `automaticlaly include releases from the target release's "needs" when --selector/-l flag is provided. Does nothing when when --selector/-l flag is not provided`,
```
FWest98, line 395: Small typo

FWest98:
As for error message, I would propose:
`Chart "myapp" depends on "mydb" which is not included in the selector. Please indicate whether to skip (--skip-needs) or include (--include-needs) these dependencies in the deployment.`

mumoshu:
@FWest98 Thanks for the feedback!  I've borrowed the error message you've suggested mostly as-is. The biggest change may be that I used the term "release" like in `Release "myapp" depends on ` instead of calling it a chart.

mumoshu:
I'm going to add `--include-needs` to `helmfile template` as well, so that you can use `helmfile template --include-needs` to review the exact K8s resources being installed by `helmfile sync --include-needs`.

PR #1777, opened 2021-04-14
Fix destroy not reversing the order of releases

### LINKED ISSUE #1775: destroy not reversing the order of releases
helmfile.yaml

```
repositories:
  - name: cert-manager
    url: https://charts.jetstack.io
  - name: prometheus
    url: https://prometheus-community.github.io/helm-charts
  - name: kiali
    url: https://kiali.org/helm-charts
  - name: autoscaler
    url: https://kubernetes.github.io/autoscaler
  - name: bitnami
    url: https://charts.bitnami.com/bitnami
  - name: grafana
    url: https://grafana.github.io/helm-charts
  - name: elastic
    url: https://helm.elastic.co

helmDefaults:
  cleanupOnFail: true
  # wait for k8s resources via --wait. (default false), 300 seconds
  wait: true


releases:
  - name: cert-manager
    namespace: test
    chart: cert-manager/cert-manager
    version: v1.2.0
  - name: metrics-server
    namespace: test
    needs:
      - test/cert-manager
    chart: bitnami/metrics-server
    version: 5.8.0
  - name: prometheus
    needs:
      - test/cert-manager
    chart: prometheus/prometheus
    namespace: test
    version: 13.3.1
  - name: autoscaler
    namespace: test
    needs:
      - test/prometheus
    chart: autoscaler/cluster-autoscaler
    version: 9.4.0
  - name: prometheus-adapter
    namespace: test
    needs:
      - test/autoscaler
    chart: prometheus/prometheus-adapter
    version: 2.12.0
  - name: grafana
    namespace: test
    needs:
      - test/prometheus-adapter
    chart: grafana/grafana
    version: 6.6.3
  - name: external-dns
    namespace: test
    needs:
      - test/grafana
    chart: bitnami/external-dns
    version: 4.7.0
  - name: logstash
    namespace: test
    needs:
      - test/external-dns
    chart: elastic/logstash
    version: 7.11.1
```

Install order

```
processing 7 groups of releases in this order:
GROUP RELEASES
1     test/cert-manager
2     test/metrics-server, test/prometheus
3     test/autoscaler
4     test/prometheus-adapter
5     test/grafana
6     test/external-dns
7     test/logstash

```

uninstall order

```
processing 7 groups of releases in this order:
GROUP RELEASES
1     test/cert-manager
2     test/external-dns
3     test/grafana
4     test/prometheus-adapter
5     test/autoscaler
6     test/prometheus, test/metrics-server
7     test/logstash
```

```
bash-3.2$ helmfile --version
helmfile version v0.138.7
```

mumoshu:
Fixes #1775

### DIFF
--- a/pkg/state/state_run.go
+++ b/pkg/state/state_run.go
@@ -122,9 +122,9 @@ func SortedReleaseGroups(releases []Release, reverse bool) ([][]Release, error) 
 	}
 
 	if reverse {
-		sort.Slice(groups, func(i, j int) bool {
-			return j < i
-		})
+		for i, j := 0, len(groups)-1; i < j; i, j = i+1, j-1 {
+			groups[i], groups[j] = groups[j], groups[i]
+		}
 	}
 
 	return groups, nil


pkg/state/state_run.go
```diff
@@ -122,9 +122,9 @@ func SortedReleaseGroups(releases []Release, reverse bool) ([][]Release, error)
 	}
 
 	if reverse {
-		sort.Slice(groups, func(i, j int) bool {
-			return j < i
-		})
+		for i, j := 0, len(groups)-1; i < j; i, j = i+1, j-1 {
```
ldemailly, line 125: If I may `i++, j--` might be a bit clearer than `i, j = i+1, j-1` no?

also... a test? :) [repro for #1775 ]

ldemailly:
It's not super obvious why the original code failed; but it has to do with non stable guarantee of sort.Slice while sort.SliceStable would work - but the new implementation is more direct (otherwise Reverse could be used too)

to see the bug:
https://play.golang.org/p/68u0D2781PW

pkg/state/state_run.go
ldemailly: I forgot there is no , there in golang (old C habits die hard) so it would need to be something like
```golang
		for i, j := 0, len(groups)-1; i < j; i++ {
			groups[i], groups[j] = groups[j], groups[i]
			j--
		}
```
which is clearer to me but asymmetrical, so ignore me :-)

ps: tested with #1775 and this branch works so I guess maybe no additional test needed and just a quick patch release 0.138.8 would be most welcome - thanks!

mumoshu:
> also... a test? :) [repro for #1775 ]

@ldemailly Hey! Thanks for reviewing this. That's super helpful.

Re- test, I'd REALLY love to have one, but to be honest, I'm not entirely sure how I could reproduce it reliably. Reading your example, would it be theoretically enough to just have 7 or more elements to be sorted to reproduce it?

We already had a test that should ideally catch this kind of bug in https://github.com/roboll/helmfile/blob/master/pkg/app/destroy_test.go, but apparently it didn't due to that it only had 5 "groups" to be sorted https://github.com/roboll/helmfile/blob/ae942c5288895c84c79171e5446773e4cb41c4ce/pkg/app/destroy_test.go#L291-L295

ldemailly:
Yes anything with 7 or more groups will fail in the old code

But even without an additional test (outside of testing the fragment and manual test I did too) can we get this merged/released, it's a pretty serious/critical bug - what is holding up the merge?

mumoshu:
@ldemailly Thanks for confirming!

> what is holding up the merge?

My spare time? 😉

PR #1784, opened 2021-04-20
Feature/diff simple

astorath:
Added an option to use helm-diff's `--output` flag, e.g.:
```
helmfile diff --diff-output simple
```

see: https://github.com/databus23/helm-diff/pull/221

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ vendor/
 *.log
 .vagrant/
 *.lock
+test/integration/.gnupg/
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,11 +1,11 @@
 Vagrant.configure("2") do |config|
-  config.vm.box = "ubuntu/xenial64"
+  config.vm.box = "ubuntu/focal64"
   config.vm.hostname = "minikube.box"
   config.vm.provision :shell, privileged: false,
     inline: <<-EOS
       set -e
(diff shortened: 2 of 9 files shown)


mumoshu:
@astorath Hey! Thanks for the PR. Just a quick idea- can we use just `--output` instead if it doesn't collide with any existing flag?

astorath:
@mumoshu  Well, my guess was that in future there might be some flag to control `helmfile` output. I don't mind to replace the flag though.

mumoshu:
@astorath Hey! Thanks for confirming. I'd guess what we might add in the future for controlling output target could be something that sounds more like `--output-dir` as helm's, because we need to deal with multiple diffs of multiple releases.
I also see `kubectl get` has `-o | --output` which is for the output format. So I believe we can safely use `--output` here.

astorath:
@mumoshu Actually it does conflict with:

https://github.com/roboll/helmfile/blob/master/main.go#L804

mumoshu:
@astorath Thanks! Yeah, but I thought the meaning of `helm list --output` doesn't conflict with `helm diff`'s output flag.

astorath:
@mumoshu ok, you win...

mumoshu (approved):
LGTM 👍 Thank you so much for your contribution and cooperation @astorath

Issue #1782, opened 2021-04-20
KubeContext does not fallback to the default value


katsew:
Hi,

I try to generate manifests by running like:

```
helmfile --kube-context testing --namespace dev template \
  --output-dir ./output \
  --output-dir-template '{{ .OutputDir }}/{{ .Release.KubeContext }}/{{ .Release.Namespace }}/{{ .Release.Name }}'
```

However, the path `{{ .Release.KubeContext }}` is missing like `./output//dev/foo`, 
while I expected that the path becomes like `./output/testing/dev/foo`.
It seems that when kubeContext is missing in a release, it does not fallback to the default kubeContext provided by `--kube-context`.

For instance, 

```
releases:
- name: "foo"
  chart: ../../charts/foo
  kubeContext: testing
- name: "bar"
  chart: ../../charts/bar
```

in this helmfile, when I run the command above, the path for each release becomes like these.

- foo: `./output/testing/dev/foo`
- bar: `./output//dev/bar`

mumoshu:
@katsew Hey! Thanks for reporting. This indeed sounds like a bug. I'm labeling this as such. If you could contribute a PR for fixing it, that would be much appreciated 😄

PR #1789, opened 2021-04-21
fix: spec.KubeContext does not fallback to HelmDefaults.KubeContext

katsew:
fixes: #1782

### DIFF
--- a/pkg/app/app_apply_nokubectx_test.go
+++ b/pkg/app/app_apply_nokubectx_test.go
@@ -1,0 +1,1064 @@
+package app
+
+import (
+	"bufio"
+	"bytes"
+	"io"
+	"path/filepath"
+	"sync"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/roboll/helmfile/pkg/exectest"
+	"github.com/roboll/helmfile/pkg/helmexec"
+	"github.com/roboll/helmfile/pkg/testhelper"
+	"github.com/variantdev/vals"
+)
+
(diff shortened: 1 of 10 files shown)


katsew:
Hi, @mumoshu 

I made a fix of the issue that I opened yesterday,
but I'm not sure this fix works with no side effect.
Let me know if you have any concerns, thanks.

katsew:
I have tested the fix by running `make test`, then I fixed the failed tests,
but I'm not adding any tests for this fix.
Where should I put the test of this fix if it's necessary?

mumoshu:
@katsew Hey! Thanks for the PR. This looks great so far.

I'm still figuring that out, but could you start by adding a single test case "without the default kubecontext", in `pkg/app/app_apply_test.go`? That would help us clarifying what are differences when you've specified kubecontext or not, and also help us to not introduce any regression on cases that we don't set the default kubecontext.

katsew:
Hi, @mumoshu . Thanks for the response.
I'll add a test for the case you provided.

katsew:
@mumoshu 

I copy-n-pasted existing test cases which depend on the constants `helmV2ListFlags` and `helmV3ListFlags` ,
then I added new constants `helmV2ListFlagsWithoutKubeContext`, `helmV3ListFlagsWithoutKubeContext` to test the fix.
If it's ok, then I'll squash the commit.

Please review it, thanks.

mumoshu:
@katsew Thank you so much for your effort! Everything looks great. Let me squash it and merge this.

mumoshu (approved):

PR #1794, opened 2021-04-23
feat: moved chartify logs to Debug

astorath:
Chartify logs are very _chatty_ and stands out event in quiet mode. This PR moves them to debug to hide chart building process in normal output.


mumoshu (approved):
Wow, I didn't even remember that I had implemented `chartify.WithLogf`! Thanks for your help @astorath ☺️

Issue #1796, opened 2021-04-23
can't evaluate field KubeContext in type state.releaseTemplateDataRelease


katsew:
Hi,
 
I want to render `.Release.KubeContext` in release values,
but the error occurred and cannot run successfully.

```yaml
releases:
- name: my-release
  chart: ./charts/raw/raw
  namespace: default
  kubeContext: foo-context
  labels:
    app: test
  values:
    # this will be error
    - ./charts/raw/{{`{{ .Release.KubeContext }}`}}/{{`{{ .Release.Name }}`}}.{{ .Namespace }}.yaml
```

However, I can render the path of output template dir with `.Release.KubeContext` like this.

```sh
helmfile --namespace default --debug \
  template --output-dir "./out" \
  --output-dir-template "{{ .OutputDir }}/{{ .Release.KubeContext }}/{{ .Release.Namespace }}/{{ .Release.Name }}" \
  --args '--debug'
```

This inconsistency in `.Release` object looks weird to me, and I want to be fixed.
If it's ok to fix, let me try fixing this, thanks.

mumoshu:
@katsew Hey! Thanks for reporting. I'd appreciate it if you could fix that 👍 the current behaviour does look weird.

PR #1797, opened 2021-04-24
fix: can't evaluate field KubeContext in type state.releaseTemplateDataRelease

katsew:
fixes: #1796


mumoshu (approved):
LGTM. Thanks for your contribution @katsew ☺️

Issue #1810, opened 2021-04-26
lazy evaluation of rendering template does not work properly


katsew:
Hi, 

I fixed #1796, but `.Release.KubeContext` still not be rendered when I write template like this:

```
# helmfile --kube-context foo-context template 
---
releases:
- name: my-release
  chart: ./charts/raw/raw
  namespace: default
  labels:
    app: test
  values:
    - ./charts/raw/{{`{{ .Release.KubeContext }}`}}/{{`{{ .Release.Name }}`}}.{{ .Namespace }}.yaml
```

the field values actually become like this:

```
  values:
    - ./charts/raw//my-release.default.yaml
```

I looked into this, then I found that to make it right, call `state.ApplyOverrides` in `ExecuteTemplates` like this:
https://github.com/katsew/helmfile/blob/lazy-evaluation-of-overrides/pkg/state/state_exec_tmpl.go#L100

I think this behaviour is more appropriate to handle lazy evaluation of templates,
since `--kube-context` should override releases' kubeContext only if it is missing or empty.

However, behavior of overriding namespace is different from kube-context.
It seems that even if the namespace is specified in each releases,
namespaceOverride overrides these namespace 🤔 

I want to fix this behaviour, but I'm wondering how to fix.
Any ideas?

mumoshu:
@katsew Ah sorry! I should have noticed when I've reviewed #1789, but anything passed from command-line flags, like `--namespace` and `--kube-context` should be considered "overrides", that overrides every release's corresponding property.

That you wanted for kube-context there seems to be "defaulting", which is possible only through `helmDefaults` and happens only when the target release has no value for the field. There's no way to do "defaulting" via command-line flags, like you can see from the in-existence of `--default-namespace`.

Going forward, I think we should make a few changes.

First, do kubecontext overriding in ApplyOverrides, like:
```
func (st *HelmState) ApplyOverrides(spec *ReleaseSpec) {
	if st.OverrideKubecontext == "" {
		spec.KubeContext = st.OverrideKubecontext
	}
```

while do kubecontext defaulting in createReleaseTemplateData.

https://github.com/roboll/helmfile/blob/204f78c8ff6831bc3320fa23fc319cfbe7895b5a/pkg/state/state_exec_tmpl.go#L30

Curerntly, we are setting `KubeContext: release.KubeContext` here, even if the value is empty and there's non-empyt helmDefaults.KubeContext, which seems wrong.

WDYT?

katsew:
[...]

PR #1814, opened 2021-04-27
fix defaulting and overriding kubeContext

katsew:
Hi,

I fixed the behavior of defaulting and overriding kubeContext which is discussed in #1810.
It seems that primary issue of #1810 also solved by defaulting kubeContext properly.

### DIFF
--- a/pkg/state/state_exec_tmpl.go
+++ b/pkg/state/state_exec_tmpl.go
@@ -17,8 +17,12 @@ func (st *HelmState) Values() map[string]interface{} {
 }
 
 func (st *HelmState) createReleaseTemplateData(release *ReleaseSpec, vals map[string]interface{}) releaseTemplateData {
+	if release.KubeContext == "" {
+		release.KubeContext = st.HelmDefaults.KubeContext
+	}
 	tmplData := releaseTemplateData{
 		Environment: st.Env,
+		KubeContext: st.OverrideKubeContext,
 		Namespace:   st.OverrideNamespace,
 		Chart:       st.OverrideChart,
 		Values:      vals,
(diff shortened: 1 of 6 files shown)


pkg/state/state_exec_tmpl.go
```diff
@@ -17,8 +17,12 @@ func (st *HelmState) Values() map[string]interface{} {
 }
 
 func (st *HelmState) createReleaseTemplateData(release *ReleaseSpec, vals map[string]interface{}) releaseTemplateData {
+	if release.KubeContext == "" {
+		release.KubeContext = st.HelmDefaults.KubeContext
+	}
```
mumoshu, line 22: I admit I may have failed to do it universally, but anyway.. it would be nicer if we could make this side-effect-free, so that release isn't modified only by calling createReleaseTemplateData.

```suggestion
	kubeContext := release.KubeContext
	if kubeContext == "" {
		kubeContext = st.HelmDefaults.KubeContext
	}
```

Note that `ApplyOverrides` is ok to modify ReleaseSpec as it's designed as such and the function name states that, too. Also, `ApplyOverrides` overrides by overrids, not defaults. Overriding by defaults is not the thing we've done so far(I believe).

katsew: Ah, right. I'll fix it, thanks.

katsew:
@mumoshu 

Thanks for your review, the test which is accidentally passed by the side-effect is failed 👍 
I need an extra change to checkDuplicates to make things right.
Please double check it, thanks.

mumoshu (approved):
LGTM. Thank you so much for your continued effort @katsew ☺️ 

> I need an extra change to checkDuplicates to make things right.

It definitely looks right now! Good job 👍

PR #1822, opened 2021-05-01
Fix --selector to correctly deduplicate releases

### LINKED ISSUE #1818: helmfile template --selector not works in some cases
Hi,

I use helmfile template to the same release and filter by labels to output different manifests within the same release.
However, `--selector` not filter the target release on master.

Here is the sample command and helmfile.yaml to reproduce this.

```sh
helmfile --selector index=1 template --output-dir "./out" 

# I also tried
helmfile --selector index=1 template --include-needs=[false|true] --output-dir "./out" 
```

```yaml
releases:
  - name: foo
    chart: incubator/raw
    namespace: default
    labels:
      app: test
      component: raw
      index: '1'
    values:
      - resources:
          - value: 1

  - name: foo
    chart: incubator/raw
    namespace: default
    labels:
      app: test
      component: raw
      index: '2'
    values:
      - resources:
          - value: 2
```

expected output is this:

```
---
# Source: raw/templates/resources.yaml
metadata:
  labels:
    app: raw
    chart: raw-0.2.5
    heritage: Helm
    release: foo
value: 1 # same as --selector index=1
```

however, actual output is this:

```
---
# Source: raw/templates/resources.yaml
metadata:
  labels:
    app: raw
    chart: raw-0.2.5
    heritage: Helm
    release: foo
value: 2 # expected 1
```

It seems that selector not works and render all releases in the same namespace, so the last one win.
I looked into the tree and It seems that #1772 is the one this change introduced.

UPDATE:
I also try appending argument `--include-needs=true|false`, since #1772  mentioned about that, I still not work.

mumoshu:
Fixes #1818

cc/ @katsew

### DIFF
--- a/pkg/state/state_run.go
+++ b/pkg/state/state_run.go
@@ -165,21 +165,6 @@ func GroupReleasesByDependency(releases []Release, opts PlanOptions) ([][]Releas
 		}
 	}
 
-	for _, r := range releases {
-		if !r.Filtered {
-			for _, n := range r.Needs {
-				// To map n into idToReleases correctly, prepend KubeContext to n.
-				if r.KubeContext != "" {
-					n = r.KubeContext + "/" + n
-				}
-				if _, ok := idToReleases[n]; !ok {
-					id := ReleaseToID(&r.ReleaseSpec)
-					return nil, fmt.Errorf("%q depends on nonexistent release %q", id, n)
-				}
-			}
-		}
-	}
-
 	var selectedReleaseIDs []string
 
 	for _, r := range opts.SelectedReleases {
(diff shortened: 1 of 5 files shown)


pkg/state/state_run.go
```diff
@@ -165,21 +165,6 @@ func GroupReleasesByDependency(releases []Release, opts PlanOptions) ([][]Releas
...
-				if r.KubeContext != "" {
-					n = r.KubeContext + "/" + n
-				}
-				if _, ok := idToReleases[n]; !ok {
-					id := ReleaseToID(&r.ReleaseSpec)
-					return nil, fmt.Errorf("%q depends on nonexistent release %q", id, n)
-				}
-			}
-		}
-	}
```
mumoshu, line 181: See L209-219- One of the intentions of #1772 was to delegate this duplication detection to the `variantdev/dag`. This is now redundant. I've updated a few test cases for missing dependency errors accordingly.

PR #1825, opened 2021-05-03
bump Go to 1.16.3

itscaro:
Go 1.16 supports multiline in template, this might be useful.
https://github.com/golang/go/issues/29770


itscaro:
Update to another chart raw repository but the new one needs k8B 1.16, tried to install k8s 1.16 but kubeadm manifests are not compatible :-(

mumoshu:
@itscaro The `charts.helm.sh` one hosted on GitHub pages seems to work as I've verified in #1833. Let me merge this and see how it goes!

mumoshu (approved):

PR #1829, opened 2021-05-06
fix kubeContext does not globally defaulting its value

katsew:
related issue: https://github.com/roboll/helmfile/issues/1810#issuecomment-833268566

### DIFF
--- a/pkg/state/state_exec_tmpl.go
+++ b/pkg/state/state_exec_tmpl.go
@@ -17,10 +17,6 @@ func (st *HelmState) Values() map[string]interface{} {
 }
 
 func (st *HelmState) createReleaseTemplateData(release *ReleaseSpec, vals map[string]interface{}) releaseTemplateData {
-	kubeContext := release.KubeContext
-	if kubeContext == "" {
-		kubeContext = st.HelmDefaults.KubeContext
-	}
 	tmplData := releaseTemplateData{
 		Environment: st.Env,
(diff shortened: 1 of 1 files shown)


katsew:
@mumoshu 

Hi, can you fix this error?

```
Error: looks like "https://repo.chartcenter.io" is not a valid chart repository or cannot be reached: failed to fetch https://repo.chartcenter.io/index.yaml : 405 Not Allowed
```

I think this is not a problem of a test itself but its environment.
I run `make test` locally and all tests have passed.

mumoshu:
@katsew CI has been fixed via https://github.com/roboll/helmfile/pull/1833 👍

pkg/state/state_exec_tmpl.go
```diff
@@ -94,6 +90,9 @@ func (st *HelmState) ExecuteTemplates() (*HelmState, error) {
 	vals := st.Values()
 
 	for i, rt := range st.Releases {
+		if rt.KubeContext == "" {
+			rt.KubeContext = r.HelmDefaults.KubeContext
+		}
```
mumoshu, line 95: Ah, nice. This looks like the right place to do defaulting. I might rename `ExecuteTemplates` later to reflect that... perhaps `ApplyDefaultsAndReleaseTemplates` 🤣

mumoshu (approved):
LGTM. Thanks a lot for your continued effort @katsew ☺️

Issue #1830, opened 2021-05-06
0.139.0 -  "panic: bug: unexpectedly failed to get releases for id"


jseiser:
Once we upgraded, all our pipelines started failing.

Version: 0.138.7

```
❯ helmfile -e qa3 --selector name=mauka status
Getting status mauka
```

Version: 0.139.0

```
❯ helmfile -e qa3 --selector name=mauka status
panic: bug: unexpectedly failed to get releases for id "kube-system/high-priority": [qa3-mauka/mauka]

goroutine 1 [running]:
github.com/roboll/helmfile/pkg/state.GroupReleasesByDependency(0xc0000b9800, 0x1, 0x1, 0x10000, 0x0, 0x0, 0x0, 0x1, 0x1, 0x0, ...)
	/private/tmp/helmfile-20210506-26861-1vo93z9/helmfile-0.139.0/pkg/state/state_run.go:248 +0x1ac5
github.com/roboll/helmfile/pkg/state.SortedReleaseGroups(0xc0000b9800, 0x1, 0x1, 0x10000, 0x0, 0x0, 0x0, 0xc0005e52c0, 0x20, 0x37b5010, ...)
	/private/tmp/helmfile-20210506-26861-1vo93z9/helmfile-0.139.0/pkg/state/state_run.go:126 +0x91
github.com/roboll/helmfile/pkg/state.(*HelmState).PlanReleases(0xc00039d8c0, 0x10000, 0x0, 0x0, 0x0, 0x100fa38, 0x10, 0x20bbee0, 0xc00015a601, 0xc0005e8530)
	/private/tmp/helmfile-20210506-26861-1vo93z9/helmfile-0.139.0/pkg/state/state_run.go:115 +0xa9
github.com/roboll/helmfile/pkg/app.withDAG(0xc00039d8c0, 0x26fcc88, 0xc00053e840, 0xc000011168, 0x10000, 0x0, 0x0, 0x0, 0xc0005e8530, 0x0, ...)
	/private/tmp/helmfile-20210506-26861-1vo93z9/helmfile-0.139.0/pkg/app/app.go:872 +0x7d
github.com/roboll/helmfile/pkg/app.(*App).status(0xc00021e480, 0xc0001e2200, 0x26d86d0, 0xc0002b6ea0, 0xc00015ac08, 0xc0002c66c8, 0xadec0e9b42460901, 0xad0000c00015aa58)
[...]

PR #1835, opened 2021-05-09
Make `--skip-needs=true` the default option

mumoshu:
Ref https://github.com/roboll/helmfile/issues/1830#issuecomment-835632716
Ref #1692
Follow-up for #1772

### DIFF
--- a/main.go
+++ b/main.go
@@ -199,9 +199,9 @@ func main() {
 					Name:  "include-tests",
 					Usage: "enable the diffing of the helm test hooks",
 				},
-				cli.BoolFlag{
+				cli.BoolTFlag{
 					Name:  "skip-needs",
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -761,7 +761,11 @@ func (c configImpl) HasCommandName(name string) bool {
 }
 
 func (c configImpl) SkipNeeds() bool {
-	return c.c.Bool("skip-needs")
+	if !c.IncludeNeeds() {
+		return c.c.Bool("skip-needs")
```
yxxhero, line 765: @mumoshu I think it will return false. when --include-needs is not provided.

mumoshu: Why? `--include-needs` is `false` by default and in that case it falls into L765 and return `true`, as `BoolTFlag` defaults to `true`.

yxxhero: oh. I know.

PR #1844, opened 2021-05-12
Add support for fetching helmfiles as .tar.gz

### LINKED ISSUE #1842: remote helmfiles via http (go-getter) not working
Hi,

using this minimal helmfile.yaml...

```
helmfiles:
  - path: https://raw.githubusercontent.com/cloudposse/helmfiles/master/releases/external-dns/helmfile.yaml
```

... when running `helmfile deps` I get this error:

```
no matches for path: https://raw.githubusercontent.com/cloudposse/helmfiles/master/releases/external-dns/helmfile.yaml
There are no repositories defined in your helmfile.yaml.
This means helmfile cannot update your dependencies or create a lock file.
See https://github.com/roboll/helmfile/issues/878 for more information.
err: no releases found that matches specified selector() and environment(default), in any helmfile
```

Is fetching via HTTP not supported?

Kind regards,
Michael

mseiwald:
Fixes #1842

Example:

```
helmfiles:
  - path: https://host/path@helmfile.yaml.tar.gz
```

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -1079,6 +1079,7 @@ func (a *App) WrapWithoutSelector(converge func(*state.HelmState, helmexec.Inter
 
 func (a *App) findDesiredStateFiles(specifiedPath string, opts LoadOpts) ([]string, error) {
 	path, err := a.remote.Locate(specifiedPath)
+	path = strings.TrimSuffix(path, ".tar.gz")
 	if err != nil {
 		return nil, fmt.Errorf("locate: %v", err)
 	}
--- a/pkg/remote/remote.go
+++ b/pkg/remote/remote.go
@@ -4,14 +4,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
 	"github.com/hashicorp/go-getter"
 	"github.com/hashicorp/go-getter/helper/url"
 	"go.uber.org/multierr"
 	"go.uber.org/zap"
 	"gopkg.in/yaml.v2"
-	"os"
-	"path/filepath"
-	"strings"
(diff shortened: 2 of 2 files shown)


mseiwald:
@mumoshu  First super simplistic implementation but it actually works. Feedback appreciated.

mumoshu:
@mseiwald Hey! Thanks for your effort.

I gave it a shot and to me this doesn't work:

```
helmfiles:
  - path: https://github.com/helmfile/testdata/releases/download/test@helmfile-test.tar.gz
```

But this one works:

```
helmfiles:
  - path: https://github.com/helmfile/testdata/releases/download/test/helmfile-test.tar.gz@
```

Probably you expect the tarball to contain a single directory named `$BASE` where your tarball is named `$BASE.tar.gz`, right?

I'd say the latter form is better in respect to the meaning of `@` in helmfile, whose right part refers to a path in the downloaded directory.

that said, can you verify that it works with `helmfiles:
  - path: https://host/path/helmfile.yaml.tar.gz@the-dir-contained-in-your-tarball` for your tarball?

mseiwald:
@mumoshu Actually it works like this:

`helmfile.yaml.tar.gz` needs to contain a `helmfile.yaml` at the root level of the archive. (everything before `.tar.gz` is supposed to be the path of the "entry level" helmfile.

So in your first example you probably just need to rename the tar file.

Would that work for you?

mseiwald:
I just realized that your suggestion already works with the latest unmodified helmfile. 🤦 

```
helmfiles:
  - path: https://host/path/to/helmfile.yaml.tar.gz@helmfile.yaml
```

I'll just close this PR then since I agree that this is also the clearest syntax. Thanks anyway. :)

PR #1849, opened 2021-05-13
Fix Helmfile swallowing secrets decryption errors

raxod502-plaid:
There appears to be a simple typo that prevents Helmfile from correctly reporting an error when a secrets file cannot be decrypted (e.g. with sops). Instead, empty values are silently substituted into the rendered template.


mumoshu (approved):
Whoa!! Good catch! Thanks a lot for finding and fixing it @raxod502-plaid ☺️

yurrriq:
Thanks so much!

PR #1856, opened 2021-05-19
fix go getter download directory

itscaro:
go-getter curretly downloads to remote.DefaultCacheDir. fetch cmd (download charts to local for debug) has --output-dir flag and this is not respected for go-getter fetched charts.

This PR will breaks the caching mecanism because default folder used to download charts is temporary.

### DIFF
--- a/pkg/state/helmx.go
+++ b/pkg/state/helmx.go
@@ -39,9 +39,14 @@ type Chartify struct {
 	Clean func()
 }
 
-func (st *HelmState) downloadChartWithGoGetter(r *ReleaseSpec) (string, error) {
-	pathElems := []string{
-		remote.DefaultCacheDir,
+// downloadChartWithGoGetter downloads chart in ReleaseSpec to dir directory
+func (st *HelmState) downloadChartWithGoGetter(r *ReleaseSpec, dir string) (string, error) {
+	var pathElems []string
+
+	if dir == "" {
+		pathElems = append(pathElems, remote.DefaultCacheDir)
+	} else {
+		pathElems = append(pathElems, dir)
 	}
 
 	if r.Namespace != "" {
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -1085,7 +1085,7 @@ func (st *HelmState) PrepareCharts(helm helmexec.Interface, dir string, concurre
 
 				chartName := release.Chart
 
-				chartPath, err := st.downloadChartWithGoGetter(release)
+				chartPath, err := st.downloadChartWithGoGetter(release, dir)
 				if err != nil {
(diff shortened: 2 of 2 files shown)


mumoshu:
@itscaro Hey! Thanks for your efforts. To be clear, this seems to affect `helmfile fetch` only. You wanted to fix `helmfile-fetch` only, right?

> This PR will breaks the caching mecanism for go-getter fetched charts because default folder used to download charts is temporary.

It breaks only when you run `helmfile fetch`, and doesn't break for other helmfile commands, right?

I think `opts.OutputDir` (i.e. `dir` in PrepareCharts func) is set only on helmfile-fetch:
https://github.com/roboll/helmfile/blob/92ba347d6f899bd4b29a4279efef3199d0fa36b0/pkg/app/app.go#L341

itscaro:
@mumoshu No this impacts all go getter fetch operations, i mentioned fetch cmd because this cmd has possibility to change output dir (temporary working dir in other commands)

Another example, helmfile sync will download normal charts to temporary dir and go getter fetched charts to .helmfile dir next to helmfile.yaml (due to default dir set in go getter fetch function)

That's why I wanted to harmonize where charts fetched with go getter by passing dir from PrepareCharts

itscaro:
To summarize 

fetch cmd => normal charts go in either TempDir or OutputDir, go-getter charts go in CacheDir
other cmds => normal charts go in ~~either~~ TempDir ~~ or OutputDir~~, go-getter charts go in CacheDir

I think we should use either TempDir or CacheDir for all operations, Output-Dir is just optional for user to change the (output/cache/working dir)

I made another commit to use Cache Dir as default location instead of Temp Dir to demonstrate what I'm trying to explain :), tell me what you think?

https://github.com/roboll/helmfile/pull/1856/commits/4e9bf3fb396fe106118951d9c6e4ffdc3c7d6f85

Thank you @mumoshu for all of your efforts 🥇

itscaro:
Humh with the cached go-getter fetched chart, this error might show up

```
WARN: `helm dep build` failed. While processing release "pinpoint", Helmfile observed that remote chart "..." fetched by go-getter is seemingly broken. One of well-known causes of this is that the chart has outdated Chart.lock, which needs the chart maintainer needs to run `helm dep up`. Helmfile is tolerating the error to avoid blocking you until the remote chart gets fixed. But this may result in any failure later if the chart is broken badly.
```

mumoshu:
@itscaro That's just a warning. Did you actually see an error that stopped helmfile?

mumoshu:
Anyway I'll try having some time testing this manually this week to move this forward.

itscaro:
> @itscaro That's just a warning. Did you actually see an error that stopped helmfile?

No, it's just a WARN ;-) just noted the behavior here, my first time using go-getter fetched chart ;)

Issue #1783, opened 2021-04-20
Ubuntu/Debian-based docker images


z0rc:
Alpine-based docker images has their own limitations: https://dev.to/asyazwan/moving-away-from-alpine-30n4.

My personal issue: missing glibc, which leads to [inability to install aws-cli v2](https://github.com/aws/aws-cli/issues/4685) when running helmfile image in CI pipeline, and aws cli is needed when working with AWS EKS clusters. aws cli v1 works on alpine, but it isn't clear how long it will be supported.

Of course there is always an option to maintain my own docker images with tools I need, but maybe community would see value in this too.

mumoshu:
@z0rc Hey! Thanks for bringing it up. I've no strong opinion on sticking with alpine and will happily accept contributions to add more OSes. It would be great to start by having another Dockerfile to build Ubuntu or Debian based images, so that the community can start testing.

Would you mind contributing?

mumoshu:
#1863 for Ubuntu-based Dockerfiles(not in the release process yet. Use it as a reference to build your own)
#1956 for Debian-based images.

Closing as resolved. Thanks for your contributions @vseryozhenko and @naphta!

PR #1863, opened 2021-05-27
ubuntu docker based image for helmfile

vseryozhenko:
Implementation of ubuntu based docker image for issue [1783](https://github.com/roboll/helmfile/issues/1783)

### DIFF
--- a/Dockerfile.helm3.ubuntu
+++ b/Dockerfile.helm3.ubuntu
@@ -1,0 +1,63 @@
+FROM golang:1.16.3-buster as builder
+
+RUN apt update -qq && \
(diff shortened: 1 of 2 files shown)


mumoshu (approved):
Thanks for your contribution @vseryozhenko! I think this is worth being merged as this should be a good foundation for anyone trying to build their own container image for Helmfile based on Ubuntu, even though this isn't part of your release assets yet.

If anyone is interested in maintaining these Ubuntu variants of the image in a longer term, please chime in and add it to the release job in our CircleCI workflow. Thanks.

PR #1882, opened 2021-06-14
Added example ArgoCD + helmfile deployment

lukeplausin:
* Add an example deployment for using ArgoCD with Helmfile, private chartmuseum, private Vault server and private Git repository simultaneously.


mumoshu (approved):
LGTM. Thanks for your patience and thanks a lot for your contribution @lukeplausin ☺️

PR #1883, opened 2021-06-14
write temporary secret files to temp dir

itscaro:
To avoid writing these temporary files in current working dir. On my environment, os.Remove() does not work well due to error 32 (windows) which means file is used by other process, maybe antivirus scanning. Therefore secrets*.yaml files are left behind in working dir.


mumoshu:
Thanks for your contribution! I think #2023 is the successor to this change. Let's close this one and get #2023 done

Issue #1193, opened 2020-04-09
Add support for referencing base helmfiles from a URL


zoonage:
# Use Case
If you have multiple repositories with their own helmfiles being deployed it would be helpful to syncronise certain aspects, for example the `environments` section. This would make management of the per-repository helmfiles much simpler.

# Example

## Repository helmfile.yaml

```yaml
helmfiles:
  - "./010/helmfile.yaml"
  - "./020/helmfile.yaml"
bases:
  - git+ssh://git@github.com/zoonage/helm-lib@bases/environment.yaml
```

## Library bases/environment.yaml

```yaml
environments:
  local:
    values:
      - local.yaml
  testing:
    values:
      - testing.yaml
  staging:    
    values:
      - staging.yaml
  production:
    values:
      - production.yaml
```

PR #1891, opened 2021-06-22
bases fetchable as remote files

hypnopotamus:
`bases` pulled and cached from remote sources like `helmfiles`

I went to refactor such that environment was in a repo for common sharing and found that bases didn't yet support `git::` paths like `helmfiles` does


nml11c:
@mumoshu any chance you can review this?

mumoshu:
This seems to resolve #1193 and it turned out #1998 does basically the same thing.

Would you mind trying v0.143.0 which includes #1998?

Sorry for the delay, but thanks a lot for all your efforts!

PR #1905, opened 2021-07-03
bug: write-values deletes provided --values files.

### LINKED ISSUE #1904: write-values deletes provided values files
When running `helmfile write-values`, the provided `--values` files are deleted.   For example, running  `helmfile write-values --values /my/provided/values.yaml` will delete the file `/my/provided/values.yaml`

I wasn't expecting my source files to be deleted, so I'm running with the assumption that this is a bug 😉   More importantly, this means that this will always fail when sub-helmfiles are in play (e.g. the glob of helmfiles pattern) as the provided `--values` file won't exist when processing the first sub-helmfile.

Looking at the code that deletes the files, I wonder if this is bug whereby the `defer`ed function that is intended to delete the _generated_ values files  isn't taking in to the account that the `generatedFiles` is subsequently modified before the `defer` is activated.  PR to follow... hopefully making this clear.

jgchristian:
This fixes the bug by not including provided values files in the array of generated values, which is evaluated in a deferred block.  

Resolves #1904

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -1469,7 +1469,6 @@ func (st *HelmState) WriteReleasesValues(helm helmexec.Interface, additionalValu
 			if _, err := os.Stat(valfile); os.IsNotExist(err) {
 				return []error{err}
 			}
-			generatedFiles = append(generatedFiles, valfile)
 		}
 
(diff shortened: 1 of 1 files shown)


mumoshu (approved):
@jgchristian I didn't expect this to happen at all until I read your detailed investigation. Awesome work, and thanks a lot for fixing it!

PR #1921, opened 2021-07-19
Correct enabled property in helmfile list

### LINKED ISSUE #1920: helmfile list doesn't shows installed as enabled
As mentioned in https://github.com/roboll/helmfile/issues/1651#issuecomment-809507026 the `helmfile list` command shows the value of the `installed` property as `enabled`. This is confusing, moreover `helmfile list` doesn't offer a method to show only enabled charts for the current environment.

hikhvar:
Use the value of the `condition` field instead of the `installed` field of a release in the `enabled` column of helmfile list.
The value of the `installed` field is shown in a new `installed` column.

Fixes #1920

New output of `helmfile list`:
```
NAME      	NAMESPACE	ENABLED	INSTALLED	LABELS                    	CHART   	VERSION
myrelease1	         	true   	false    	common:label,id:myrelease1	mychart1	       
myrelease2	         	false  	true     	common:label              	mychart1	       
myrelease3	         	true   	true     	                          	mychart1	       
myrelease4	         	true   	true     	id:myrelease1             	mychart1	       
```

for `Helmfile`:
```
commonLabels:
  common: label
environments:
  default:
    values:
     - myrelease2:
         enabled: false
releases:
- name: myrelease1
  chart: mychart1
  installed: no
  labels:
    id: myrelease1
- name: myrelease2
  chart: mychart1
  condition: myrelease2.enabled
- name: myrelease3
  chart: mychart1
  installed: yes
- name: myrelease4
  chart: mychart1
  labels:
    id: myrelease1
````


zifter:
any updates to this isssue?

mumoshu (approved):
I'm still unsure if `INSTALLED` is the right word because it may be misinterpreted like a flag that indicates if it is actually installed onto the target cluster or not.

But this is much better than the current state! Let's go ahead. Thanks for your contribution.

PR #1925, opened 2021-07-22
Add IKEA to users.

almed4:


mumoshu (approved):
Wow, great to hear! Thanks for using Helmfile 😄

PR #1938, opened 2021-08-10
Add more examples for remote env value files

reixd:
I had difficulties loading remote value files even after looking at the docs here and go-getter.
The solution was in the test code here: 

https://github.com/roboll/helmfile/blob/ae942c5288895c84c79171e5446773e4cb41c4ce/pkg/remote/remote_test.go#L67

Use an `@` to separate the path from the GIT url... 

To help others I want to update the docs with the solution which worked for me ;-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -854,16 +854,20 @@ environments:
     values:
       - git::https://git.company.org/helmfiles/global/azure.yaml?ref=master
       - git::https://git.company.org/helmfiles/global/us-west.yaml?ref=master
+      - git::https://gitlab.com/org/repository-name.git@/config/config.test.yaml?ref=main # Public Gilab Repo
   cluster-gcp-europe-west:
     values:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -854,17 +854,21 @@ environments:
     values:
       - git::https://git.company.org/helmfiles/global/azure.yaml?ref=master
       - git::https://git.company.org/helmfiles/global/us-west.yaml?ref=master
```
mumoshu, line 856: I'm now wondering if this was never correct 🤔 Perhaps it should be `git::https://git.company.org/helmfiles@global/us-west.yaml?ref=master` ?

mumoshu (approved):
Ah, good catch! I do often forget about the fact that we need to use `@` myself, so this's a great addition. Thanks for your contribution @reixd

README.md
reixd: Yes, you need to use the `@` to separate the folder from the URL.

PR #1946, opened 2021-08-20
Update x/sys to support go 1.17

gordonbondon:
Updating x/sys so that this project supports go 1.17.

More context: https://github.com/Homebrew/homebrew-core/pull/83413


gordonbondon:
Scratch this, formula install failure appears to be related to SHA mismatch, not go build. See https://github.com/Homebrew/homebrew-core/pull/83630

PR #1951, opened 2021-08-28
build: update to use go1.17

chenrui333:


mumoshu:
We've upgraded to go1.17 via #2024. Thanks for your help!

PR #1956, opened 2021-09-06
Add debian docker image, drop helm2, and upgrade packages.

naphta:
- Add debian image based on `stable-slim`, desire for this is largely
  around my use case using Azure DevOps which makes it challenging to
  use images which are not glibc based.
- Drop support for helm2 in the docker images. This is a tricky one but
  given that I was having errors during the docker build for helm2 and
  the fact that it has been EoL for a long time now made me think that
  this was the correct move.
- As a "while I'm in here" I've upgraded kubectl and helm. I've popped
  on the most current patch of the last release (v1.20.3) to give a
  slightly broader support for different Kubernetes versions.
- Reworked CI to support pushing a debian and alpine base, and dropped
  support for the helm2 versions.

### DIFF
--- a/.circleci/Makefile
+++ b/.circleci/Makefile
@@ -1,5 +1,4 @@
 HELM_VERSION ?= v3.5.3
-HELM2_VERSION ?= v2.17.0
 KUSTOMIZE_VERSION ?= v3.8.8
 K8S_VERSION ?= v1.13.12
 MINIKUBE_VERSION ?= v0.30.0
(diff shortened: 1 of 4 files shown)


mumoshu (approved):
I was blocked by the fact that Helm 2.17.0 binary is nowhere while I was working on #1960.
In addition to that, this has a number of good things. Yeah, I'm not particularly tied to alpine so as far as it is maintained I'm eager to investigate one or two alternatives. Debian will be a good first alternative :)

So, LGTM. Thanks a lot for your contribution @naphta!

PR #1960, opened 2021-09-13
Add support for --validate on chartify

mumoshu:
Apparently we needed to pass `--validate` on helm-template run by chartify when the targeted chart contains Capabilities.APIVersions in a chart template. Otherwise, you can never make such chart work with chartify, as at apply time helm template expressions that involved Capabilities.APIVersions are already nowhere.


mumoshu:
Ah, finally helm 2.17.0 has gone

CleanShot 2021-09-14 at 04 24 50@2x

Issue #1889, opened 2021-06-18
Helmfile 0.139.0+ breaks chartify/jsonPatch


jwitko:
Helmfile versions 0.139.0 and later have broken chartify/kustomize integration.   Please see the following examples:
**test-helmfile.yaml**
```
releases:
  - name: testPatch
    namespace: default
    chart: incubator/raw
    values:
      - resources:
          - apiVersion: v1
            kind: Secret
            metadata:
              name: test
              namespace: default
            type: Opaque
            data:
              oci_config: FOO
    # Use kustomize to patch manifests with relevant labels
    jsonPatches:
      - target:
          version: v1
          kind: Secret
          name: test
          namespace: default
        patch:
          - op: replace
            path: "/data/oci_config"
            value: "BAR"
```

**Helmfile 0.138.7**
```
$ helmfile --debug -f test-helmfile.yaml
processing file "test-hemlfile.yaml" in directory "."
first-pass rendering starting for "test-hemlfile.yaml.part.0": inherited=&{default map[] map[]}, overrode=<nil>
first-pass uses: &{default map[] map[]}
first-pass rendering output of "test-hemlfile.yaml.part.0":
 0: releases:
 1:   - name: testPatch
 2:     namespace: default
 3:     chart: incubator/raw
 4:     kubeContext: "dev_uk-london-1_dataplane"
 5:     values:
 6:       - resources:
 7:           - apiVersion: v1
 8:             kind: Secret
 9:             metadata:
10:               name: test
11:               namespace: default
12:             type: Opaque
13:             data:
14:               oci_config: FOO
15:     # Use kustomize to patch manifests with relevant labels
16:     jsonPatches:
17:       - target:
18:           version: v1
19:           kind: Secret
20:           name: test
21:           namespace: default
22:         patch:
23:           - op: replace
24:             path: "/data/oci_config"
25:             value: "BAR"
26:

first-pass produced: &{default map[] map[]}
first-pass rendering result of "test-hemlfile.yaml.part.0": {default map[] map[]}
vals:
map[]
defaultVals:[]
second-pass rendering result of "test-hemlfile.yaml.part.0":
 0: releases:
 1:   - name: testPatch
 2:     namespace: default
 3:     chart: incubator/raw
 4:     kubeContext: "dev_uk-london-1_dataplane"
 5:     values:
 6:       - resources:
 7:           - apiVersion: v1
 8:             kind: Secret
 9:             metadata:
10:               name: test
11:               namespace: default
[...]

PR #1962, opened 2021-09-14
Fix json patches and strategic patches to do work when there was only one resource in the targeted chart/directory

### LINKED ISSUE #1954: Does kustomize work?
I have been trying jsonPatches and strategicMergePatches. Neither seems to work. For example:

the template

```
repositories:
- name: incubator
  url: https://charts.helm.sh/incubator

releases:
- name: raw1
  chart: incubator/raw
  values:
  - resources:
    - apiVersion: v1
      kind: ConfigMap
      metadata:
        name: raw1
        namespace: default
      data:
        foo: FOO
  strategicMergePatches:
    - apiVersion: v1
      kind: ConfigMap
      metadata:
        name: raw1
        namespace: default
      data:
        bar: BAR
```

renders into nothing and no configmap is ever deployed.
Does anyone have clarification on this? Thanks.

mumoshu:
Fixes #1954
Ref https://github.com/variantdev/chartify/commit/0b1378d379b91a5a1f60132ad80ff8dd3f91f447

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,7 @@ require (
 	github.com/spf13/cobra v1.1.1
 	github.com/tatsushid/go-prettytable v0.0.0-20141013043238-ed2d14c29939
 	github.com/urfave/cli v1.22.5
-	github.com/variantdev/chartify v0.9.0
+	github.com/variantdev/chartify v0.9.1
 	github.com/variantdev/dag v1.0.0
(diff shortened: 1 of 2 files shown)


jwitko:
This may actually fix https://github.com/roboll/helmfile/issues/1889 as well.  I will test.

PR #1982, opened 2021-10-05
Bump hashicorp/go-getter dependency to v1.5.9

### LINKED ISSUE #1981: Bump hashicorp/go-getter dependency to v1.5.9
When using `go-getter`-style URLs to manage remote helmfiles/values, it is currently impossible to specify a `ref` and a `depth` on the same URL. For example using a remote path like this: 

```
helmfiles:
- path: git::https://github.com/cloudposse/helmfiles.git@releases/kiam.yaml?ref=0.40.0&depth=1
```
will cause the following error:

`2021/10/01 09:31:20 Error downloading: error downloading 'https://github.com/cloudposse/helmfiles.git?ref=0.40.0&depth=1': /usr/bin/git exited with 1: error: pathspec '0.40.0' did not match any file(s) known to git.`

This is an issue with the go-getter library which was fixed in the most recent release [v1.5.9](https://github.com/hashicorp/go-getter/releases/tag/v1.5.9) (helmfile is currently at v1.5.0)

simoncaron:
This PR updates hashicorp/go-getter to v1.5.9

Resolves #1981

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/golang/mock v1.5.0
 	github.com/google/go-cmp v0.5.5
 	github.com/gosuri/uitable v0.0.4
-	github.com/hashicorp/go-getter v1.5.0
+	github.com/hashicorp/go-getter v1.5.9
 	github.com/hashicorp/go-retryablehttp v0.6.3 // indirect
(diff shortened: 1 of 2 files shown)


mumoshu (approved):
LGTM. Thanks for your help!

Issue #1915, opened 2021-07-13
Include Transitive dependencies


pjotre86:
Hello!
Imagine you have three services:  
`serviceA` which depends on `serviceB` which depends on `serviceC`.  
So in helmfile you would declare those dependencies with the helm of the `needs` feature:
```
serviceA:
  ...
  needs:
    - needs: serviceB

serviceB:
  ...
  needs:
    - needs: serviceC

serviceC:
  ...
```
Now I'd like to start up serviceA with all its dependencies using:  
`helmfile --selector name=serviceA  template --include-needs`  
But what's happening then is that just the direct dependency of `serviceA` (which `serviceB`) is started. serviceC which you could call a transitive dependency of `serviceA` is **not started**. That would have been the expected behavior for me (being used to docker-compose, which does exaclty that).  
  
So my question:  
### Is there a way to tell helmfile to also include transitive dependencies?

Issue #1881, opened 2021-06-14
[BUG] needs dependency graph is broken


dudicoco:
With helmfile v0.139.9 it seems that the `needs` dependencies behaviour is broken.

With the following example:
```yaml
- name: amazing-app-test-1
  chart: my-repo/amazing-app
  version: 2.0.14
  installed: true
  timeout: 300
  needs:
  - devops/amazing-app-test-3
  values: []
- name: amazing-app-test-2
  chart: my-repo/amazing-app
  version: 2.0.14
  installed: true
  timeout: 300
  needs:
  - devops/amazing-app-test-1
  values: []
- name: amazing-app-test-3
  chart: my-repo/amazing-app
  version: 2.0.14
  installed: true
  timeout: 300
  needs: []
  values: []
```

Previous behaviour - v0.138.4 (correct):
```
Upgrading release=amazing-app-test-3, chart=my-repo/amazing-app
Release "amazing-app-test-3" does not exist. Installing it now.
NAME: amazing-app-test-3
LAST DEPLOYED: Mon Jun 14 10:55:29 2021
NAMESPACE: devops
STATUS: deployed
REVISION: 1
TEST SUITE: None

Listing releases matching ^amazing-app-test-3$
amazing-app-test-3	devops   	1       	2021-06-14 10:55:29.884771 +0300 IDT	deployed	amazing-app-2.0.14	1.0

Upgrading release=amazing-app-test-1, chart=my-repo/amazing-app
Release "amazing-app-test-1" does not exist. Installing it now.
NAME: amazing-app-test-1
LAST DEPLOYED: Mon Jun 14 10:56:28 2021
NAMESPACE: devops
STATUS: deployed
REVISION: 1
TEST SUITE: None

Listing releases matching ^amazing-app-test-1$
amazing-app-test-1	devops   	1       	2021-06-14 10:56:28.275669 +0300 IDT	deployed	amazing-app-2.0.14	1.0

Upgrading release=amazing-app-test-2, chart=my-repo/amazing-app
Release "amazing-app-test-2" does not exist. Installing it now.
NAME: amazing-app-test-2
LAST DEPLOYED: Mon Jun 14 10:57:09 2021
NAMESPACE: devops
STATUS: deployed
REVISION: 1
TEST SUITE: None

Listing releases matching ^amazing-app-test-2$
[...]

PR #1983, opened 2021-10-05
Add support for transitive dependencies (needs).

pjotre86:
### Problem
When using selectors in helmfile one can use the `--include-needs` parameter to also install the dependencies of the selected release(s). However, this doesn't work for transitive needs, only the direct needs of the specified releases will be included. The problem is described in more detail with an example here: 
https://github.com/roboll/helmfile/issues/1915 
### Solution
After conferring with @mumoshu we decided to introduce a new parameter called `--include-transitive-needs`, which will also include transitive needs. We deliberately decided to not change the behavior of `--include-needs` (even if including transitives might be the expected behavior for most users) to ensure backwards compatibility.
### Notes
`--include-transitive-needs` will override any potential exclusions done by selectors or conditions. So even if you explicitly exclude a release via a selector it will still be part of the deployment in case it is a direct or transitive need of any of the specified releases.

Fixes roboll/helmfile#1915
Closes roboll/helmfile#1915


dudicoco:
Hi @pjotre86 and @mumoshu,

As far as I know, `needs` has been broken for a few months now, see https://github.com/roboll/helmfile/issues/1881.

Does this PR fix this bug?

pjotre86:
Hey @Hey @dudicoco !
Not sure, the I never ran into the problem with `needs` described in #1881.

dudicoco:
Hi @pjotre86,

Did you try to recreate the issue per the instructions in https://github.com/roboll/helmfile/issues/1881?

PR #2000, opened 2021-11-03
Make .Environment values available in .gotmpl files.

### LINKED ISSUE #1090: .Environment.Name is not set in environment values template
I'm trying to do dynamic environment values:
```
environments:
  services:
    values:
      - values.yaml.gotmpl
```
values.yaml.gotmpl contains simple exec of a file with ".Environment.Name" as an argument, but it is not set (empty).

Is it expected behavior? Considering environment name is passed as a cli parameter and should be available from the start I don't see a reason for it to not be set at this point.

pjotre86:
Currently it's not possible to use `.Environment` values in `*.gomtpl` files. The documentation states the opposite: 
https://github.com/roboll/helmfile#environment (2nd paragraph).
The problem is already described in #1090.
This PR fixes this bug.
Fixes #1090

### DIFF
--- a/pkg/app/desired_state_file_loader.go
+++ b/pkg/app/desired_state_file_loader.go
@@ -52,7 +52,7 @@ func (ld *desiredStateLoader) Load(f string, opts LoadOpts) (*state.HelmState, e
 		storage := state.NewStorage(opts.CalleePath, ld.logger, ld.glob)
 		envld := state.NewEnvironmentValuesLoader(storage, ld.readFile, ld.logger, ld.remote)
 		handler := state.MissingFileHandlerError
-		vals, err := envld.LoadEnvironmentValues(&handler, args)
+		vals, err := envld.LoadEnvironmentValues(&handler, args, &environment.EmptyEnvironment)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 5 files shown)


mumoshu (approved):
LGTM. Thanks for the fix @pjotre86!

lindhe:
How surprising. I've been using with Helmfile v0.141.0 `.Environment.Name` already, and it has been working fine. Is this PR just for `.Environment.Values` (`.Environment.values`?) or what's up?

pjotre86:
@lindhe :
Are you using `{{ .Environment.Name }}`  in a `*.gotmpl` file (like here: https://github.com/roboll/helmfile#values-files-templates) or in a base helmfile?

lindhe:
Yep, I'm using {{ .Environment.Name }} in a `*.gotmpl` file.

pjotre86:
That's weird!
For me and the reporters in #1090 it was always empty, before I fixed it. I don't have an explanation why it is working for you as there are myriads of ways how helmfile can be used an I also don't have a complete overview code-wise...

lindhe:
Weird indeed. It's not particularly important for me to figure out though, so as far as I'm concerned we can drop this topic. Just thought I'd bring it up, in case it helps the maintainer figure out some related bug...

jduepmeier:
There are two different "values". One is on in releases and one in environments.

Before this patch the .Environment.Name in environments value files was not present.

```yaml
releases:
- name: test
  values:
    - test.yaml.gotmpl # this was working
environments:
- test:
    values:
    - test.yaml.gotmpl # this was not working

PR #2002, opened 2021-11-12
feat: Option to pass kubeVersion to helm template

msvticket:
To be able to use `helmfile template` to render helm charts for cluster with other than the default Kubernetes version in helm you might need to add the --kube-version flag to helm. This adds that functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,13 +318,15 @@ bases:
 # Advanced Configuration: API Capabilities
 #
 # 'helmfile template' renders releases locally without querying an actual cluster,
-# and in this case `.Capabilities.APIVersions` cannot be populated.
-# When a chart queries for a specific CRD, this can lead to unexpected results.
+# and in this case `.Capabilities.APIVersions` cannot be populated. Also `Capabilities.KubeVersion` will probably be incorrect.
+# When a chart queries for a specific CRD or the Kubernetes version, this can lead to unexpected results.
 #
 # Configure a fixed list of api versions to pass to 'helm template' via the --api-versions flag:
 apiVersions:
 - example/v1
 
+# Configure a Kubernetes version to  pass to 'helm template' via the --kube-version flag:
+kubeVersion: v1.21
 ```
 
 ## Templating
--- a/pkg/app/app_test.go
+++ b/pkg/app/app_test.go
@@ -2692,12 +2692,15 @@ releases:
 	}
 }
 
-func TestTemplate_ApiVersions(t *testing.T) {
+func TestTemplate_ApiVersionsAndKubeVersion(t *testing.T) {
 	files := map[string]string{
 		"/path/to/helmfile.yaml": `
 apiVersions:
 - helmfile.test/v1
 - helmfile.test/v2
+
+kubeVersion: v1.21
+
 releases:
 - name: myrelease1
   chart: stable/mychart1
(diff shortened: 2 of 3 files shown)


44smkn:
Hi,

I found Datadog Chart use `.Capabilities.KubeVersion.GitVersion`.
https://github.com/DataDog/helm-charts/blob/de93d7c45467c2bf65a34afec0729f13341aed4f/charts/datadog/templates/_helpers.tpl#L521-L526

```python
{{- define "enable-service-internal-traffic-policy" -}}
{{- if or (semverCompare "^1.22-0" .Capabilities.KubeVersion.GitVersion) .Values.agents.localService.forceLocalServiceEnabled -}}
true
{{- else -}}
false
{{- end -}}
{{- end -}}
```

If `enable-service-internal-traffic-policy` is `true`, the `kind: Service` resource manifest which contains [internalTrafficPolicy](https://kubernetes.io/docs/concepts/services-networking/service-traffic-policy/) will be generated.
`internalTrafficPolicy` feature can be used only on Kubernetes 1.22+.
The clusters we have are v1.19 and v1.21, so we're a little troubled.

This pull request may resolve our problem.
I would like to know the status of this.

msvticket:
The reason I need this feature myself is that I need to set pathType in the ingresses for prometheus. But the prometheus chart only allows setting the pathType in Kubernetes 1.19+

https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/templates/_helpers.tpl#L200

msvticket:
What can I do to get this pull request merged?

README.md
```diff
@@ -318,13 +318,15 @@ bases:
...
-# When a chart queries for a specific CRD, this can lead to unexpected results.
+# and in this case `.Capabilities.APIVersions` cannot be populated. Also `Capabilities.KubeVersion` will probably be incorrect.
+# When a chart queries for a specific CRD or the Kubernetes version, this can lead to unexpected results.
 #
 # Configure a fixed list of api versions to pass to 'helm template' via the --api-versions flag:
 apiVersions:
 - example/v1
 
+# Configure a Kubernetes version to  pass to 'helm template' via the --kube-version flag:
+kubeVersion: v1.21
```
mumoshu, line 329: I'm still unsure this is the right level to where kubeVersion is defined, as Helmfile allows a different k8s cluster hence k8s version per release, by specifying `releases[].kubecontext`. But I do believe this is a good enough for a starter!

mumoshu (approved):
LGTM. Thanks for your patience and contribution!

mumoshu:
@msvticket @44smkn Hey! I recently learned that `--kube-version` is only present in helm2's `helm template`. So if your chart is relying solely on `Capabilities.KubeVersion` to detect the correct API version for a resource, this doesn't help.

So if you're using Helm 3, you'd just list API versions available to `helm template` by specifying `apiVersions` in your helmfile.yaml.

mumoshu:
Updated README to note the above info https://github.com/roboll/helmfile/commit/91aa0acddbacaa8c8295b9dfa95f045a39d0609b

msvticket:
$ helm template --help | grep kube-version
      --kube-version string          Kubernetes version used for Capabilities.KubeVersion
$ helm version
version.BuildInfo{Version:"v3.6.3", GitCommit:"d506314abfb5d21419df8c7e7e68012379db2354", GitTreeState:"dirty", GoVersion:"go1.16.5"}

--kube-version was added again with helm/helm#9040

44smkn:
When I ran `helmfile template` using helm v3.7.2 and helmfile v0.143.0, I got the expected results.
I think the feature proposed in this PR is useful in helm3 as well.

mumoshu:
@msvticket @44smkn Good to know! Thank you so much for sharing ☺️

PR #2003, opened 2021-11-13
Fix capitalization of disableOpenAPIValidation config in README

trumant:
See https://github.com/roboll/helmfile/blob/master/pkg/state/state.go#L220


mumoshu (approved):
Good catch! Thanks for your contribution @trumant ☺️

PR #2022, opened 2021-12-16
Use os.UserCacheDir()/helmfile to store downloaded artifacts instead of relative .helmfile directory

itscaro:
Current behavior:
Artifacts (remote helmfile or chart) are downloaded into `.helmfile` relative to working directory

Proposed behavior:
There artifacts are downloaded and cached in user's  `.cache` directory. Since the caching system is moved out of working directory, new commands are proposed `cache info` and `cache cleanup`

Test with `examples/remote`

```
root@e3f962517cc9:/source/examples/remote# /source/helmfile cache info
Cache directory: /root/.cache/helmfile
- apache
- https_github_com_cloudposse_helmfiles_git.ref=master

root@e3f962517cc9:/source/examples/remote# /source/helmfile cache cleanup
Cleaning up cache directory: /root/.cache/helmfile
- apache
- https_github_com_cloudposse_helmfiles_git.ref=master
```

### DIFF
--- a/examples/remote/helmfile.yaml
+++ b/examples/remote/helmfile.yaml
@@ -1,0 +1,10 @@
+helmfiles:
+- path: git::https://github.com/cloudposse/helmfiles.git@releases/echo-server/helmfile.yaml?ref=master
+  values:
+  - installed: true
+    stage: test
+    environment: test
+
+releases:
+- name: apache
+  chart: git::https://github.com/bitnami/charts.git@bitnami/apache?ref=master
--- a/main.go
+++ b/main.go
@@ -691,6 +691,27 @@ func main() {
 			}),
 		},
 		{
+			Name:      "cache",
+			Usage:     "cache management",
+			ArgsUsage: "[command]",
+			Subcommands: []cli.Command{
+				{
+					Name:  "info",
+					Usage: "cache info",
+					Action: action(func(a *app.App, c configImpl) error {
+						return a.ShowCacheDir(c)
+					}),
+				},
+				{
+					Name:  "cleanup",
+					Usage: "clean up cache directory",
+					Action: action(func(a *app.App, c configImpl) error {
+						return a.CleanCacheDir(c)
+					}),
+				},
+			},
+		},
+		{
 			Name:      "version",
 			Usage:     "Show the version for Helmfile.",
 			ArgsUsage: "[command]",
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -998,12 +998,7 @@ func (a *App) visitStatesWithSelectorsAndRemoteSupport(fileOrDir string, converg
 		opts.Environment.OverrideValues = envvals
 	}
 
-	dir, err := a.getwd()
-	if err != nil {
-		return err
-	}
-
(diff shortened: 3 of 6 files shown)


z0rc:
Please don't put dotfiles into user's home dir. Refer to https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html for Linux. This is more than a decade old spec, there is no good reason to not follow it. On macOS you can use `~/Library/Caches` and `~/Library/Application Support` for same purposes.

mumoshu:
@z0rc Thanks a lot for sharing your knowledge! It definitely sounds like what we should consider.

I also see several projects use `~/.cache` to store cache:

```
$ ls -lah ~/.cache/
go-build/             motd.legal-displayed
helm/                 staticcheck/
kubebuilder-envtest/  typescript/
```

Do you have any idea which spec (or sort of that) it is based on?

z0rc:
> Do you have any idea which spec (or sort of that) it is based on?

Linked XDG Base Directory Specification isn't based on anything AFAIK, it's self-contained. It's evolution of idea, that programs shouldn't treat root of user home dir as place to dump all state/config/cache files.

You can read more about how it's implemented at https://wiki.archlinux.org/title/XDG_Base_Directory.

Also there are go libraries that help implementing this spec https://github.com/adrg/xdg.

mumoshu:
@z0rc Thanks! I was more interested in what `~/.cache` is based on, rather than what `XDG Base Directory Specification ` is based on(if that makes sense

z0rc:
@mumoshu I'm not sure I understand you question. `~/.cache` folder is described in there:

> $XDG_CACHE_HOME defines the base directory relative to which user-specific non-essential data files should be stored. If $XDG_CACHE_HOME is either not set or empty, a default equal to $HOME/.cache should be used.

mumoshu:
@z0rc Ahhhh, sorry I somehow missed that part. We'd just follow the spec then! Thanks a lot for sharing your insights ☺️

itscaro:
@mumoshu should I use `https://github.com/adrg/xdg` ?

mumoshu:
@itscaro Yeah, let's build on top of `adrg/xdg`- I browsed code to see if it can be easily reimplement so that we don't need additional dependency, but it turned out not that easy especially when we need to make it cross-platform https://github.com/adrg/xdg/search?q=envCacheHome

itscaro:
@mumoshu I gave a try with os.UserCacheDir()
I modified the instance of remote used in `goGetterChart()` to set base / home dir as empty, when empty the cache dir  is used. This change applies to remote chart only. The mechanism of loading remote helmfiles in `visitStatesWithSelectorsAndRemoteSupport()` is not modified.

z0rc:
@itscaro one minor comment. Folders created in `UserCacheDir()` shouldn't start with dot. Right now on linux code uses `~/.cache/.helmfile`, while it should be `~/.cache/helmfile`.

itscaro:
@mumoshu @z0rc
Finally, I added a commit to move downloaded remote helmfiles to the cache dir as well. I added a new command group `cache` to provide a basic cache management.

I tried to separate these changes in different commits to be able to remove from PR if needed.

mumoshu (approved):
Awesome! Thank you for your contribution @itscaro and @z0rc ☺️

PR #2023, opened 2021-12-16
harmonise HELMFILE_TEMPDIR in secret decrypting & chart pulling

itscaro:


pkg/helmexec/exec.go
```diff
@@ -344,6 +347,18 @@ func (helm *execer) DecryptSecret(context HelmContext, name string, flags ...str
...
+	workDir := os.Getenv("HELMFILE_TEMPDIR")
+	if workDir == "" {
+		var err error
+		workDir, err = ioutil.TempDir(os.TempDir(), pattern)
+		if err != nil {
+			return "", err
+		}
+	}
+	return workDir, nil
+}
```
mumoshu, line 360: I guess it would be even more nicer if you could also refactor the below:

https://github.com/roboll/helmfile/blob/ae942c5288895c84c79171e5446773e4cb41c4ce/pkg/state/temp.go#L35-L41

to use this `tempDir` func as well!

itscaro: Yes I saw that, but it's in a different pkg and will cause cyclic import. I will try to put to another pkg such as utils, how do you think?

mumoshu: Another pkg sounds good! If I were you I'd name it `pkg/runtime` because I imagine there would be a lot of runtime-environment specific code that are scattered across the code-base now.

itscaro: @mumoshu I renamed the package, but now CircleCI gives me errors that I don't understand, if you have some time, could you help me.

itscaro:
@mumoshu  I found out that this file https://github.com/roboll/helmfile/blob/master/test/integration/secrets_templated.yaml.gotmpl contains `exec` which call a relative script https://github.com/roboll/helmfile/blob/master/test/integration/scripts/echo.sh

When I move decrypted files into temporary directory, these files lose the context of relative directory. So for now I leave decrypted files in place as before. What do you think?

mumoshu:
> So for now I leave decrypted files in place as before. What do you think?

@itscaro Sounds good! Yeah we need to keep the values and decrypted secret files directory consistent as declared in helmfile.yaml's `values` section, otherwise it will break existing use-cases as you noticed.

PR #2024, opened 2021-12-16
bump to go 1.17.3 with alpine 3.13

itscaro:


mumoshu:
Oh well what's going on with the image build? 🤔 

```
make: /bin/sh: Operation not permitted
make: git: Operation not permitted
make: /bin/sh: Operation not permitted
```

mumoshu:
I guess there should be some go.mod/sum changes like we've seen in #1951

itscaro:
> I guess there should be some go.mod/sum changes like we've seen in #1951

Since we update Go version, go.mod is not required to be updated, I voluntarily did this to avoid updating dependencies along with the Go version bump. If you prefer, we could do it in the same PR :-)

> Oh well what's going on with the image build? 🤔

Something is not right with Circle CI, locally the build passes :(

itscaro:
@mumoshu All green ✅

mumoshu (approved):
LGTM! Thanks for your contribution ☺️

Issue #2027, opened 2021-12-19
--state-values-set / --state-values-file are not propagated to sub helmfiles


dany74q:
Heya @mumoshu !

The --state-values-set and --state-values-file flags are not propagated to sub-helmfiles, I've wondered if this is intetional.
That is, the following doesn't work:

`helmfile --state-values-set=pasten=pasten template`

```yaml
# helmfile.yaml
helmfiles:
  - pasten.helmfile.yaml

# pasten.helmfile.yaml
bases:
  - pasten.base.yaml

# pasten.base.yaml
environments:
  dev:
    values:
    - {{ .Values.pasten }}.state-values.yaml.gotmpl

# pasten.state-values.yaml.gotmpl
pasten: 1337
```

It yields the following:
`in ./helmfile.yaml: in .helmfiles[0]: in ./pasten.helmfile.yaml: error during pasten.base.yaml.part.0 parsing: template: stringTemplate:4:16: executing "stringTemplate" at <.Values.pasten>: map has no entry for key "pasten"`

The templating works as expected if you're using it at the parent helmfile level, though.
Was that by design ? if not, would it make sense to propagate them (and / or introduce a flag that will propagate them if set) ?

Thanks !

dany74q:
Friendly ping @mumoshu, happy Christmas !

dany74q:
PING @mumoshu (1.3.3.7): 56 data bytes
64 bytes from 1.3.37: icmp_seq=0 ttl=114 time=75.400 ms
64 bytes from 1.3.37: icmp_seq=1 ttl=114 time=65.069 ms

(:

mumoshu:
@dany74q Hey! Unfortunately, it's by design. I've repeatedly introduced everyone wanted this to write something like

```
helmfiles:
- path: subhelmfile.yaml
  values:
  - {{ .Values | toYaml | nindent 4 }}
```

if that's really what you wanted.

I remember that where was another feature request to add `inheritValues: true` for a sub-helmfile so that you don't need to write the above template for inheriting values. But I forgot the issue number.

dany74q:
Thanks @mumoshu !

What are your thoughts on introducing such a flag ? I think in our use case, at least, it'd be easier to set such a flag and have it be populated implicitly.

Let me know if you'd be ok with such a change, I'll make those changes to the PR.
Thanks !

mumoshu:
[...]

Issue #762, opened 2019-07-19
Allow opting in for inheriting helmfile values to sub-helmfile


mumoshu:
Context: https://github.com/roboll/helmfile/issues/759#issuecomment-512706651

iAnomaly:
@mumoshu I think this would nice if I am understanding correctly. I have struggled a bit trying to DRY out our existing helmfile in such a way that sane and protective defaults exist in a generic "service" helmfile which is then included as a sub-helmfile of actual service repo helmfiles in such a way that developers of the service don't have to concern (or accidentally hurt) themselves with generic infrastructure and CI/CD details if that makes sense.

My wish would be to set the environment and values in the sub-helmfile and have those merged with the implementing service helmfile values which are specific to that service.

For example, imagine basic ingress annotations that we want enforced for ALL services, but each service also merges in its own ingress hosts/paths. For our use case it would be nice if all values could be inherited implicitly without having to explicitly add them.

Currently as [described by you](https://github.com/roboll/helmfile/issues/725#issuecomment-506101418):
```yaml
helmfiles:
- path: path/to/subhelmfile.yaml
   values:
   - foo: {{whatever available in the parent}}
```

Feature flag [proposed by you](https://github.com/roboll/helmfile/issues/759#issuecomment-512706651):
```yaml
valuesInherited: true
```

It would also be interesting if you could just pass the {{.Values}} root and somehow merge that with the sub-helmfile values root:
```yaml
helmfiles:
- path: path/to/subhelmfile.yaml
   values:
   - .: {{.Values}}
```

mikesplain:
Yeah I'm pretty surprised this given the way helm it's self's values and subchart values work, which was a bit of a learning curve when starting to use helmfile.  I totally understand that there's always reasons, based on the descriptions above.  Are there any good work arounds for this?

mumoshu:
@mikesplain Hey! One possible workaround would be:

```yaml
helmfiles:
- path: path/to/subhelmfile.yaml
  values:
  - {{ toYaml .Values | nindent 4 }}
```

And I guess merging just works by adding one array item to to the above:

```yaml
helmfiles:
- path: path/to/subhelmfile.yaml
  values:
  - {{ toYaml .Values | nindent 4 }}
  - foo: bar
```

Given the versatility of this workaround, I'm uncertain we really need `valuesInherited: true` 🤔 

[...]

PR #2028, opened 2021-12-19
Propagate parent environment override values to sub helmfiles (#2027)

dany74q:
Fixes https://github.com/roboll/helmfile/issues/2027, unless it's by-design (:

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -794,9 +794,10 @@ func (a *App) visitStates(fileOrDir string, defOpts LoadOpts, converge func(*sta
 			if len(st.Helmfiles) > 0 {
 				noMatchInSubHelmfiles := true
 				for i, m := range st.Helmfiles {
+					overrideValues := append(defOpts.Environment.OverrideValues, m.Environment.OverrideValues...)
 					optsForNestedState := LoadOpts{
 						CalleePath:        filepath.Join(d, f),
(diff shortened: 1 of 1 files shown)


pkg/app/app.go
```diff
@@ -794,9 +794,10 @@ func (a *App) visitStates(fileOrDir string, defOpts LoadOpts, converge func(*sta
 			if len(st.Helmfiles) > 0 {
 				noMatchInSubHelmfiles := true
 				for i, m := range st.Helmfiles {
+					overrideValues := append(defOpts.Environment.OverrideValues, m.Environment.OverrideValues...)
 					optsForNestedState := LoadOpts{
 						CalleePath:        filepath.Join(d, f),
-						Environment:       m.Environment,
+						Environment:       state.SubhelmfileEnvironmentSpec{OverrideValues: overrideValues},
```
mumoshu, line 800: Thanks for your effort! But this is incorrect as you are now unable to specify `values` for each sub `helmfiles:` entry.

Can you just explicitly inherit values using go template?

mumoshu: Ah well sorry I forgot that I already mentioned that in https://github.com/roboll/helmfile/issues/2027#issuecomment-1001465338 😅

mumoshu:
I think #2027 can be merged into #762. Let's discuss there!

Issue #766, opened 2019-07-23
feat: add "fileExists" function


bitsofinfo:
I'm generating a list of additional values files for a release driven by some other dynamic configuration. I'd like to be able to detect existence of those files prior to declaring them in each `release` to avoid unexpected helm errors when the configured "extra value file" does not actually exist.

Would be great if we had a simple `fileExists` function available that could work like

```
{{ if fileExists $relativePath }}
...
```

somewhat related to #761

morganchristiansson:
Yes! This would be awesome!

Currently I hardcode an array of what files exists:
```
    values:
{{ if has (env "ENV") (list "dev" "stage" "prod")}}
      - values-{{ env "ENV" }}.yaml
{{ end }}
```

Something like this would be much better:
```
    values:
{{ if fileExists "values-{{ env "ENV" }}.yaml" }}
      - values-{{ env "ENV" }}.yaml
{{ end }}
```

AssafKatz3:
@bitsofinfo I think that you can set missingFileHandler to Info (per #2063 ) for it

anessi:
There is the `isFile` function implemented in this commit since quite a while: https://github.com/helmfile/helmfile/commit/ed436ba68bd80057be99c20296c26dd4996b69dc

Unfortunately it's undocumented on the page https://helmfile.readthedocs.io/en/latest/templating_funcs/

The usage is very simple:
```
{{ if isFile .file }}
  {{ $fileContent := readFile .file }}
  ...
{{- end }}
```

Before I have used the workaround from https://github.com/roboll/helmfile/pull/2045#issue-1100870665
with some adaption to make it cross-platform:

* in helmfile
```
  # set fileexists to sh or bat depending on OS
  {{- $fileexists := "./fileexists.sh" -}}
  {{- if contains "WINDOWS" ( upper (env "OS") ) -}}
    {{- $fileexists = "fileexists.cmd" -}}
  {{- end -}}

...
  {{ if eq (trim (exec $fileexists (list .ConfigmapValueFile))) "true" }}
...
  {{- end -}}
```

* shell script:
```bash
#!/bin/bash
if [ ! -f "$1" ]; then
  echo -n 'false'
else
  echo -n 'true'
fi
```

* batch file:
```bash
@echo off
if exist "%1" (
  echo true
) else (
  echo false
)
```

anessi:
[...]

PR #2045, opened 2022-01-12
Add template function isFile

itscaro:
My use case is to check for file existence before using it with `readFile`, or we might implement `must` functions instead.

For now as a workaround, I'm using something like `if eq "ok" (exec test -f <file> && echo "ok" || echo "ko")` to do the job :-)

### DIFF
--- a/pkg/tmpl/context_funcs.go
+++ b/pkg/tmpl/context_funcs.go
@@ -1,6 +1,7 @@
 package tmpl
 
 import (
+	"errors"
 	"fmt"
 	"github.com/ghodss/yaml"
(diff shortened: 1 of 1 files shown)


mumoshu:
Linking #766

mumoshu (approved):
LGTM. Thanks as always for your contribution ☺️

mhienle:
Can you please give an example of how the funtion should be used inside a helmfile.yaml?

jamie-sweeney:
I also am interested in seeing a simple example for this use-case, as I am encountering hurdles in usage; 
if I succeed I will post my simple example.

PR #1744, opened 2021-03-29
Add .StateValues as alias of .Values in releaseTemplateDataRelease

itscaro:
ralates #1743

PR #2056, opened 2022-01-22
Add .StateValues as alias of .Values in EnvironmentTemplateData

itscaro:
So that the same alias can be used in helmfile.yaml

related to https://github.com/roboll/helmfile/pull/1744


mumoshu (approved):
Thanks! Let's see how it goes

lindhe:
Is there any documentation on how to use `.StateValues`?

PR #1934, opened 2021-08-01
feat: Add ReadDir as a templating function

nlueb:
This PR adds a new templating function called `readDir`. With `readDir` users can read the contents of a specified directory and get a list of all contained files. This is useful when reading a bunch of files from a directory. The following example shows a snippet of a values file for configuring a Logstash deployment. Using only `readFile`, a user must specify each file by hand and adjust this list as the number of files to be read grows.
```yaml
logstash:
  configs:
    logstash.yml: |
      {{- tpl (readFile "config/logstash.yml.gotmpl") . | nindent 6 }}
    jvm.options: |
      {{- readFile "config/jvm.options" | nindent 6 }}
    pipelines.yml: |
      {{- readFile "config/pipelines.yml" | nindent 6 }}
  pipelines:
    beats-log4j.conf: |
      {{- readFile "config/pipelines/beats-log4j.conf" | nindent 6 }}
    nginx-access.conf: |
      {{- readFile "config/pipelines/nginx-access.conf" | nindent 6 }}
    nginx-error.conf: |
      {{- readFile "config/pipelines/nginx-error.conf" | nindent 6 }}
    syslog-logs.conf: |
      {{- readFile "config/pipelines/syslog-logs.conf" | nindent 6 }}
    tcp-logs.conf: |
      {{- readFile "config/pipelines/tcp-logs.conf" | nindent 6 }}
    udp-debug.conf: |
      {{- readFile "config/pipelines/udp-debug.conf" | nindent 6 }}
    udp-logs.conf: |
      {{- readFile "config/pipelines/udp-logs.conf" | nindent 6 }}
  certificates:
    ca.crt: |
      {{- readFile "config/certificates/ca.crt" | nindent 6 }}
    logstash.crt: |
      {{- readFile "config/certificates/logstash.crt" | nindent 6 }}
    logstash.key: |
      {{- readFile "config/certificates/logstash.key" | nindent 6 }}
```
With `readDir` the above snippet can be rewritten as follows:
```yaml
logstash:
  configs:
  {{- range readDir "config" }}
    {{ base . }}: |
      {{- if hasSuffix "gotmpl" . }}
          {{- tpl (readFile .) $ | nindent 6 }}
      {{- else }}
          {{- readFile . | nindent 6 }}
      {{- end }}
  {{- end }}
  pipelines:
  {{- range readDir "config/pipelines" }}
    {{ base . }}: |
      {{- readFile . | nindent 6 }}
  {{- end }}
  certificates:
[...]

PR #2058, opened 2022-01-22
Fix context of the ReadDir templating function

nlueb:
This PR fixes an issue brought up by @bhester01 in #1934. I made `ReadDir` a method of `*Context` as suggested by @mumoshu, and that seems to have resolved the issue.

### DIFF
--- a/pkg/tmpl/context_funcs.go
+++ b/pkg/tmpl/context_funcs.go
@@ -20,7 +20,7 @@ func (c *Context) createFuncMap() template.FuncMap {
 	funcMap := template.FuncMap{
 		"exec":             c.Exec,
 		"readFile":         c.ReadFile,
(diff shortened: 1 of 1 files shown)


mumoshu (approved):
@nlueb I tweaked the code style and ran `make fmt` for automated code formatting a bit, but this LGTM. Thanks a lot for your contribution!

nlueb:
@mumoshu There was a reason why `context_path` was separated from `path`. The context must be added for the call to `os.ReadDir()`, but returning the file path with the context added causes `readFile` to add the context a second time.

mumoshu:
@2061 Ah, thanks for pointing that out! I've tried to fix it in #2061. I'd appreciate it if you could review it.

PR #2061, opened 2022-01-23
Fix readdir regression while merging

mumoshu:
Ref https://github.com/roboll/helmfile/pull/2058#issuecomment-1019439394


nlueb:
@mumoshu Thanks for being so fast to respond. From my local tests, everything seems to be working now. :tada:

mumoshu:
@nlueb Thanks for your confirmation ☺️

PR #2064, opened 2022-01-25
Allow more characters in label selectors

### LINKED ISSUE #970: feat: Allow more characters in release labels
I have a release label

```
version: x.y.z
```

and helmfile cmd w/ `--selector whatever=dog,version=x.y.z`

```
Malformed label: version=x.y.z. Expected label in form k=v or k!=v in ...[mytemplate.yaml]...
```

Looking at:

https://github.com/roboll/helmfile/blob/c68fc5bc50ea80f5782381f1f4a1ce975ad98e06/pkg/state/release_filters.go#L51

Why are there any character restrictions on helmfile release selectors/labels anyway? They just are used to control what releases helmfile processes, and are not necessarily labels that end up in k8s so I don't see the need to comply with a specific character restriction format?

Sajfer:
label selectors now also allows . and / in filter.

Not sure if we should allow more characters? Maybe everything except whitespace?

Fixes #970

### DIFF
--- a/pkg/state/release_filters.go
+++ b/pkg/state/release_filters.go
@@ -54,8 +54,8 @@ func ParseLabels(l string) (LabelFilter, error) {
 	lf.negativeLabels = [][]string{}
 	var err error
 	labels := strings.Split(l, ",")
-	reMissmatch := regexp.MustCompile("^[a-zA-Z0-9_-]+!=[a-zA-Z0-9_-]+$")
-	reMatch := regexp.MustCompile("^[a-zA-Z0-9_-]+=[a-zA-Z0-9_-]+$")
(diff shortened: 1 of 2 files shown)


itscaro:
@Sajfer You can add `+` too, semver uses `+` to separate metadata part.

Thank your for your contribution.

Sajfer:
I will see if I can fix it during the weekend.

Sajfer:
@itscaro added + and a pair of tests.

mumoshu (approved):
LGTM. Good catch & Thanks a lot for fixing it @Sajfer!

Also- Thank you so much for reviewing @itscaro ☺️

PR #2077, opened 2022-02-02
feat: Add --suppress option for diff and apply commands

sjentzsch:
We need to suppress the output of certain Kubernetes API objects in the diff/apply output.

As the helm-diff plugin supports `--suppress` to allow suppression of the values (kubernetes objects) listed in the diff output, with this MR, helmfile also supports using this flag now.


mumoshu (approved):
LGTM. Thank you for your contribution @sjentzsch ☺️

PR #2081, opened 2022-02-08
enable get() to be used with alias $.StateValues

itscaro:


mumoshu (approved):
LGTM. Thanks for your contribution as always!

Issue #2080, opened 2022-02-08
[Security][Question] Is there a possibility to disable `exec` in templates?


keskad:
Hi :slightly_smiling_face: 

I wanted to use Helmfile with ArgoCD and I was looking at this plugin https://github.com/travisghansen/argo-cd-helmfile
Then a security notice drew my attention that helmfile has insecure functionalities that allows execution of any scripts.

**Background:**
- ArgoCD executes Helmfiles with privileges of `argocd-repo-server`
- I would like to achieve an architecture where CI produces artifacts and does not have access to cluster, after merging to GIT - ArgoCD can apply those changes to separate access to cluster from CI pipelines which can be modified by development teams (https://github.com/jenkins-x/jx/issues/8060)

**My question is**
Is it possible to block `exec` and any other possible security-breaking features by e.g. setting an environment variable?

Thanks for your time, thanks for this fantastic project :blush:

mumoshu:
@keskad It's not possible today! But I'm very eager to include some feature flag to disable the functionality (perhaps something like `DISABLE_INSECURE_FEATURES=true` as a starter) if anyone could contribute one 👍 

Probably the list of things to be disabled for security would include:

- template functions (exec, readFile + relative path)
- go-getter support for remote helmfile fetching)
- helmfile hooks (allows hooking arbitrary shell commands on helmfile execution)

keskad:
I see @itscaro already started working on this, nice! Thank you! :smile:

PR #2083, opened 2022-02-11
introduce DISABLE_INSECURE_FEATURES to disable insecure executions

itscaro:
related to #2080


keskad:
Wow, awesome work!

mumoshu (commented):
This looks great as a starter!
It would be great if we could write a test to ensure this functionality is working.
(Maybe another go test file that executes Helmfile directly with a test helmfile.yaml that tries to use insecure features? 🤔

PR #2089, opened 2022-02-22
Update USERS.md

yohanb:
Added ShareGate to the list of companies that happily use helmfile in production :+1:


mumoshu (approved):
Thank you for using Helmfile! ☺️

PR #2121, opened 2022-03-30
Enhance support for kube-version and api-versions

### LINKED ISSUE #1864: Chartify renders objects with incorrect apiVersion
Hey guys,

Just another issue I've noticed with Chartify and Helmfile: while trying to install `prometheus-community/kube-prometheus-stack` chart with Helmfile and setting `forceNamespace` key in my `helmfile.yaml`, the Ingress object is generated with `apiVersion: networking.k8s.io/v1` which is not supported by our current K8s version. So it seems that the logic in `kube-prometheus-stack.ingress.apiVersion` [helper function](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/_helpers.tpl#L104) is just ignored.

And then, when I'm trying to do, for example, `helmfile diff` I get the following error:
```
error: unable to recognize "/var/folders/km/x2g0dhc17mn7sb5czxsmz3gw0000gp/T/chartify134751137/arn:aws:eks:eu-west-1:XXX:cluster/XXX/monitoring/prometheus-operator-ec2/kube-prometheus-stack/files/templates/prometheus/ingress.yaml": no matches for kind "Ingress" in version "networking.k8s.io/v1"
```

However, once I remove the `forceNamespace`, everything is working just fine.

Could you please assist?

How to reproduce: basically the configurations are the same as I described in #1857, but the cluster version should be the one that doesn't support the `networking.k8s.io/v1` API version yet(in our case it's 1.18, which only supports `networking.k8s.io/v1beta1`).

K8s version: 1.18
Helm version: 3.5.4
Helmfile version: 0.139.7

Thanks a lot,

Alex.

mumoshu:
This adds support for `kube-version` and `api-versions` to be available to `chartify` so that it works even if your release requires `chartify` due to that you use features like `forceNamespace`, `jsonPatches`, `strategicMergePatches`, and so on.

This also enhances `ReleaseSpec` which corresponds to each item of `releases[]` in your `helmfile.yaml` to also accept `kubeVersion` and `apiVersions`, in addition to the top-level `kubeVersion` and `apiVersions` we have today.

The top-level ones works as the default values for release-specific ones. If you have been using the top-level ones, keep using it. It is backward-compatible. If you want to specify it per release, because, for example, your releases are deployed across clusters(in case you differentiate `kubeContext` fields), try the new fields added to the release spec.

Resolves #1864

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -60,8 +60,12 @@ type ReleaseSetSpec struct {
 	CommonLabels        map[string]string `yaml:"commonLabels,omitempty"`
 	Releases            []ReleaseSpec     `yaml:"releases,omitempty"`
 	Selectors           []string          `yaml:"-"`
-	ApiVersions         []string          `yaml:"apiVersions,omitempty"`
-	KubeVersion         string            `yaml:"kubeVersion,omitempty"`
+
+	// Capabilities.APIVersions
+	ApiVersions []string `yaml:"apiVersions,omitempty"`
+
+	// Capabilities.KubeVersion
+	KubeVersion string `yaml:"kubeVersion,omitempty"`
 
 	// Hooks is a list of extension points paired with operations, that are executed in specific points of the lifecycle of releases defined in helmfile
 	Hooks []event.Hook `yaml:"hooks,omitempty"`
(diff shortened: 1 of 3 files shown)


rayterrill:
@mumoshu This looks awesome - will this also work for specifying kubeVersion at the environment level? We do a lot of custom configs per environment - would be nice to be able to specify it there if possible vs at the release.

PR #2132, opened 2022-04-11
Redact sensitive informations like password in chart url

dol:
The chart name can also be an URL that contains basic authentication information. This information are printed out by helmfile. This PR tries to redact this information.

```
$ cat <<EOT >> helmfile.yaml
releases:
    - name: test
      chart: https://foo:bar@charts.helm.sh/incubator/packages/druid-0.2.13.tgz
EOT

$ helmfile template
> Templating release=test, chart=https://foo:bar@charts.helm.sh/incubator/packages/druid-0.2.13.tgz
```

With this patch the output looks like:
```
> Templating release=test, chart=https://foo:xxxxx@charts.helm.sh/incubator/packages/druid-0.2.13.tgz
````

### DIFF
--- a/pkg/helmexec/exec.go
+++ b/pkg/helmexec/exec.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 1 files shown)


mumoshu:
@dol Hey. Thanks for your contribution! This looks awesome, but we've recently moved to https://github.com/helmfile/helmfile/. Would you mind reopening this PR there?

dol:
@mumoshu Will do.

